PR #3, opened 2023-09-21
Dr 2446 dockerize application

avertrees:
## Ticket:

- JIRA ticket [DR-2446](https://jira.nypl.org/browse/DR-2446)

- Added Dockerfile and docker-compose.yml files.
- Creates initial Docker support for the application.
- Adds Docker documentation to Readme
- Creates .env.example file

## How has this been tested?

- by running `docker-compose build` and `docker-compose up` locally and going to http://localhost:3000/. 

-

### Checklist:

- [x ] I have added relevant accessibility documentation for this pull request.
- [ x] All new and existing tests passed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM node:18-alpine AS production
+
+#RUN apt-get update
+#RUN apt-get upgrade -y
+
+ARG NYPL_HEADER_URL
+
+WORKDIR /app
+
+
+# Set environment variables. NODE_ENV is set early because we
+# want to use it when running `npm install` and `npm run build`.
+ENV PATH /app/node_modules/.bin:$PATH 
+# \
+#     NODE_ENV=production
+
+# Set the timezone
+# ENV TZ=America/New_York
+# RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# Install dependencies.
+COPY package.json ./
+COPY package-lock.json ./
+RUN npm install
+
+# Copy the app files.
+COPY . .
+
+ENV NYPL_HEADER_URL=${NYPL_HEADER_URL}
+
+# Build the app!
+RUN npm run build
+
+# Explicitly set port 3000 as open to requests.
+EXPOSE 3000
+
+# CMD is the default command when running the docker container.
+CMD [ "npm", "start" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,13 +7,6 @@ ARG NYPL_HEADER_URL
 
 WORKDIR /app
 
-
-# Set environment variables. NODE_ENV is set early because we
-# want to use it when running `npm install` and `npm run build`.
-ENV PATH /app/node_modules/.bin:$PATH 
-# \
-#     NODE_ENV=production
-
 # Set the timezone
 # ENV TZ=America/New_York
 # RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ARG NYPL_HEADER_URL
+
+WORKDIR /app
+
+
+# Set environment variables. NODE_ENV is set early because we
+# want to use it when running `npm install` and `npm run build`.
+ENV PATH /app/node_modules/.bin:$PATH 
+# \
+#     NODE_ENV=production
```
EdwinGuzman, line 15: This can be uncommented and the comment above can directly above this line.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Set the timezone
+# ENV TZ=America/New_York
+# RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
EdwinGuzman, line 19: I think this was done in a project because of timezones affecting unit tests. Do you think DC will have timezone issues? If not, we can remove it.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: "3"
+
+services:
+  app:
+    image: digital-collections
+    build:
+      context: .
+      args:
+        - NODE_ENV=local
```
EdwinGuzman, line 9: This should be `NODE_ENV=production`.

EdwinGuzman: this points to the Dockerfile that has its own NODE_ENV value set but it's currently commented out. It's fine to set it in both.

README.md
```diff
@@ -43,6 +43,21 @@ The `pages/api` directory is mapped to `/api/*`. Files in this directory are tre
...
+First, install ["Docker"](https://www.docker.com/) on your local machine. 
+
+After installing docker, cd into the app directory and run the following commands:
+
+```bash 
+docker-compose build # this will build the application and install dependencies
+```
+
+```bash 
+docker-compose up # this will run the application
```
EdwinGuzman, line 56: Can you also add the `docker-compose down` command to stop the application. I don't think that  removes the image from the local machine.

EdwinGuzman (commented):
Some minor updates but I ran the docker commands and it works. This should work on ECR/ECS once we get a chance to test it.

Dockerfile
avertrees: The build failed when `\ NODE_ENV=production`  was uncommented because of issues installing the dependencies. I'll try again and see if there's anything I can do to fix the issue.

EdwinGuzman: Try moving it right before the `npm run build` command.

EdwinGuzman: I'm not sure you need the `\`, that doesn't seem right.

avertrees: I think the ` \` is a new line, supposed to be at the end of line 13. Copied this line from https://github.com/NYPL/dc-poc/blob/main/dockerfile#L11. 

I'll try moving it before `npm run build`.

EdwinGuzman: Ah, yes, that is a new line character. Since this is a single env var, we don't need the `\`.

avertrees: Yeah, the newline character was creating some issues. Moving the line closer to `npm run build` and using "ENV  NODE_ENV=production" fixed it.

README.md
```diff
@@ -43,6 +43,111 @@ The `pages/api` directory is mapped to `/api/*`. Files in this directory are tre
...
+
+### Building Docker Images
+
+To build a Docker image for this application, run:
+
+```
+$ docker build -t digital-collections .
+```
+
+This command will build an image tagged (`-t`) as `digital-collections-app` using the current directory. Any changes to the application will require a new tagged image to be created to view those changes. Either remove the existing image (copy the image ID to use in the `docker image rm` command) and run the same command above:
```
EdwinGuzman, line 80: Minor: the examples use `digital-collections` without the `-app` at the end so remove it here to be consistent. Or you can add `-app` to the end of the other places where the `digital-collections` string is used for the image name.

avertrees: oy, good catch

avertrees: do you have a preference?

EdwinGuzman: I don't.

EdwinGuzman (approved):
Tested it again with the env var updates and it build and runs. The text update is not a blocker so feel free to merge once that's done.

PR #6, opened 2023-10-06
Feature/dr 2464 hero component

avertrees:
## Ticket:

- JIRA ticket [xxx](url)

-

-

### Checklist:


src/components/campaignHeroSubText.tsx
```diff
@@ -0,0 +1,30 @@
...
+          This site is a living database with new materials added every day,
+          featuring prints, photographs, maps, manuscripts, streaming video, and
+          more.
+        </Text>
+        <Text color="ui.typography.body">
+        Our collections include some content that may be harmful or difficult to view. <DSLink href="https://digitalcollections.nypl.org/about#nypl_harmful_content_statement">Learn more</DSLink>
+        </Text>
+        <HorizontalRule/>
+        <Text color="ui.typography.body">
+        Featured Image:
```
7emansell, line 23: Space missing at the end, add `{" "}`

7emansell (commented):
Looks good, pair reviewed with Edwin

EdwinGuzman:
> Do we want separate tickets for mobile?
> Do we want to include exact sizes that images should be based on breakpoint or referring to design is enough?

This is mostly for QA, right? For the second question, if we can do that, that'd be great. I'm thinking something along the lines of "if it's less than 600px (our mobile width), then use image v with dimension w,x. Otherwise, use image v with dimensions y,z".

This does mean that every _image_ needs it's own mobile/desktop dimensions or can the mobile/desktop dimensions be used for all general featured images?

src/components/campaignHeroSubText.tsx
avertrees: this is a live saving comment, thank you

7emansell: you're so welcome 😇

src/pages/_app.tsx
```diff
@@ -1,6 +1,10 @@
-import '@/styles/globals.css'
+// import '@/styles/globals.css'
 import type { AppProps } from 'next/app'
-
+import {DSProvider} from '@nypl/design-system-react-components'
 export default function App({ Component, pageProps }: AppProps) {
-  return <Component {...pageProps} />
+  return (
```
EdwinGuzman, line 5: This might've gotten lost somewhere but can you add the following:
```
<Head>
  <meta name="viewport" content="width=device-width, initial-scale=1" />
</Head>
```

You'd need to import `Head` from `next/head` and you can place it above the DSProvider:
```
<>
  <Head>
    <meta name="viewport" content="width=device-width, initial-scale=1" />
  </Head>
  <DSProvider>
    ...
  </DSProvider>
</>
```

src/utils/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+	region = "full",
+	size = "!1600,1600",
+	rotation = "0",
+) => {
+	return `https://iiif.nypl.org/iiif/2/${imageId}/${region}/${size}/${rotation}/default.jpg`;
+};
+export default imageURL;
+
+/* 
+REGION
```
EdwinGuzman, line 22: The updated documentation in the README is really good and I don't think we need this anymore.

avertrees: ha, good catch

avertrees:
> > Do we want separate tickets for mobile?
> > Do we want to include exact sizes that images should be based on breakpoint or referring to design is enough?
> 
> This is mostly for QA, right? For the second question, if we can do that, that'd be great. I'm thinking something along the lines of "if it's less than 600px (our mobile width), then use image v with dimension w,x. Otherwise, use image v with dimensions y,z".
> 
> This does mean that every _image_ needs it's own mobile/desktop dimensions or can the mobile/desktop dimensions be used for all general featured images?

@EdwinGuzman These points came from the ticket, I don't really know what the answer is. 

Re: your comment on the second question - should I create the dimensions guide before or after merging this PR?

EdwinGuzman:
If it's post merging this PR, would it be possible to get others to review the guide as well? If it's in this PR it's okay but it might get lost with other JS- and hero-specific updates.

src/components/hero/campaignHeroSubText.tsx
```diff
@@ -0,0 +1,30 @@
...
+          This site is a living database with new materials added every day,
+          featuring prints, photographs, maps, manuscripts, streaming video, and
+          more.
+        </Text>
+        <Text color="ui.typography.body">
+        Our collections include some content that may be harmful or difficult to view. <DSLink href="https://digitalcollections.nypl.org/about#nypl_harmful_content_statement">Learn more</DSLink>
+        </Text>
+        <HorizontalRule/>
+        <Text color="ui.typography.body">
+        Featured Image:
```
EdwinGuzman, line 23: Btw, I'm not sure if you got a chance to add the `{" "}` code here for the extra space. The comment thread is no longer appearing, so just reiterating in case that got lost.

avertrees: oops, I think I resolved it by accident. thanks for the additional set of eyes.

avertrees:
@EdwinGuzman by "others" do you mean Lily and other members of the DS?

EdwinGuzman:
Others who are more familiar with or can review IIIF-related docs.

src/components/hero/campaignHeroSubText.tsx
```diff
@@ -0,0 +1,30 @@
...
+          featuring prints, photographs, maps, manuscripts, streaming video, and
+          more.
+        </Text>
+        <Text color="ui.typography.body">
+        Our collections include some content that may be harmful or difficult to view. <DSLink href="https://digitalcollections.nypl.org/about#nypl_harmful_content_statement">Learn more</DSLink>
+        </Text>
+        <HorizontalRule/>
+        <Text color="ui.typography.body" mb="0px">
+        Featured Image:{" "}
+        <DSLink href="https://digitalcollections.nypl.org/items/510d47e0-cb17-a3d9-e040-e00a18064a99"> Momoyogusa = Flowers of a Hundred Generations. </DSLink>
```
EdwinGuzman, line 24: For now, add this as a prop `color="var(--nypl-colors-ui-link-primary) !important` to both links. I'll talk to Marty about this so we can make it easier. There's a rule on the Hero component that makes this hard to update.

src/pages/index.tsx
```diff
@@ -1,114 +1,14 @@
-import Head from 'next/head'
-import Image from 'next/image'
-import { Inter } from 'next/font/google'
-import styles from '@/styles/Home.module.css'
+// import Head from 'next/head'
```
EdwinGuzman, line 1: The commented out code can be removed.

EdwinGuzman (approved):
Two minor comments but otherwise looks good! Approving now.

avertrees:
@EdwinGuzman going to merge, will work on the image breakpoint documentation after.

PR #7, opened 2023-10-12
Add Jest, Axe, React Testing library

7emansell:
## Ticket:

Ticket [DR-2571](https://jira.nypl.org/browse/DR-2571)

- Adds Jest
- Adds jest-axe
- Adds React Testing Library
- Adds some supporting types etc.
- Example test to show testing works

## How has this been tested?

Run locally.

N/A.

### Checklist:

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -1,0 +1,23 @@
+import nextJest from 'next/jest.js'
+ 
+const createJestConfig = nextJest({
+  // Provide the path to your Next.js app to load next.config.js and .env files in your test environment
+  dir: './',
+})
+ 
+// Add any custom config to be passed to Jest
+/** @type {import('jest').Config} */
+const config = {
+  // Automatically clear mock calls, instances and results before every test
+  clearMocks: true,
+  // Indicates whether the coverage information should be collected while
+  // executing the test
+  collectCoverage: true,
+  // Add more setup options before each test is run
+  // setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
+ 
+  testEnvironment: 'jest-environment-jsdom',
+}
+ 
+// createJestConfig is exported this way to ensure that next/jest can load the Next.js config which is async
+export default createJestConfig(config)
--- a/package.json
+++ b/package.json
@@ -18,16 +18,17 @@
     "next": "13.4.19",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "ts-node": "10.9.1",
     "typescript": "4.9.5"
   },
   "devDependencies": {
     "@nypl/design-system-react-components": "2.0.1",
-    "@testing-library/jest-dom": "^6.1.4",
-    "@testing-library/react": "^14.0.0",
-    "@types/jest": "^29.5.5",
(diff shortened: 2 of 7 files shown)


7emansell:
I now can't recreate it but was getting an issue with different React versions, though I could not find any other than 18.2.0 ?? But this is still in progress until I have clarity on that and I'd like to see if it happens for anyone else

tsconfig.json
```diff
@@ -17,17 +17,18 @@
...
     },
     "forceConsistentCasingInFileNames": true
   },
   "include": [
     "next-env.d.ts",
     "**/*.ts",
-    "**/*.tsx"
+    "**/*.tsx", 
+    "jest.config.mjs", 
+    "__tests__/index.test.js"
```
EdwinGuzman, line 31: Update this to `__tests__/**/*.test.ts` and add a similar one for `.tsx`. This way, it captures any subdirectories and any files that end in `.test.ts` or `.test.tsx`.

package.json
```diff
@@ -20,6 +21,13 @@
     "typescript": "4.9.5"
   },
   "devDependencies": {
-    "@nypl/design-system-react-components": "2.0.1"
+    "@nypl/design-system-react-components": "2.0.1",
+    "@testing-library/jest-dom": "^6.1.4",
```
EdwinGuzman, line 25: Let's pin the specific versions for now. Sometimes, minor update or patches can install new packages or possibly break something. This means removing the caret `^` for all these packages.

jest.config.mjs
```diff
@@ -0,0 +1,23 @@
...
+// Add any custom config to be passed to Jest
+/** @type {import('jest').Config} */
+const config = {
+  // Automatically clear mock calls, instances and results before every test
+  clearMocks: true,
+  // Indicates whether the coverage information should be collected while
+  // executing the test
+  collectCoverage: true,
+  // Add more setup options before each test is run
+  // setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
```
EdwinGuzman, line 17: Do create this file (and try making it a `.ts` file) and it can be similar to this one: https://github.com/NYPL/research-catalog/blob/main/jest.setup.js

With this file, you can remove line 6 from `__tests__/index.test.tsx` and place it in the setup file. It's not a problem, but it becomes tedious when you have to import that line for every test file.

jest.config.mjs
```diff
@@ -0,0 +1,23 @@
+import nextJest from 'next/jest.js'
```
EdwinGuzman, line 1: I've never used this and didn't know next and jest are so closely compatible. This is great. I wonder if the `.mjs` could be converted to `.ts`. Can you try it? Otherwise, keep it. It's also not a problem but we can chat about esm and commonjs at a later time.

EdwinGuzman (commented):
Some comments knowing you're still working through this.

EdwinGuzman (approved):
Great! Runs fine locally. Just wait on Ali's review.

avertrees:
Works for me locally!

The Hero Component PR was merged into main earlier today, so merge main into this branch and then it's good to go :)

PR #8, opened 2023-10-13
DR-2465 Json Swim Lanes

7emansell:
## Ticket:

Fixes JIRA tickets [DR-2465](https://jira.nypl.org/browse/DR-2465) and [DR-2453](https://jira.nypl.org/browse/DR-2453).

- Adds in the static individual swim lanes JSON(s) and the list of all lanes to be displayed JSON
- Creates API endpoints for each lane with structure `data/lanes/slug-name/data.json` (default Next behavior)
- Creates API endpoint for all lanes with structure `data/lanes.json`
- Tests endpoints

## How has this been tested?

Tested locally.

N/A.

### Checklist:


7emansell:
Just sample data rn

README.md
```diff
@@ -41,6 +41,43 @@ You can start editing the page by modifying `pages/index.tsx`. The page auto-upd
...
+### Individual Lane - GET /api/lanes/[...slug]
+---
+#### Query Parameter
+ type| required | note |
+ ----------- |----------- | ----------- |
+ string      | yes | Slugified title of swim lane
+
+Example: 
+```sh
+{ slug: "books-and-periodicals" }
```
EdwinGuzman, line 66: I think the example can be `api/lanes/books-and-periodicals` because the requestor will not be sending the data object itself to the API endpoint.

README.md
```diff
@@ -41,6 +41,43 @@ You can start editing the page by modifying `pages/index.tsx`. The page auto-upd
...
+- [/manuscripts-and-correspondence](http://localhost:3000/api/lanes/manuscripts-and-correspondence)
+- [/maps](http://localhost:3000/api/lanes/maps)
+- [/photographs](http://localhost:3000/api/lanes/photographs)
+- [/prints-and-drawings](http://localhost:3000/api/lanes/prints-and-drawings)
+- [/recently-digitized-collections](http://localhost:3000/api/lanes/recently-digitized-collections)
+
+
+### Individual Lane - GET /api/lanes/[...slug]
+---
+#### Query Parameter
```
EdwinGuzman, line 59: Do you think we need to make a distinction between URL query parameters (`api/lanes?slug=books-and-periodicals`) and the way nextjs defines query parameters (`api/lanes/books-and-periodicals`)?

src/pages/api/lanes/[...slug].ts
```diff
@@ -0,0 +1,32 @@
+import path from "path";
+import { promises as fs } from "fs";
+import type { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(
+  request: NextApiRequest,
+  response: NextApiResponse
+) {
+  const slug = request.query.slug;
```
Comment (EdwinGuzman) [inline src/pages/api/lanes/[...slug].ts:9]: It should be okay to make the name of this file `src/pages/api/lanes/[slug].ts` and on this line you can extract the value directly like `const { slug } = request.query;`. The `[...slug]` works as a catch-all but do we need that?

src/pages/api/lanes/[...slug].ts
```diff
@@ -0,0 +1,32 @@
...
+
+  // Find the absolute path of the json directory for the lanes/[...slug].json file
+  const lanesJsonDirectory = path.join(
+    process.cwd(),
+    "/src/data/lanes/" + slug
```
Comment (EdwinGuzman) [inline src/pages/api/lanes/[...slug].ts:14]: nitpick: interpolate strings. In this case: `/src/data/lanes/${slug}`.

EdwinGuzman (commented):
Works well. Just a question about using the slug name or the catch-all type.

src/pages/api/lanes/[...slug].ts
Comment (7emansell) [reply src/pages/api/lanes/[...slug].ts]: We do not need it since we don't anticipate data being nested any further in the `data/lanes` folder (as discussed)

EdwinGuzman (approved):

__tests__/[slug].test.tsx
```diff
@@ -0,0 +1,57 @@
...
+  it("should respond with 200 status and cleaned file contents for a GET request", async () => {
+    const mockReadFile = jest.spyOn(fs, "readFile");
+    mockReadFile.mockResolvedValue(JSON.stringify({ key: "value" }));
+
+    const request: NextApiRequest = {
+      method: "GET",
+      query: {
+        slug: "test-slug",
+      },
+    } as unknown as NextApiRequest;
```
Comment (EdwinGuzman) [inline __tests__/[slug].test.tsx:22]: Don't need to do anything but just pointing out this is odd 🙃

7emansell:
Current state works but I'm refactoring so that endpoints are pulled from just one JSON

__tests__/[slug].test.tsx
```diff
@@ -1,57 +1,22 @@
...
-    } as unknown as NextApiResponse;
-
-    await handler(request, response);
-
-    expect(response.status).not.toHaveBeenCalledWith(200);
-  });
-});
+import slugHandler from "@/pages/api/lanes";
+import { NextApiRequest } from 'next';
+
```
Comment (7emansell) [inline __tests__/[slug].test.tsx:3]: Not sure how to mock this thoroughly now since it's not just any slug? Is this test sufficient?

EdwinGuzman (approved):
Latest update looks good!

__tests__/[slug].test.tsx
Comment (EdwinGuzman) [reply __tests__/[slug].test.tsx:3]: Ah true. Now you're returning a 404 so you can mock a request with something like
```
query: {
  slug: "this-should-not-work",
}
```

and in the expect, it should be called with a status of 404. `mockResponse` should also have the data object that says `{ error: "Lane not found"}`.

Comment (EdwinGuzman) [reply __tests__/[slug].test.tsx:3]: Sorry, so for this test this is enough (I don't think you need to verify the data itself since that'll change over time). But add another test for the 404 case.

__tests__/[slug].test.tsx
```diff
@@ -1,57 +1,43 @@
...
+  const mockResponse: any = {
+    status: jest.fn(function () {
+      return this;
+    }),
+    json: jest.fn(),
+  };
+
+  slugHandler(request, mockResponse);
+
+  expect(mockResponse.status).toHaveBeenCalledWith(404);
```
Comment (7emansell) [inline __tests__/[slug].test.tsx:41]: @EdwinGuzman If you have a minute to look at this– I've been trying new approaches since Friday afternoon and cannot get this test to pass, though it enters the `else` block correctly (in `[slug].ts`) and sends the error message? Status is always 200.

Comment (EdwinGuzman) [reply __tests__/[slug].test.tsx:41]: @7emansell it's because of where the `slugHandler` function is being imported from. It should be `"@/page/api/lanes/[slug]"` to get the right handler for the endpoint. It should work once that's updated.

This brings up another good point about the test that _is_ passing above. It only tests for 200 status but not if the response is only one lane. You **don't** need to make that update now.

What would be great is to structure the `__tests__` folder so that this file and `lanes.test.tsx` are under a similar structure as the structure where the file is located. So in this case `__tests__/api/lanes/...`.

While we're at it, `__tests__/index.test.tsx` is not an issue now but it could be confusing later when new files are added. Do you think that should be renamed now or at all? It can wait until later.

Comment (7emansell) [reply __tests__/[slug].test.tsx:41]: Oh my god never would have noticed that! Thank you. And I can rename it to `__tests__/homepage.test.tsx` if that makes sense to you? I guess we don't have a strict test naming structure yet

Comment (EdwinGuzman) [reply __tests__/[slug].test.tsx]: Yes, that does make sense, thanks!

EdwinGuzman (approved):

PR #13, opened 2023-10-20
DR-2448 setup travis

chrisb87:
I used a modified version of this script to push the `DR-2448-setup-travis-latest` image you can see in ECR. I also used another modified version to push `qa-latest` and could see that the QA cluster tasks updated correctly. However, the full test of this script will only happen when it is merged to qa.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,42 @@
+language: minimal
+services:
+  - docker
+branches:
+  only:
+    - qa
+env:
+  global:
+    - AWS_DEFAULT_REGION=us-east-1
+    - LOCAL_TAG_NAME=dc-frontend:$TRAVIS_BRANCH-$TRAVIS_BUILD_NUMBER
+    # DOCKER_LOGIN_PASSWORD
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+      env:
+        - ECR_URL=685731035297.dkr.ecr.us-east-1.amazonaws.com/dc-frontend:$TRAVIS_BRANCH-latest
+        - AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID_QA
+        - AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY_QA
+      install:
+        - pip install --user awscli
+      script:
+        - eval $(aws ecr get-login --no-include-email)
+        - docker build --tag "$LOCAL_TAG_NAME" .
+        - docker tag $LOCAL_TAG_NAME "$ECR_URL";
```
EdwinGuzman, line 30: I think you need quotes around `$LOCAL_TAG_NAME`.

EdwinGuzman (commented):
Looks good but just one potential issue.

.travis.yml
chrisb87: :+1: good catch, I updated that

EdwinGuzman (approved):

PR #15, opened 2023-10-23
DR-2443 Featured items endpoint

7emansell:
## Ticket:

Fixes JIRA ticket [DR-2443](https://jira.nypl.org/browse/DR-2443)

- Adds featured items endpoint
- Adds tests for endpoint

## How has this been tested?

Tested locally.

N/A.

### Checklist:


EdwinGuzman (commented):
Code-wise, this is working well. A question that might change a few things in this PR: should the feature items be conceptually the same as a lane? If not, perhaps this needs it's own API route such as `/api/featuredItems`?

I'm not sure how it's currently handled or the mental model in the current app, but what do both (@avertrees ) think?

avertrees:
> Code-wise, this is working well. A question that might change a few things in this PR: should the feature items be conceptually the same as a lane? If not, perhaps this needs it's own API route such as `/api/featuredItems`?
> 
> I'm not sure how it's currently handled or the mental model in the current app, but what do you both (@avertrees ) think?

I was also going to suggest an update to the API Route to something like /api/featuredItems or /api/featured/items. In the ticket, it's expected to be /api/lanes/items but retrospectively that is not very intuitive (my fault) and we have the flexibility to change it. 

Note though that we will need to update the AC in the ticket to reflect whichever API route we choose to go with.

7emansell:
Planning to discuss tomorrow morning before stand up and will edit accordingly!

EdwinGuzman (approved):

PR #18, opened 2023-10-25
DR-2574 Template for homepage

7emansell:
## Ticket:

Fixes JIRA ticket [DR-2574](https://jira.nypl.org/browse/DR-2574)

- Adds the template to the DC homepage

## How has this been tested?

Tested locally.

N/A.

### Checklist:


__tests__/homepage.test.tsx
7emansell, line 1: Best I could come up with for unit tests on components that don't actually have any content, but let me know if there's a more thorough way to do this

__tests__/homepage.test.tsx
```diff
@@ -17,4 +17,15 @@ describe("Home", () => {
     });
     expect(header).toBeInTheDocument();
   });
+
+  it("renders the expected text content", () => {
+    const { getByText } = render(<Home />);
```
7emansell, line 22: Best I could come up with for unit tests on components that don't actually have any content, but is there a more thorough way to do this?

EdwinGuzman: This is okay! We also had a hard time testing this on the DS. The accessibility test covers what we want as well.

src/pages/index.tsx
```diff
@@ -1,5 +1,38 @@
...
+  TemplateContent,
+  TemplateFooter,
+  TemplateHeader,
+} from "@nypl/design-system-react-components";
 
 export default function Home() {
-  return <CampaignHero />;
+  return (
+    <>
+      <SkipNavigation />
```
EdwinGuzman, line 15: Thanks for adding this, I forgot about this. It also needs a bit more clarity from Clare or Lana about where the "Skip to Main Content" link should go to. The link goes to `#mainContent` but there's no element on the page with this id. Can you add it to the h1 in the Hero (currently has `campaign-hero`) for now?

We can bring this up once the swim lanes are complete -- should we point to the hero heading or the first swim lane?

src/pages/index.tsx
```diff
@@ -1,5 +1,38 @@
...
+      <Template>
```
EdwinGuzman, line 16: Just curious, did you try the `TemplateAppContainer` component as well? You don't need to change it but just wondering if there were any issues with it.

src/pages/index.tsx
```diff
@@ -1,5 +1,38 @@
...
+          </TemplateHeader>
+          <CampaignHero />
+        </TemplateBreakout>
+        <TemplateContent>
+          <p>First swim lane</p>
+          <p>Featured Content</p>
+          <p>Rest of swim lanes</p>
+          <p>Explore further links</p>
+        </TemplateContent>
+        <TemplateFooter>
```
EdwinGuzman, line 32: The NYPL Footer will go in a different location so this is okay to remove.

EdwinGuzman (changes requested):
Just had one question about the `TemplateAppContainer` component and two minor updates.

src/pages/index.tsx
7emansell: Does this mean remove the template footer entirely?

src/pages/index.tsx
7emansell: Refactored to use `TemplateAppContainer`

src/pages/index.tsx
EdwinGuzman: Yes, correct.

EdwinGuzman (approved):

avertrees (approved):

PR #19, opened 2023-10-25
DR-2565 change image on reload

avertrees:
## Ticket:

- JIRA ticket [DR-2565](https://jira.nypl.org/browse/DR-2565)

- adds code to change featured image on reload

## How has this been tested?

tested locally, not sure what tests can be written for this atm?

-

### Checklist:


src/components/hero/campaignHero.tsx
```diff
@@ -3,16 +3,19 @@ import imageURL from "../../utils/utils";
 import CampaignHeroSubText from "./campaignHeroSubText";
 import CampaignHeroHeading from "./campaignHeroHeading";
 
-const CampaignHero = () => {
+const CampaignHero = ({ featuredItems }: any) => {
+  const randomIndex = Math.floor(Math.random() * featuredItems.images.length);
+  const imageID = featuredItems.images[randomIndex].split(".")[0];
```
EdwinGuzman, line 8: Let's create a function for this and move it, perhaps in `src/utils/utils.ts`? And call it in `getServerSideProps` in `src/pages/index.tsx` instead of in this component.

A few reasons why:

- `CampaignHero` doesn't need to know about the entire array of featured items, only the single object or model that contains the information it needs to render,
- it separates the functionality away from the component, so now this component only needs to worry about rendering information,
- it will make this random function easier to call in other functions, which I _think_ you may need to in some manner to also get the `featuredItems.images[randomIndex]`'s title and alt text,
- by calling it in `getServerSideProps`, this random function feature happens on the server side. Admittedly, this isn't a resource-heavy process but it'd be good to get into a habit of moving these calls on the server and have the data to render ready for the components.

EdwinGuzman (changes requested):
One change. Let me know if it makes sense.

I'd like to talk about starting to type the prop objects but it doesn't need to happen for this PR. Once you start working on the featured item title and alt text it'll be clearer how to create the type.

EdwinGuzman (approved):
More work needs to be done for the types so that `any` is not used, but approving now because there is more upcoming work on the `Hero` component for the title and link.

The object model for the hero "featured content" can be worked on once we have more info.

7emansell (approved):

PR #21, opened 2023-10-30
DR-2588 Swimlanes

7emansell:
## Ticket:

Fixes JIRA ticket [DR-2588](https://jira.nypl.org/browse/DR-2588)

- Adds the swim lanes to DC home page
- Adds testing for swim lanes rendering

## How has this been tested?

Tested locally.

N/A.

### Checklist:


src/pages/index.tsx
```diff
@@ -1,24 +1,48 @@
...
   );
 }
+
+export async function getServerSideProps() {
+  const lanes = data.lanes;
+  const lanesWithNumItems = await Promise.all(
+    lanes.map(async (lane) => {
+      const updatedCollections = await Promise.all(
+        lane.collections.map(async (collection) => {
+          const apiUrl = `/api/getNumItems/?uuid=${collection.uuid}`;
```
EdwinGuzman, line 24: This can work if the URL is updated to `http://localhost:3000/api/getNumItems/?uuid=${collection.uuid}`. However, `getServerSideProps` and the internal Next.js API endpoint `/api/getNumItems` are both on the server. Making a server call to a server endpoint is not necessary. When deployed to Vercel or AWS, it won't really know what "http://localhost:3000" is and it's not best practice to include that in API calls. (For a client call to the server endpoint, using `/api/getNumItems?uuid...` is correct).

Recommendation:

Remove the `/api/getNumItems` endpoint and convert the `getNumItems` function so that it returns the item count instead of a `res.status(...).json(..)` response. Then you can call the function here in `getServerSideProps` as `const { numItems } = getNumItems(collection.uuid)`. This function can be moved to a "home page utils" or "collection utils" file in case more related functions are added to the file later on.

If `/api/getNumItems` is needed, you can reuse the logic of the `getNumItems` function in both the API endpoint handler function and in this `getServerSideProps` function.

7emansell: @EdwinGuzman That makes way more sense now thank you

src/pages/index.tsx
```diff
@@ -1,24 +1,43 @@
...
-  SkipNavigation,
-  TemplateAppContainer,
-} from "@nypl/design-system-react-components";
+import SwimLanes from "../components/swimlanes/swimLanes";
+import { TemplateAppContainer } from "@nypl/design-system-react-components";
+import data from "@/data/lanes";
+import { getNumItems } from "@/utils/utils";
 
-export default function Home() {
+export default function Home({ lanesWithNumItems }) {
```
7emansell, line 7: This seems to be really unwieldy code to pass this prop through this parent component. As far as I can tell, the solution would be to write a hook for `getNumItems` and move it inside the `swimLanes` component 
(` useEffect(() => {
    async function fetchData() {
      const lanes = data.lanes;` etc... )
But that doesn't work and I'm still unclear why.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,74 @@
...
+  CardHeading,
+  Flex,
+  Heading,
+  SimpleGrid,
+  Spacer,
+  Text,
+  Link,
+} from "@nypl/design-system-react-components";
+import React from "react";
+
```
7emansell, line 14: This where I'd like to use a custom hook, something like this:
`const [lanesWithNumItems, setLanesWithNumItems] = useState([]);

  useEffect(() => {
    async function fetchData() {
      const lanes = data.lanes;
      const updatedLanes = await Promise.all(
        lanes.map(async (lane) => {
          const updatedCollections = await Promise.all(
            lane.collections.map(async (collection) => {
              try {
                const numItems = await getNumItems(collection.uuid);
                return { ...collection, numItems };
              } catch (error) {
                return { ...collection, numItems: 0 };
              }
            })
          );
          return { ...lane, collections: updatedCollections };
        })
      );

      setLanesWithNumItems(updatedLanes);
    }

    fetchData();
  }, []);

  return lanesWithNumItems;
}
`
Maybe bc this is client-side/getting CORS errors, and when the API call is in `getServerSideProps`, it doesn't have that issue?

src/pages/index.tsx
EdwinGuzman: Yes, I understand the desire to not do prop drilling but because it's one level down, I don't think it's a big deal if the `lanesWithNumItems` gets passed from this `Home` component to the `SwimLanes` components. I'll continue by comment in the other thread.

src/components/swimlanes/swimLanes.tsx
EdwinGuzman: This is a good idea but you're right it's a client-side CORS issue. This makes sense since Repo API should be protected.

Even if Repo API updates and allows this site to make requests, the other issue with this approach is around security. In `getNumItems`, it's getting a value for `process.env.AUTH_TOKEN` and the client side cannot read (intentionally) that server side env variable.

You can create this custom hook if you:

- add back the `/api/getNumItems` endpoint,
- inside that API handler, call `getNumItems(...)`,
- create a new client-side function that makes calls to `/api/getNumItems?uuid=...`. This new function is what you would call here instead of `await getNumItems(collection.uuid)`.

It's not a bad approach but some of my thoughts:
- If you can make the request on the server to get the data and pass it to the client, it'll be faster since the server is faster than the client. When the page loads, you'd be making multiple requests to the internal API endpoint for every item and in turn that will make requests to the Repo API.
- From the previous comment, because there's one parent component between `getServerSideProps` and `SwimLanes`, it seems fine to pass the data down as a prop.
- This approach makes it _slightly_ harder to test. You can test the custom hook in isolation but now the `SwimLanes` component is coupled with the hook. It can be easier to test `SwimLanes` if you can control the test data passed to it.
- This doesn't take into account that you need to render one swim lane, then featured content, then more swim lanes. The custom hook might make it harder to wedge the new component between lanes. This is what I'm thinking but there are definitely more ways to do this:
```
<SwimLanes lanesWithNumItems={lanesWithNumItems[0]} />
<FeaturedContent {...} />
<SwimLanes lanesWithNumItems={lanesWithNumItems[...]} />
```

I definitely think it's worth your time to try the approach and see the pros and cons of it. It'll give you a better idea of API calls and structure for this and future features.

src/pages/index.tsx
```diff
@@ -1,29 +1,44 @@
...
     />
   );
 }
 
 export async function getServerSideProps() {
-  return { props: { featuredImageID: featuredImageID() } };
+  const lanes = data.lanes;
+  const lanesWithNumItems = await Promise.all(
+    lanes.map(async (lane) => {
+      const updatedCollections = await Promise.all(
```
EdwinGuzman, line 23: Look into `Promise.allSettled` and see if you can use it. And you may not need the outer `Promise.all`.

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+    "typescript.tsdk": "node_modules/typescript/lib"
```
EdwinGuzman, line 2: I want to propose the following lines to this file
```
  "editor.formatOnSave": true,
  "editor.defaultFormatter": "esbenp.prettier-vscode",
```

It's a personal preference that allows Prettier to format the code when the file is saved. A proposal so not necessary!

__tests__/homepage.test.tsx
```diff
@@ -18,13 +18,77 @@ describe("Home", () => {
...
-  it("renders the expected text content", () => {
-    const { getByText } = render(<Home />);
-    expect(getByText("Notification banner")).toBeInTheDocument();
-    expect(getByText("Header")).toBeInTheDocument();
-    expect(getByText("First swim lane")).toBeInTheDocument();
-    expect(getByText("Featured Content")).toBeInTheDocument();
-    expect(getByText("Rest of swim lanes")).toBeInTheDocument();
-    expect(getByText("Explore further links")).toBeInTheDocument();
+  it("renders the Recently Digitized Items row", () => {
+    render(<Home />);
```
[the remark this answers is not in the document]
EdwinGuzman: Btw, for the `swimLanes.test.tsx` file, since it's a component test file and not a _page_ or API test file, it can be colocated in the same directory as the component file. It doesn't have to be in this `__test__` directory.

7emansell:
Requesting review @EdwinGuzman @avertrees on everything except whether the "See more" links work (they don't, but once the pages exist that should be a very quick fix). Out of scope but included in this PR are some updates to VS Code settings and a reorganization of the test suites.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,76 @@
...
+                >
+                  <CardHeading
+                    id={`row-card-heading-${key}-${index}`}
+                    level="h3"
+                    size="heading4"
+                    url={collection.url}
+                    style={{
+                      display: "-webkit-box",
+                      WebkitLineClamp: 3,
+                      WebkitBoxOrient: "vertical",
```
EdwinGuzman, line 54: I'm not familiar with these rules. Why do you need them?

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,76 @@
...
+                      overflow: "hidden",
+                      whiteSpace: "pre-wrap",
+                    }}
+                  >
+                    {collection.title}
+                  </CardHeading>
+                  <CardContent>
+                    <Text role="heading" aria-level="4" size="subtitle2">
```
EdwinGuzman, line 62: This should be fine as regular text and not a heading so you can remove the `role` and `aria-level` props.

Is it suppose to be a heading? If it is, the better element to use is the `Heading` component.

src/pages/index.tsx
```diff
@@ -1,29 +1,48 @@
...
+        return { ...collection, numItems: 0 };
+      }
+    })
+  );
+  const updatedLanes = lanes.map((lane) => {
+    const updatedCollections = lane.collections.map(() => {
+      const result = collectionsWithNumItems.shift();
+      return result.status === "fulfilled"
+        ? result.value
+        : { ...result, value: 0 };
```
EdwinGuzman, line 37: `value` might be better as an empty object since the successful result contains `uuid`, `title`, etc. Otherwise, you may get errors attempting to make a reference such as `collection.uuid` when the returned result is `0`.

src/pages/index.tsx
```diff
@@ -1,29 +1,48 @@
...
       renderSkipNavigation={true}
     />
   );
 }
 
 export async function getServerSideProps() {
-  return { props: { featuredImageID: featuredImageID() } };
+  const lanes = data.lanes;
+  const flatCollections = [].concat(...lanes.map((lane) => lane.collections));
+  const collectionsWithNumItems = await Promise.allSettled(
```
EdwinGuzman, line 22: This and the `result.status === "fulfilled"` is good. We can refactor it a bit more to add error handling but let's make that a refactor ticket to work on later.

src/utils/utils.ts
```diff
@@ -28,3 +28,26 @@ export const imageURL = (
 ) => {
   return `https://iiif.nypl.org/iiif/2/${imageId}/${region}/${size}/${rotation}/default.jpg`;
 };
+
+export const getNumItems = async (uuid: string) => {
```
EdwinGuzman, line 32: Add jsdoc style comment to this function, similar to the `imageURL` comment block above.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <Flex alignItems="baseline">
+            <Heading id={`row-heading-${lane.slug}`} level="h2" size="heading3">
+              {lane.title}
+            </Heading>
+            <Spacer />
+            <Link
+              id={`row-see-more-${lane.slug}`}
+              type="standalone"
+              href={"https://digitalcollections.nypl.org/search/index?"}
+              aria-label="See more digitized materials"
```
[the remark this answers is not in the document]
EdwinGuzman: To be clear, it should be dynamic for every swim lane.

src/components/swimlanes/swimLanes.tsx
7emansell: These styles are the only solution I could find that would create the ellipsis when there's text overflow past the third line

src/components/swimlanes/swimLanes.tsx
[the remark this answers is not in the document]
EdwinGuzman: Sorry, the ticket was unclear and I confused "digitized materials" with the title of the first swim lane. This would be good to get clarification from Lana if that text is intentional and is not meant to have the title of the swim lane. And also clarification from Clare if multiple links on the page with the same aria-label is accessible/clear to the user.

src/components/swimlanes/swimLanes.tsx
EdwinGuzman: Yes, makes sense. Briefly chatted with @bigfishdesign13 (Marty) about this. I was concerned that the webkit prefix would not be supported in Firefox and other browsers but it looks like it's well supported: https://caniuse.com/?search=webkit-line-clamp.

Can you try it without the webkit prefix, so just `lineClamp`, and see if that still displays correctly in Chrome and Firefox? 

One other concern is that the ellipsis is not blue or (it seems) part of the link. Not sure yet if that's a big issue or not.

7emansell: No go without the webkit prefix unfortunately. And yes I don't know if that's a problem

src/components/swimlanes/swimLanes.tsx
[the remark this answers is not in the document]
7emansell: I'll leave a note in the ticket rn

EdwinGuzman:
Something to look into that isn't obvious in the code, when you tab through the page, the collection title's link focus ring is not visible. This is true for single line titles and for multiple lines you can roughly see the focus ring:

Screen Shot 2023-11-03 at 3 54 32 PM

The ticket doesn't specify if the images or the entire card should be linked, so it's functionally built correctly but the focus ring should be entirely visible around the tabbable elements.

7emansell:
> The ticket doesn't specify if the images or the entire card should be linked, so it's functionally built correctly but the focus ring should be entirely visible around the tabbable elements.

Noted, I'm looking into it bc I don't think there's an immediate easy solve

src/components/swimlanes/swimLanes.tsx
bigfishdesign13: I would suggest including both the standard and the `webkit` version of the two attributes.  This is still common for some CSS attributes.  In fact, I am suprised the packager does not take care of that for us as it does with some other CSS attributes.

EdwinGuzman: I don't think postcss was added or configured to add the prefixes. @7emansell that can be done at a later time with a specific ticket that has better details.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <Box data-testid={lane.slug} key={lane.rank} mb="l">
+          <Flex alignItems="baseline">
+            <Heading id={`row-heading-${lane.slug}`} level="h2" size="heading3">
+              {lane.title}
+            </Heading>
+            <Spacer />
+            <Link
+              id={`row-see-more-${lane.slug}`}
+              type="standalone"
+              href={`${process.env.DC_URL}collections/lane/${lane.slug}`}
```
EdwinGuzman, line 30: There are some ways to encapsulate the env variables so you don't have to type `process.env` within react components. One suggestion is to create an `appConfig.ts` file that contains all your env variables. Then you can import it in this file and not have `process.env`. 
```
// appConfig.ts

export {
  DC_URL: process.env.DC_URL,
  ...
};
```

EdwinGuzman (commented):
The update looks good. One minor suggestion with env variables.

avertrees (approved):

src/components/swimlanes/swimLanes.tsx
EdwinGuzman: I find it most helpful for organization so you can see all the env vars in one config file. Once you start adding more variables, it will also be helpful see the categorization of server-side only vars (e.g. for credentials) and client-side only vars that have the `NEXT_PUBLIC_` prefix.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,110 @@
...
+                  </CardContent>
+                </Card>
+              ))}
+            </SimpleGrid>
+          )}
+          <Link
+            id={`row-see-more-${lane.slug}-mobile`}
+            type="standalone"
+            href={`${appConfig.DC_URL}collections/lane/${lane.slug}`}
+            aria-label={`See more ${lane.title.toLowerCase()} mobile`}
```
EdwinGuzman, line 90: We chatted about this but here are more details: you can use `aria-hidden` so that a screenreader ignores that flagged element. Adding that will help remove the "mobile" from the aria-label here.

This _should_ be conditional, however, because in the mobile context you do want this link to be picked up by screenreaders and the desktop link to be ignored.

The DS has a helper hook `useNYPLBreakpoints` https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/docs/hooks-usenyplbreakpoints--docs that can help with that. It uses the `matchMedia` API to check the current width of the viewport. This will tell you if it "isLargerThanMobile" and you can then apply the `aria-hidden={true}` appropriately.

Another approach, which is an example we give in the doc, is to only render this `Link` on mobile and when `isLargerThanMobile` is false, then render `null` when it's true. The reverse would be the case for the other link above. Both approaches are similar and I don't think any is better than another.

EdwinGuzman:
Just had that one comment and we'll wait on new VQA feedback. Otherwise, the component and data fetching itself looks good.

src/components/swimlanes/swimLanes.tsx
7emansell: Change made and then changed back– the Axe issue was with the ID, which can't be a duplicate anyways. `display: none` implies `aria-hidden: true` to the screenreader, which Clare just explained on the ticket. The more you know!

EdwinGuzman: Great. I was confused because of the failing test you mentioned and having the same labels. I didn't know it was a duplicate id issue so this update makes sense.

EdwinGuzman (approved):

src/components/swimlanes/swimLanes.tsx
```diff
@@ -0,0 +1,104 @@
...
+                    src: imageURL(collection.image_id, "full", "288,", "0"),
+                  }}
+                >
+                  <CardHeading
+                    id={`row-card-heading-${key}-${index}`}
+                    level="h3"
+                    size="heading4"
+                    className={styles.collectiontitle}
+                    url={collection.url}
+                    noOfLines={3}
```
7emansell, line 60: @EdwinGuzman @bigfishdesign13 Stumbled on https://chakra-ui.com/docs/components/text/usage#truncate-text while I was working on the header, I think this is a way better solve for now even if doesn't address all the concerns in https://jira.nypl.org/browse/DSD-1625 ?

src/components/swimlanes/swimLanes.tsx
```diff
@@ -62,7 +62,7 @@ const SwimLanes = ({ lanesWithNumItems }) => {
                       <Text
                         size="subtitle2"
                         sx={{
-                          fontWeight: "400",
+                          fontWeight: "510",
```
bigfishdesign13: Why are custom styles being added for the subtitle?

7emansell: I accidentally changed it back to the default, but we are keeping it at `fontWeight: 400`, requested by Lily! https://jira.nypl.org/browse/DR-2588

src/components/swimlanes/swimLanes.tsx
EdwinGuzman: Yes! that's helpful. Can you add this comment to the DSD-1625 ticket? The Chakra versions we are using is v1 so the link should be https://v1.chakra-ui.com/docs/components/typography/text

It's still available in v1 and we're moving to v2 soon.

bigfishdesign13: We are repurposing the Chakra `Text` component to compose our `Text` component, so the truncate props should work on our component as is.

PR #22, opened 2023-11-08
DR-2547 featured image link

avertrees:
## Ticket:

- JIRA ticket [DR-2547](https://jira.nypl.org/browse/DR-2547)

-

-

### Checklist:


src/utils/utils.ts
```diff
@@ -37,7 +37,6 @@ export const imageURL = (
 export const getNumItems = async (uuid: string) => {
```
avertrees, line 37: @EdwinGuzman @7emansell wondering how to refactor this to use the getAPIUri method, main concern is that getAPIUri returns the api response and getNumItems returns the numItems.

src/components/hero/campaignHeroSubText.tsx
```diff
@@ -46,10 +48,10 @@ const CampaignHeroSubText = () => {
               color: "var(--nypl-colors-ui-link-tertiary) !important",
             },
           }}
-          href="https://digitalcollections.nypl.org/items/510d47e0-cb17-a3d9-e040-e00a18064a99"
+          href={featuredItem.href}
         >
           {" "}
-          Momoyogusa = Flowers of a Hundred Generations.{" "}
+          {featuredItem.title.$}{" "}
```
avertrees, line 54: @EdwinGuzman for longer titles, I was wondering how we could break them up to look more like the titles in the figma file https://www.figma.com/file/NpjG47HqZG9GknfijGqJri/DC-Facelift-Homepage?type=design&node-id=3966-5868&mode=design&t=KLTjKriR6HBxc8Jz-0

EdwinGuzman: Try updating to `display: inline`. It's currently set to `display: inline-block`. A few examples:

Screen Shot 2023-11-09 at 12 14 21 PM
Screen Shot 2023-11-09 at 12 13 36 PM
(on mobile)
Screen Shot 2023-11-09 at 12 14 30 PM

In the Figma file, the mobile view has it at a max two-lines long. @7emansell recently did text truncation that _might_ work in this case and you'd have to add it to the `Text` parent component. Charka docs: https://v1.chakra-ui.com/docs/components/typography/text#truncate-text

src/utils/utils.ts
```diff
@@ -47,8 +77,8 @@ export const getNumItems = async (uuid: string) => {
 
     if (response.status === 200) {
       const data = await response.json();
-      const numItems = data.nyplAPI.response.numItems;
-      return numItems;
+      const responseData = data.nyplAPI.response;
+      return responseData;
     } else {
       return 0;
```
EdwinGuzman, line 83: Now that this function doesn't return `numItems`, return an empty object (or perhaps undefined?) instead of a 0.

src/utils/utils.ts
```diff
@@ -36,8 +36,38 @@ export const imageURL = (
...
+};
+
+/**
+ * Returns the uuid, API uri, and numResults of an item given an identifier type and identifier value.
+ * @param {string} identifierType - the identifier type
+ * @param {string} identifier - the identifier value
+ */
+
+export const getAPIUri = async (identifierType: string, identifier: string) => {
+  const apiUrl = `https://api.repo.nypl.org/api/v2/items/${identifierType}/${identifier}`;
```
EdwinGuzman, line 60: It doesn't have to be for this PR but this should be environment-aware and the domain should be an env var. When it's on QA it should be the QA API domain and for prod it should use this domain. If not for this PR, add a todo so it doesn't get lost.

src/utils/utils.ts
```diff
@@ -36,8 +36,38 @@ export const imageURL = (
...
+  const res = await apiCall(apiUrl);
+  return res.numItems || 0;
+};
+
+/**
+ * Calls Repo API collectins/uuid/items endpoint
+ * @param {string} uuid - collection uuid
+ */
 
+export const getItemsFromUUID = async (uuid: string) => {
```
EdwinGuzman, line 48: This function isn't being called anywhere. Should it exist?

src/pages/index.tsx
```diff
@@ -42,10 +47,19 @@ export async function getServerSideProps() {
...
   });
 
+  const imageID = featuredImageID();
+  const apiUri = await getAPIUri("local_image_id", imageID);
+  const dataFromUri = await apiCall(apiUri.apiUri);
+  const featuredItemObject = {
+    imageID: imageID,
+    uuid: apiUri.uuid,
+    title: dataFromUri.mods.titleInfo.title,
+    href: `${process.env.DC_URL}/items/${apiUri.uuid}`,
```
EdwinGuzman, line 57: A double slash is added in the URL. You don't need the slash before "items". In the Vercel preview I get a link like this: `"https://qa-digitalcollections.nypl.org//items/5722fe60-5b93-0131-8261-58d385a7bbd0"`.

EdwinGuzman (commented):
The update is functional and added some suggestions and questions I have.

src/utils/utils.ts
avertrees: I will remove it

src/pages/index.tsx
avertrees: I'd like to remove the '/' from the DC_URL env variable so that DC_URL and API_URL have the same format, I think it looks cleaner to remove it from the variable vs. keeping it in. This will require an update in our local environments, vercel, QA and Prod environments.

7emansell: I second this– I shouldn't have had the slash in there in the first place

src/utils/utils.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+  featuredImageID,
+  getAPIUri,
+  apiCall,
+} from "@/utils/utils";
+
+// test that this data returned in dataFromUri is valid:
+describe("featuredImageID Object", () => {
+  it("should return a defined imageID, uuid, title, and href if query param image id is not valid", async () => {
+    const imageID = featuredImageID("1234");
+    const apiUri = await getAPIUri("local_image_id", imageID);
```
EdwinGuzman, line 14: This means it's making real API requests in the test, right? these should be mocked.

src/utils/utils.ts
```diff
@@ -1,13 +1,51 @@
 import featuredItemsData from "../data/featureditems.json";
+let featuredItemArray = [];
+
+/**
+ * Sets global featuredItemArray so other functions can use it.
+ **/
+export const setFeaturedItemArray = () => {
```
EdwinGuzman, line 7: You can update this function so that you don't have to make this logic check and call twice in different functions:
```
if (featuredItemArray.length === 0) {
  setFeaturedItemArray();
}
```.

Check if `featuredItemArray.length` is not 0 and if it is not, then return. Otherwise, go through the process of setting it.

```
if (featuredItemArray.length !== 0) {
  // The global array already has what it needs.
  return;
}

// If it's empty, now is the time to generate and update the array.
let itemsArray = [];
for (let i = 0; i < featuredItemsData.featuredItems.images.length; i++) {
  itemsArray.push(featuredItemsData.featuredItems.images[i].split(".")[0]);
}
// set featuredItemArray
featuredItemArray = itemsArray;

return featuredItemArray;
```

Now, in `generateRandomImageID` and `featuredImageID` (or anywhere else), can you call `setFeaturedItemArray()` directly without any if statement wrapper. All the logic will be inside of `setFeaturedItemArray`.

**Comment by EdwinGuzman** [comment]
Here's what I got for the 500 error: 
Screen Shot 2023-11-20 at 5 42 56 PM

I am using this url as my test: `http://localhost:3000/?imageID=1657176`.

1. I initially tried to add some optional chaining and an empty string as a default to that specific line of code as `title: data?.mods?.titleInfo?.title?.$ || "",`. 

- Either check if there is a `title` property and get the `$` value or default to an empty string.
- This worked but then I got the, now not surprising, 403 forbidden error.

2. Other image IDs have returned 403 initially, so why does this return the 500?

- The IIIF image url is still 403 but now there's a different issue.
- Go deeper into the API call. The image ID of `1657176` makes a call to `"http://api.repo.nypl.org/api/v2/mods/6bdb7bc0-c5eb-012f-a788-58d385a7bc34"` to get more information.
- This returns `data.mods.titleInfo` as an _array_ when the code is expecting a single object!
Screen Shot 2023-11-20 at 5 50 20 PM

3. If you update the line to `title: data.mods.titleInfo[0].title.$ || "",` this works now.
 
- The bigger problem is that `data.mods.titleInfo` can be an object or an array of objects.
- You can make an update to the code to try both or default to "" 
```
// a bit sloppy and could be cleaned up
let title = data?.mods?.titleInfo?.length
    ? data?.mods?.titleInfo[0]?.title?.$
    : data?.mods?.titleInfo?.title?.$;
  title = title || "";

  return {
    uuid: apiUri.uuid,
    title,
  };
```

Are there other properties that could be returned as an array of objects or a single object? If that's true, this will be troublesome but something that can be tackled in the future. For now, the snippet of code should fix the 500 issue. The IIIF 403s still exist.

**Comment by avertrees** [comment]
> Here's what I got for the 500 error: Screen Shot 2023-11-20 at 5 42 56 PM
> 
> I am using this url as my test: `http://localhost:3000/?imageID=1657176`.
> 
> 1. I initially tried to add some optional chaining and an empty string as a default to that specific line of code as `title: data?.mods?.titleInfo?.title?.$ || "",`.
> 
> * Either check if there is a `title` property and get the `$` value or default to an empty string.
> * This worked but then I got the, now not surprising, 403 forbidden error.
> 
> 2. Other image IDs have returned 403 initially, so why does this return the 500?
> 
> * The IIIF image url is still 403 but now there's a different issue.
> * Go deeper into the API call. The image ID of `1657176` makes a call to `"http://api.repo.nypl.org/api/v2/mods/6bdb7bc0-c5eb-012f-a788-58d385a7bc34"` to get more information.
> * This returns `data.mods.titleInfo` as an _array_ when the code is expecting a single object!
>   Screen Shot 2023-11-20 at 5 50 20 PM
> 
> 3. If you update the line to `title: data.mods.titleInfo[0].title.$ || "",` this works now.
> 
> * The bigger problem is that `data.mods.titleInfo` can be an object or an array of objects.
> * You can make an update to the code to try both or default to ""
> 
> ```
> // a bit sloppy and could be cleaned up
> let title = data?.mods?.titleInfo?.length
>     ? data?.mods?.titleInfo[0]?.title?.$
>     : data?.mods?.titleInfo?.title?.$;
>   title = title || "";
> 
>   return {
>     uuid: apiUri.uuid,
>     title,
>   };
> ```
> 
> Are there other properties that could be returned as an array of objects or a single object? If that's true, this will be troublesome but something that can be tackled in the future. For now, the snippet of code should fix the 500 issue. The IIIF 403s still exist. I also made the call to get the first object in the array but I don't know the context well enough to say that it is the correct object to use. There is a `usage: "primary"` which you can go by to verify that it should be used. (I also haven't read the Repo API to learn more about this just yet!)

@EdwinGuzman I approached solving this in a little bit of different way but I think the solution still works. Let me know if you have thoughts or strong feelings about my approach lol

File: src/components/hero/campaignHeroSubText.tsx [hunk 9]
```diff
@@ -29,29 +31,37 @@ const CampaignHeroSubText = () => {
...
-        >
-          {" "}
-          Momoyogusa = Flowers of a Hundred Generations.{" "}
-        </DSLink>
-      </Text>
+      <Tooltip content={featuredItem.title} shouldWrapChildren>
+        <Box>
+          <Text color="ui.typography.body" mb="0px" noOfLines={2}>
+            Featured Image:{" "}
+            {/* <Tooltip content={featuredItem.title} shouldWrapChildren> */}
```
avertrees, line 44: @EdwinGuzman for our 1:1, idk why but whenever I add the tooltip here it breaks the link text onto a new line.

src/utils/utils.ts
```diff
@@ -47,12 +117,35 @@ export const getNumItems = async (uuid: string) => {
...
+
+function addCommas(number) {
+  // Return the formatted number
+  return Number(number).toLocaleString("en-US");
+}
+
+function getTitleFromRepoAPIResponseData(data) {
+  // if titleInfo is an array of objects, use the first object. if titleInfo is a hash, return the title.
+  const titleInfo = data.mods.titleInfo;
+  return Array.isArray(titleInfo) ? titleInfo[0].title.$ : titleInfo.title.$;
```
EdwinGuzman, line 150: Using the first object in the array is fine. We chatted about this and using the `"usage": "primary"`.

EdwinGuzman (approved):
The comments i left are mostly nitpicky.

There's the issue with calling properties on the `data` object being risky if the API call fails. There isn't a lot of error handling for the API calls but that itself should be a new ticket so we can do all of them at once. I hope it's easier to see the potential issues now that more of this is built out.

I'm approving. I think regardless of the outcome of our chat about the Tooltip component, the Tooltip should be removed from this PR. Let's wait and see.

src/utils/utils.ts
```diff
@@ -1,13 +1,56 @@
...
  * Returns a random image ID from the list of featured items.
  */
-export const featuredImageID = () => {
-  const randomIndex = Math.floor(
-    Math.random() * featuredItemsData.featuredItems.images.length
-  );
-  return featuredItemsData.featuredItems.images[randomIndex].split(".")[0];
+export const generateRandomImageID = () => {
+  console.log("generating random imageID");
+  if (featuredItemArray.length === 0) {
```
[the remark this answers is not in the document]
avertrees: removing these lines breaks the app because these two lines are where that method setFeaturedItemArray() is invoked. is there another place to invoke setFeaturedItemArray(), like at the end of the utils file or something?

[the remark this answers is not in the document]
EdwinGuzman: Mentioned offline: let's leave this the way it is for now.

EdwinGuzman (approved):

src/utils/utils.ts
```diff
@@ -47,12 +122,44 @@ export const getNumItems = async (uuid: string) => {
...
+function addCommas(number) {
+  // Return the formatted number
+  return Number(number).toLocaleString("en-US");
+}
+
+function getTitleFromRepoAPIResponseData(data) {
+  // If titleInfo is an array of objects, use the title whose usage is primary. If titleInfo is a hash, return the title.
+  if (!data) {
+    return "";
+  } else {
```
EdwinGuzman, line 152: nitpick: you don't need the `else` (code style)

7emansell (approved):
Beautiful

PR #28, opened 2023-11-09
DR-2590 and DR-2584 Header and logo

7emansell:
## Ticket:

JIRA tickets [2584](https://jira.nypl.org/browse/DR-2584) and [2590](https://jira.nypl.org/browse/DR-2590)

- Adds the sticky header with placeholders for header components
- Adds the logo

## How has this been tested?

- Does not include component tests since they will likely be outdated once search is added

### Checklist:

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -2,7 +2,7 @@
 const nextConfig = {
   reactStrictMode: true,
   env: {
-    DC_URL: "https://qa-digitalcollections.nypl.org/",
+    DC_URL: "https://qa-digitalcollections.nypl.org",
   },
 };
 
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  env: {
+    DC_URL: "https://qa-digitalcollections.nypl.org/",
+  },
 };
 
 module.exports = nextConfig;
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -24,6 +24,7 @@ const Header = () => {
     <Box
       position="sticky"
       top={0}
+      py="s"
       zIndex={10}
       bgColor="var(--nypl-colors-ui-white)"
     >
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -21,7 +21,12 @@ const Header = () => {
     };
   }, []);
   return (
-    <Box position="sticky" top={0} zIndex={10} bgColor="white">
+    <Box
+      position="sticky"
+      top={0}
+      zIndex={10}
+      bgColor="var(--nypl-colors-ui-white)"
+    >
       <HStack
         justify="center"
         align="center"
@@ -33,7 +38,7 @@ const Header = () => {
         <HStack
           justify="space-between"
           align="center"
-          px="16px"
+          mx="16px"
           sx={{
             display: "flex",
             width: "1280px",
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -1,0 +1,15 @@
+import { HStack } from "@nypl/design-system-react-components";
+import React from "react";
+import DCLogo from "../logo/logo";
+
+const Header = () => {
+  return (
+    <>
+      <HStack ps="2" pe="2" justify="space-between">
+        <DCLogo />
+      </HStack>
+    </>
+  );
+};
+
+export default Header;
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -1,10 +1,10 @@
 import { Box, HStack, VStack } from "@nypl/design-system-react-components";
 import React from "react";
 import DCLogo from "../logo/logo";
-import { useScrolled } from "@/utils/useScrolled";
+import { useScrolledHeader } from "@/utils/useScrolledHeader";
 
 const Header = () => {
-  const isScrolled = useScrolled("header");
+  const isScrolled = useScrolledHeader();
   return (
     <Box
       position="sticky"
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -1,14 +1,35 @@
-import { HStack } from "@nypl/design-system-react-components";
+import { Box, HStack, VStack } from "@nypl/design-system-react-components";
 import React from "react";
 import DCLogo from "../logo/logo";
 
 const Header = () => {
   return (
-    <>
-      <HStack ps="2" pe="2" justify="space-between">
-        <DCLogo />
+    <Box position="sticky" top={0} zIndex={10} bgColor="white">
+      <HStack
+        justify="center"
+        align="center"
+        sx={{
+          display: "flex",
+          alignSelf: "stretch",
+        }}
+      >
+        <HStack
+          justify="space-between"
+          align="center"
+          px="16px"
+          sx={{
+            display: "flex",
+            width: "1280px",
+          }}
+        >
+          <DCLogo />
+          <VStack py="16px">
+            <p> I am the nav links! </p>
+            <p> I am the search bar and filter! </p>
+          </VStack>
+        </HStack>
       </HStack>
-    </>
+    </Box>
   );
 };
 
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -21,7 +21,13 @@ const Header = () => {
     };
   }, []);
   return (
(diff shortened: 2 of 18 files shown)


src/components/logo/logo.tsx
```diff
@@ -0,0 +1,64 @@
...
+            size="xxsmall"
+            aria-label="NYPL Lion Logo"
+          />
+          <VStack
+            align="left"
+            sx={{
+              display: { sm: "none", md: "none", lg: "block" },
+            }}
+          >
+            <Heading
```
7emansell, line 29: Should I be passing this as an SVG instead?

src/components/logo/logo.tsx
```diff
@@ -0,0 +1,64 @@
...
+  Logo,
+  Link,
+} from "@nypl/design-system-react-components";
+import React from "react";
+import appConfig from "appConfig";
+
+const DCLogo = () => {
+  return (
+    <>
+      <Link isUnderlined={false} href={`${appConfig.DC_URL}`}>
```
EdwinGuzman, line 14: This link doesn't work in Vercel. It may not be defined as an environment variable.

src/components/logo/logo.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <HStack>
+          <Logo
+            pt="2"
+            pb="2"
+            name="nyplLionBlack"
```
EdwinGuzman, line 19: There's an SVG type for the logo called `digitalCollectionsBlack` https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/story/components-media-icons-logo--with-controls&args=name:digitalCollectionsBlack. Did you try this and it didn't render properly? If possible, try to use this. 

Sorry, that should've been in the ticket. If it does work, you may need to update this so it renders only this `"nyplLionBlack"` logo on mobile and `"digitalCollectionsBlack"` on desktop.

EdwinGuzman (commented):
Some minor updates but there's a question around the logo types being used. I was wondering at first if this was missing tests but looking at this, you may not have much to write tests for just yet...

EdwinGuzman:
A heads up, in the next version of the DS (2.1.2), the padding around the Hero campaign inner content area is updated. This header aligns nicely with the Hero but that will likely change in the next update. 2.1.2 is already out if you want to test it now in this PR. This is also something to point out to Lily when it's done.

src/components/logo/logo.tsx
EdwinGuzman: I didn't see this comment while I was adding my review, but I added a comment above that hopefully answers this question.

src/components/logo/logo.tsx
7emansell: Do you think I need to add this as appConfig.DC_URL ? Because DC_URL is definitely in there so I'm perplexed as why these are no longer working (See more links on swim lane are also broken now)

src/components/logo/logo.tsx
7emansell: I should asked about this earlier– it renders but I couldn't find a way to make it large enough?

src/components/logo/logo.tsx
EdwinGuzman: I think this may also need to be defined in the `next.config.js` file https://nextjs.org/docs/pages/api-reference/next-config-js/env which is very similar to how `appConfig` is structured but for nextjs to include it in the bundle.

src/components/logo/logo.tsx
EdwinGuzman: There's a new `sizeBasedOn` prop for the `Logo` component that might help make it render bigger. But this new prop is in 2.1.2. You could try to override the size (width/height) with the `sx` prop if the new prop doesn't work.

7emansell: `sx` isn't overriding either unfortunately

src/components/logo/logo.tsx
```diff
@@ -0,0 +1,76 @@
...
+          }}
+        />
+        <Logo
+          name="digitalCollectionsBlack"
+          aria-label="NYPL Digital Collections Logo"
+          sx={{
+            display: { sm: "none", md: "none", lg: "block" },
+          }}
+        />
+        {/* <VStack
```
chrisb87, line 36: Is there a reason to keep this comment block?

7emansell: See above discussion! Figuring out how to override DS logo size

src/components/logo/logo.tsx
EdwinGuzman: If you install DS version 2.1.2, you can add this to the component
```
height="80px"
sizeBasedOn="height"
```

While this does help, it's not accurate so you would need to adjust the height value to get it closer to the figma file.

7emansell: Awesome, sorry for some reason didn't realize from your above comment that updating to 2.1.2 was a possible thing to do

next.config.js
```diff
@@ -1,6 +1,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  env: {
+    DC_URL: "https://qa-digitalcollections.nypl.org/",
```
EdwinGuzman, line 5: From what we chatted about in the other PR, remove the trailing slash here.

src/components/logo/logo.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <Link
+      isUnderlined={false}
+      aria-label={"Digital Collections Homepage"}
+      href={`${appConfig.DC_URL}`}
+    >
+      <HStack>
+        <Logo
+          name="nyplLionBlack"
+          sizeBasedOn="height"
+          height="40px"
```
EdwinGuzman, line 23: These are the values in Figma but it looks quite small to me.

src/components/header/header.tsx
```diff
@@ -0,0 +1,64 @@
+import { Box, HStack, VStack } from "@nypl/design-system-react-components";
+import React, { useEffect, useState } from "react";
+import DCLogo from "../logo/logo";
+
+const Header = () => {
+  const [isScrolled, setIsScrolled] = useState(true);
+
+  useEffect(() => {
+    const handleScroll = () => {
+      if (window.scrollY > 100) {
```
EdwinGuzman, line 10: How did you calculate the number 100?

src/components/header/header.tsx
```diff
@@ -0,0 +1,64 @@
...
+        setIsScrolled(true);
+      }
+    };
+
+    window.addEventListener("scroll", handleScroll);
+
+    return () => {
+      window.removeEventListener("scroll", handleScroll);
+    };
+  }, []);
```
EdwinGuzman, line 22: Lines 6-22 can be a custom hook. It's not necessary because I don't think you will need to implement this in any other component. But, it _could_ be reusable in the future and added to the DS so other apps can implement a consistent sticky behavior. This also depends on the `100` number and how/why it was chosen.

src/components/header/header.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <Box
+      position="sticky"
+      top={0}
+      py={{ sm: "xs", md: "s" }}
+      zIndex={10}
```
EdwinGuzman, line 28: Some DS components also have a z-index and the 10 value may be too low that other components will "bleed through". This is also a "magic number" but 9999 is big enough to work and always be on top of other elements on a page.

EdwinGuzman (commented):
This looks good. Can you verify with Lana/Lily that the notification banner does _not_ need to be sticky as well? Unless it's already somewhere and I missed it.

src/components/logo/logo.tsx
7emansell: Same, I'll make sure to check in with Lily at VQA

src/components/header/header.tsx
7emansell: Vibes

7emansell: Custom hook uses `offsetHeight` which seems more legitimate but open to discuss

7emansell:
In Figma, notification banner is not visible on the "scrolled" header state. Will also double check with Lily at VQA

src/components/header/header.tsx
7emansell: 9999 is actually so high it covers error messages from React so let me know if 999 is okay

EdwinGuzman (approved):
This looks good and added some minor comments but they are not blockers.

The biggest concern is about the link and logo aria-labels but that's something for Clare to verify, so please do request an accessibility review from her.

avertrees (approved):

PR #30, opened 2023-11-14
DR-2543/campaign hero number of digitized items

avertrees:
## Ticket:

- JIRA ticket [DR-2543](https://jira.nypl.org/browse/DR-2543)

- Display the total number of available digitized items on DC
- Should display within the hero title
- Use the same typography as the rest of the title
- Updates on a regular schedule (dev can suggest the best cadence and update this)

### To verify:
- Should include location restricted materials as well as the ones available anywhere

-

### Checklist:

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -30,6 +30,17 @@ export const imageURL = (
 };
 
 /**
+ * Returns the number of digitized items in repo api.
+ */
+
+export const getNumDigitizedItems = async () => {
+  const apiUrl = `${process.env.API_URL}/api/v2/items/total`;
+  const res = await apiCall(apiUrl);
+  return addCommas(res.count.$) || 0;
+  // return apiCall(apiUrl);
+};
+
+/**
  * Returns the total number of items in the collection.
  * @param {string} uuid - the collection ID
  */
@@ -78,6 +89,9 @@ export const apiCall = async (apiUrl: string) => {
     if (response.status === 200) {
       const data = await response.json();
       const responseData = data.nyplAPI.response;
+      // if (apiUrl.includes('api/v2/items/total')) {
+      //   console.log("responseData for api/v2/items/total", responseData)
+      // }
       return responseData;
     } else {
       return 0;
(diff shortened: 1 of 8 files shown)


src/utils/utils.ts
```diff
@@ -47,12 +88,30 @@ export const getNumItems = async (uuid: string) => {
...
     return 0;
   }
 };
+
+function addCommas(number) {
+  // Convert the number to a string
+  let numberString = number.toString();
+
+  // Use a regular expression to add commas to the string
+  numberString = numberString.replace(/\B(?=(\d{3})+(?!\d))/g, ",");
```
EdwinGuzman, line 109: There's a native way to do this with `.toLocalString("en-US")` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

This should be called on a number though, so you don't need to convert it to a string to call the function on the number, and it'll also return a string. Will the API always return a number for `res.count.$`? If so, then this is safe to use, but if it _might_ return a string then you'll have to add some safety checks to convert the returned value before using `number.toLocalString("en-US")`.

src/utils/utils.ts
```diff
@@ -47,12 +88,30 @@ export const getNumItems = async (uuid: string) => {
...
+
+  // Return the formatted number
+  return numberString;
+}
+
+// Example usage:
+let numberWithCommas = addCommas(1234567);
```
EdwinGuzman, line 116: This would be great as tests.

src/utils/utils.ts
```diff
@@ -29,15 +29,56 @@ export const imageURL = (
...
 };
 
+/**
+ * Returns the number of digitized items in repo api.
+ */
+
+export const getNumDigitizedItems = async () => {
+  const apiUrl = `${process.env.API_URL}/api/v2/items/total`;
+  const res = await apiCall(apiUrl);
+  return addCommas(res.count.$) || 0;
```
EdwinGuzman, line 39: If there is an error with the Repo API, the fallback will render as "explore 0 items digitized ...". What if the fallback is `863848`? (But set before it's passed to `addCommas`, e.g. `const fallbackCount = 863848; const totalItems = res.count.$ || fallbackCount;`)

src/utils/utils.ts
EdwinGuzman: The link shows an example usage in Arabic. When/if this app gets localization, maybe we can use the language code for every translation, e.g. `x.toLocaleString('ar-EG');`. Trying to forward think for an accessibility feature.

PR #34, opened 2023-11-15
DR-2592 Featured Content

7emansell:
## Ticket:

Fixes JIRA ticket [2592](https://jira.nypl.org/browse/DR-2592)

- Adds the FeaturedContent component 
- Adds the static images that appear in the component
- Refactors swim lanes for correct spacing

## How has this been tested?

Tested locally.

### Checklist:


src/components/swimlanes/swimLanes.tsx
```diff
@@ -11,94 +11,95 @@ import {
...
-            <Link
-              id={`row-see-more-${lane.slug}`}
-              type="standalone"
-              href={`${appConfig.DC_URL}/collections/lane/${lane.slug}`}
-              aria-label={`See more ${lane.title.toLowerCase()}`}
-              sx={{
-                display: { sm: "none", md: "inline" },
-                fontWeight: "500",
-                alignItems: "center",
+  let hasRendered = false;
```
EdwinGuzman, line 19: What if you tried inserting the `FeaturedContent` component into the `swimLanesElements` array through the `.splice` function? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

If that works, you could move the array insertion outside of this component so that `swimLanes.tsx` only returns swimlanes and not the featured content.

src/data/featuredContentData.ts
```diff
@@ -0,0 +1,28 @@
+const featuredContentData = {
+  data: [
```
EdwinGuzman, line 2: You could make `featuredContentData` the array itself so you can call `featuredContentData[randomNumber]` instead of first checking for the `.data` property. Unless if you added this property for another reason?

src/components/featuredContent/featuredContent.tsx
```diff
@@ -0,0 +1,40 @@
...
+  FeaturedContent,
+  Button,
+  Heading,
+} from "@nypl/design-system-react-components";
+import React from "react";
+import featuredContentData from "@/data/featuredContentData";
+
+const FeaturedContentComponent = () => {
+  var randomNumber = Math.random() < 0.5 ? 0 : 1;
+  var data = featuredContentData.data[randomNumber];
```
EdwinGuzman, line 11: Use `let` or `const`, not `var`.

src/components/featuredContent/featuredContent.tsx
```diff
@@ -0,0 +1,40 @@
+import {
+  FeaturedContent,
+  Button,
+  Heading,
+} from "@nypl/design-system-react-components";
+import React from "react";
+import featuredContentData from "@/data/featuredContentData";
+
+const FeaturedContentComponent = () => {
+  var randomNumber = Math.random() < 0.5 ? 0 : 1;
```
EdwinGuzman, line 10: It'd be good to have a few simple tests for this component but this random function will make it tricky to test. You can add a `{ testRandomNumber }` prop (for example), and use that in your test as `render(<FeaturedContent testRandomNumber={0} />)`.

And here you can check if `testRandomNumber` was passed and if not then run the code you have to generate the random number. It's a somewhat typical pattern to pass test-only values or props down to functions or components when you need it to be a certain value only for testing purposes.

EdwinGuzman (commented):
Some updates. There's a visual issue with an image that seems to get cropped off but I'm not sure how big of a deal that is:
Screen Shot 2023-11-16 at 10 24 02 AM

avertrees (approved):
Looks good to me!

src/components/featuredContent/featuredContent.test.tsx
```diff
@@ -0,0 +1,51 @@
+import { render, screen, waitFor, within } from "@testing-library/react";
+import FeaturedContentComponent from "./featuredContent";
+
+describe("Featured Content component renders with expected props", () => {
+  it("renders Digital Collections store content", async () => {
```
EdwinGuzman, line 5: The description is the same as the next test. This can be for "public domain" content.

src/components/featuredContent/featuredContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+  Button,
+  Heading,
+} from "@nypl/design-system-react-components";
+import React from "react";
+import featuredContentData from "@/data/featuredContentData";
+import { FeaturedContentData } from "@/types/FeaturedContentData";
+
+const FeaturedContentComponent = ({ testRandomNumber }) => {
+  const randomNumber =
+    testRandomNumber !== undefined
```
EdwinGuzman, line 12: You can update this to `testRandomNumber || Math.floor(Math.random() * 2);`. Minor comment but this works just the same.

src/components/featuredContent/featuredContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+      data-testid={`featured-content-${randomNumber}`}
+      isFullWidth={true}
+      textContent={
+        <div>
+          <Heading level="h2" overline={data.overline}>
+            {data.heading}
+          </Heading>
+          <p>{data.text}</p>
+
+          <a href={data.link} target="_blank" rel="noopener noreferrer">
```
EdwinGuzman, line 27: If you use a DS `Link` component, you can make it look like a button: https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/story/components-navigation-link--with-controls&args=type:buttonPrimary

That's better for accessibility. The general rule is links navigate to a new page and buttons are interactive and update the current page. In this case, we want to go to a new page but the UX design is for a button, so we have to reconcile that.

src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import FeaturedContentComponent from "../featuredContent/featuredContent";
+import SwimLanes from "../swimlanes/swimLanes";
+
+const HomePageMainContent = ({ testRandomNumber, lanesWithNumItems }) => {
```
EdwinGuzman, line 5: 👍

src/components/swimlanes/swimLanes.tsx
```diff
@@ -11,94 +11,89 @@ import {
...
-            <Link
-              id={`row-see-more-${lane.slug}`}
-              type="standalone"
-              href={`${appConfig.DC_URL}/collections/lane/${lane.slug}`}
-              aria-label={`See more ${lane.title.toLowerCase()}`}
-              sx={{
-                display: { sm: "none", md: "inline" },
-                fontWeight: "500",
-                alignItems: "center",
+  return lanesWithNumItems.map((lane, key) => [
```
EdwinGuzman, line 18: This refactor is great. You shouldn't need to wrap the output in an array and can return the component directly such as

```
return lanesWithNumItms.map((lane, key) => (
  <Box ...
    ...
  </Box>
));
```

`SwimLanes` is now returning an array of arrays which React knows how to render but we can simplify it.

src/data/featuredContentData.ts
```diff
@@ -0,0 +1,26 @@
+const featuredContentData = [
```
EdwinGuzman, line 1: You can assign this the created type `const featuredContentData: FeaturedContentData[] = [ ...`.

EdwinGuzman (commented):
Update looks good with a few minor comments. The biggest part that needs an update is the button wrapped in a link.

src/components/featuredContent/featuredContent.tsx
7emansell: Okay and this drove me crazy bc it does appear to be the exact same thing, but with just the OR operator it fails to render the case where `testRandomNumber` is 0, because OR will take 0 as falsy. This makes me want to make the options 1 or 2 instead of 0 or 1, but it's indexing into an array, so that feels wrong. Open to ideas though

EdwinGuzman: Oh, that's right that we _want_ 0 as a value. That's my mistake. The original is fine.

EdwinGuzman (approved):

EdwinGuzman (approved):

PR #36, opened 2023-11-21
DR-2586 and DR-2648 Search bar with filter

7emansell:
## Ticket:

Fixes JIRA tickets [DR-2586](https://jira.nypl.org/browse/DR-2586) and [DR-2648](https://jira.nypl.org/browse/DR-2648).

## This PR does the following:
- Adds search bar to header
- Adds public domain filter for search
- Adds component tests for search bar and filter

### Checklist:

- [x] I have added relevant accessibility documentation for this pull request.
- [x] All new and existing tests passed.

PR #38, opened 2023-12-04
DR-2586 and DR-2648 Redesigned search bar with filter

7emansell:
## Ticket:

Fixes JIRA tickets [DR-2586](https://jira.nypl.org/browse/DR-2586) and [DR-2648](https://jira.nypl.org/browse/DR-2648).

- Changes breakpoints on search bar 
- See https://github.com/NYPL/digital-collections/pull/36 for previous work on search bar/filter

### Checklist:

### DIFF
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -9,6 +9,7 @@ import { useScrolled } from "@/hooks/useScrolled";
 import { useStickyMargin } from "@/hooks/useStickyMargin";
 import Search from "../search/search";
 import DCLogo from "../logo/logo";
+import { headerBreakpoints } from "@/utils/breakpoints";
 
 const Header = () => {
   useStickyMargin();
@@ -29,10 +30,10 @@ const Header = () => {
         mx="auto"
         padding="0 16px"
         sx={{
-          "@media screen and (min-width: 1024px)": {
+          [`@media screen and (min-width: ${headerBreakpoints.lgTablet})`]: {
             display: "block",
           },
-          "@media screen and (min-width: 1280px)": {
+          [`@media screen and (min-width: ${headerBreakpoints.desktop})`]: {
             display: "flex",
             justifyContent: "space-between",
           },
@@ -42,8 +43,9 @@ const Header = () => {
         <HStack
           justify="space-between"
           sx={{
-            "@media screen and (max-width: 768px)": {
-              display: "none",
+            display: "none",
+            [`@media screen and (min-width: ${headerBreakpoints.smTablet})`]: {
+              display: "flex",
             },
           }}
         >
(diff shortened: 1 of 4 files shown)


src/components/header/header.tsx
```diff
@@ -20,57 +20,76 @@ const Header = () => {
...
+      pt="xs"
     >
       <Box
-        alignItems="center"
         maxWidth="1280px"
-        padding={{ md: "0 15px" }}
         mx="auto"
+        padding="0 16px"
+        sx={{
+          "@media screen and (min-width: 1024px)": {
```
EdwinGuzman, line 32: Can you define the values as variables to not have magic numbers and to more easily see that, for example, 1024px is "large tablet"? It _might_ make it easier to update these queries in the future once they are set in the DS as well.

src/components/header/header.tsx
```diff
@@ -20,57 +20,76 @@ const Header = () => {
...
+            display: "flex",
+            justifyContent: "space-between",
+          },
+          alignItems: "center",
+        }}
       >
         <HStack
+          justify="space-between"
           sx={{
+            "@media screen and (max-width: 768px)": {
```
EdwinGuzman, line 45: Can you rewrite this using `min-width` instead? `min-width` is preferred for a mobile-first approach and it also aligns with Chakra's responsive design pattern.

If it's more trouble than it's worth, then leave it as max-width and document what makes it difficult to use `min-width`.

src/components/header/header.tsx
```diff
@@ -20,57 +20,76 @@ const Header = () => {
...
+              display: "none",
+            },
+          }}
+        >
+          <DCLogo isMobile={false} />
+          <Box
+            sx={{
+              display: "none",
+              "@media screen and (max-width: 1280px)": {
```
EdwinGuzman, line 54: Noting another instance of max-width that should be written as min-width if possible.

src/components/search/search.tsx
```diff
@@ -1,10 +1,12 @@
 import React, { useState } from "react";
-import { SearchBar } from "@nypl/design-system-react-components";
+import { Box, SearchBar } from "@nypl/design-system-react-components";
 import { useRouter } from "next/router";
 import PublicDomainFilter from "../publicDomainFilter/publicDomainFilter";
 import appConfig from "appConfig";
+import { useScrolled } from "@/hooks/useScrolled";
 
 const Search = () => {
+  const isScrolled = useScrolled("header");
```
EdwinGuzman, line 9: I don't think you're using this value in this file.

src/components/search/search.tsx
```diff
@@ -29,20 +31,36 @@ const Search = () => {
...
-          name: "textInputName",
-          onChange: handleTextChange,
-          value: keywords,
-          placeholder: "Search keyword(s)",
+      <Box
+        id="search-wrapper"
+        sx={{
+          alignItems: "start",
+          width: "100%",
+          "@media screen and (max-width: 768px)": {
```
EdwinGuzman, line 39: Noting another instance of max-width that should be written as min-width if possible.

EdwinGuzman (changes requested):
This works well but commented on two technical updates that don't affect the overall feature but just how it's written.

For the 767-769px range, there are three different views. Both 767px and 769px are correct, but when it's 768px, the logo and hamburger menu don't display. I _think_ if you first try to update the `max-width` to `min-width`, this might be resolved. Otherwise, it could be resolved by adding or subtracting `1` to the media query (depending on max or min).

Try the "iPad mini" view:

Screen Shot 2023-12-05 at 1 53 31 PM

EdwinGuzman (approved):
Great, looks good!

PR #39, opened 2023-12-06
DR-2630 Nav menu

7emansell:
## Ticket:

Fixes JIRA ticket [DR-2630](https://jira.nypl.org/browse/DR-2630)

- Adds the desktop width nav links
- Adds hamburger menu
- Adds mobile nav links on hamburger open

### Checklist:


src/components/navMenu/mobileNavMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { Box, chakra } from "@chakra-ui/react";
+import { useScrolled } from "@/hooks/useScrolled";
+
+const listItems = dcNavLinks.map(({ href, text }) => (
+  <Link
+    isUnderlined={false}
+    href={href}
+    key={text}
+    aria-label={text}
+    sx={{
```
7emansell, line 21: Not using a `<Text>` element here because it throws the focus ref onto the link, still looking at this

src/components/navMenu/mobileNavMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+            id="mobile-menu"
+            sx={{
+              display: "flex",
+              flexDirection: "column",
+              alignItems: "left",
+              position: "fixed",
+              width: "100%",
+              paddingTop: "m",
+              paddingLeft: "xs",
+              marginTop: isScrolled ? "4px" : "-36px",
```
7emansell, line 86: Added an option to remove the delay from the scroll stickiness (`(useScrolled("header", false/true)` ) so that this could change immediately but even still there's a tiny but noticeable delay before the menu snaps into place under the header. Would love any ideas

7emansell: Also this shouldn't be hardcoded- I think the solution will probably have to do with `getBoundingClientRect()` but still working on how to incorporate that separately from `useScrolled`

EdwinGuzman: Is this question related to the gap that appears until you scroll past a certain point?
Screen Shot 2023-12-06 at 3 58 25 PM

src/components/navMenu/mobileNavMenu.tsx
EdwinGuzman: I'm a bit confused, did you need to use a `Text` component before? Was it because the link wrapped the `Text` component? That probably throws off the focus and outline.

Sorry, we did not chat about the structure of the nav lists before. They should be structured as
```
<nav aria-label="..."> //every nav on the page should have a unique aria-label
  <ul>
    <li><a ...>...</a></li>
    <li><a ...>...</a></li>
    ...
  </ul>
</nav>
```

Try this instead of the stack components. To get the focus ref align properly, the "li"s can be "empty" (no styles) and the anchor elements should have the padding to expand to the right size.

src/components/navMenu/mobileNavMenu.tsx
EdwinGuzman: This worked for me but would like to know if this is how you expect it to work. Update position to absolute and keep marginTop to 4px.

```
position: "absolute",
width: "100%",
paddingTop: "m",
paddingLeft: "xs",
marginTop: "4px",
```

The margin is needed to keep it aligned and spaced and this works when there is no notification text or when there's long notification text as you scroll.

Screen Shot 2023-12-06 at 8 22 25 PM

7emansell: Oh man

avertrees (approved):

src/components/header/header.tsx
```diff
@@ -3,74 +3,115 @@ import {
...
+            [`@media screen and (min-width: ${headerBreakpoints.smTablet})`]: {
+              display: "none",
+            },
+            justifyContent: "space-between",
+          }}
+        >
+          <DCLogo isMobile={true} />
+          <MobileNavMenu />
         </HStack>
+        <Divider
```
EdwinGuzman, line 83: The DS has a [`HorizontalRule`](https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/docs/components-page-layout-horizontalrule--docs). Does that work? Mostly so you don't have to import from `"@chakra-ui/react"`.

I think this would be better as a bottom border but the DOM structure doesn't really allow it.

src/components/navMenu/mobileNavMenu.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const menuItems = screen.queryAllByText("Items");
+    expect(menuItems).toHaveLength(0);
+  });
+
+  it("has links", () => {
+    render(<MobileNavMenu />);
+    fireEvent.click(screen.getByLabelText("Open Navigation"));
+    expect(screen.getByLabelText("Items")).toHaveAttribute(
+      "href",
+      appConfig.DC_URL + `/search/index?utf8=%E2%9C%93&keywords=`
```
EdwinGuzman, line 29: Nitpick, include the variable in the string interpolation 
```
`${appConfig.DC_URL}/search/index?utf8=%E2%9C%93&keywords=`
```
Same for the others below.

src/components/navMenu/mobileNavMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+      width: "100%",
+      ":hover": {
+        textDecoration: "none",
+      },
+      ":visited": {
+        color: "ui.black",
+      },
+    }}
+  >
+    <Text size="subtitle2" sx={{ marginBottom: "unset" }}>
```
EdwinGuzman, line 34: `p` elements should not render inside `a` elements so you can remove the `Text` component and move the styles to the `Link` component.

src/components/navMenu/mobileNavMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+  Text,
+  Link,
+  Icon,
+  Button,
+  useCloseDropDown,
+  List,
+} from "@nypl/design-system-react-components";
+import React, { useRef, useState } from "react";
+import FocusLock from "@chakra-ui/focus-lock";
+import { Box, chakra } from "@chakra-ui/react";
```
EdwinGuzman, line 12: You can import `Box` from the DS. Why did you need to wrap the `MobileNavMenu` component in the `chakra` factory function?

src/components/navMenu/mobileNavMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+            },
+            "not:([disabled]):focus": {
+              outline: "none",
+            },
+          }}
+        >
+          <Icon name={isOpen ? "close" : "utilityHamburger"} size="large" />
+        </Button>
+        {isOpen && (
+          <nav
```
EdwinGuzman, line 73: Add an `aria-label` of "Mobile header links"

src/components/navMenu/mobileNavMenu.tsx
```diff
@@ -0,0 +1,101 @@
...
+  useCloseDropDown,
+  List,
+} from "@nypl/design-system-react-components";
+import React, { useRef, useState } from "react";
+import FocusLock from "@chakra-ui/focus-lock";
+import { Box, chakra } from "@chakra-ui/react";
+import { useScrolled } from "@/hooks/useScrolled";
+
+const listItems = dcNavLinks.map(({ href, text }) => (
+  <Link
```
EdwinGuzman, line 16: For lists, the `ul` or `ol` elements should have `li` children. In this case, wrap the `Link` components in `<li>`s.

src/components/navMenu/navMenu.tsx
```diff
@@ -0,0 +1,27 @@
+import { dcNavLinks } from "@/data/dcNavLinks";
+import { NavMenuProps } from "@/types/NavMenuProps";
+import { HStack, Text, Link } from "@nypl/design-system-react-components";
+import React from "react";
+
+const NavMenu = () => {
+  const listItems = dcNavLinks.map(({ href, text }) => (
+    <Link isUnderlined={false} href={href} key={text} aria-label={text}>
+      <Text mb="0px" size="subtitle2" color="black">
```
EdwinGuzman, line 9: This has a similar element issue as described in the `MobileNavMenu` links.

src/hooks/useScrolled.ts
```diff
@@ -1,12 +1,15 @@
...
 
 /* Returns if the page is scrolled past the header, could be refactored for other elements. */
-export function useScrolled(elementId) {
+export function useScrolled(elementId, delay) {
   const [isScrolled, setIsScrolled] = useState(true);
 
   useEffect(() => {
-    let DELAY_STICKY_TRANSITION = window.innerWidth < 600 ? 700 : 300;
-
+    let DELAY_STICKY_TRANSITION = delay
```
EdwinGuzman, line 8: This is fine but do you still need the delay?

src/components/navMenu/navMenu.tsx
```diff
@@ -0,0 +1,27 @@
+import { dcNavLinks } from "@/data/dcNavLinks";
+import { NavMenuProps } from "@/types/NavMenuProps";
```
EdwinGuzman, line 2: You're not using `NavMenuProps` in this file. It's also okay to keep interfaces or types within the same file as the components that use them unless if the interface or type is generic.

src/data/dcNavLinks.ts
```diff
@@ -0,0 +1,20 @@
+import appConfig from "appConfig";
+
+export const dcNavLinks = [
+  {
+    href: appConfig.DC_URL + `/search/index?utf8=%E2%9C%93&keywords=`,
```
EdwinGuzman, line 5: Nitpick: include the variable in the string interpolation.

EdwinGuzman (changes requested):
This looks and works well. I added comments for some changes, mostly around the DOM structure for the lists.

src/components/navMenu/mobileNavMenu.tsx
7emansell: Holdover from first attempt at focus lock

src/components/navMenu/mobileNavMenu.tsx
7emansell: Passing this to `listitems` does this automatically, so that creates nested `<li>` elements unnecessarily

EdwinGuzman: Yes, I forgot that's how that works! :face-palm: (there's no face palm emoji)

src/components/navMenu/navMenu.tsx
```diff
@@ -0,0 +1,51 @@
...
+          textDecoration: "none !important",
+        },
+      }}
+    >
+      {text}
+    </Link>
+  ));
+
+  return (
+    <nav
```
EdwinGuzman, line 30: Add an aria-label here.

EdwinGuzman (approved):
Update looks good, just missing one aria-label for the nav menu component.

PR #46, opened 2023-12-21
DR-2595 implement feedback button and form

avertrees:
## Ticket:

- JIRA ticket [DR-2595](https://jira.nypl.org/browse/DR-2595)

- Adds the feedback button to the home page and submits content to new spreadsheet: https://docs.google.com/spreadsheets/d/1H-noI-R9iGXOXlhB0CrwKJBrII37pKazLUczI19C4Dw/edit#gid=0

## How has this been tested?

locally

-

### Checklist:


src/pages/_app.tsx
```diff
@@ -1,16 +1,40 @@
...
+
+const onSubmit = async (values) => {
+  const response = await fetch("/api/feedback", {
+    method: "POST",
+    headers: {
+      Accept: "application/json",
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify(values),
+  });
```
EdwinGuzman, line 18: I think you may need to move the `onSubmit` function inside of `App` to implement error handling: https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/docs/components-form-elements-feedbackbox--docs#error

This doesn't work on Vercel but it does locally. I _think_ it'll work on the QA instance once it's deploy, but until then, we should render the error message to patrons.

src/pages/_app.tsx
```diff
@@ -1,16 +1,40 @@
...
+};
 
 export default function App({ Component, pageProps }: AppProps) {
+  const { onOpen, isOpen, onClose, FeedbackBox } = useFeedbackBox();
```
EdwinGuzman, line 22: You can use the `FeedbackBox` component directly from the DS instead of the `useFeedbackBox` hook. The `onOpen`, `isOpen`, and `onClose` functions are not being used in a custom way in DC so it's okay to use the component directly. If you think that at some point we'll have a link that opens the form, then this should be kept. 

Here's an example: https://www.nypl.org/research/research-catalog/bib/b10833141
Click on the "contact a librarian" links and those will open the form. If DC will want/need similar functionality, then keep the `useFeedbackBox` hook.

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,100 @@
...
+  const { method } = request;
+  if (method !== "POST") {
+    return response
+      .status(405)
+      .send({ message: "only POST requests are allowed" });
+  }
+
+  // Values to pass to spreadsheet
+  const body = request.body;
+  const type = body.category;
```
EdwinGuzman, line 19: Nitpick: you can destructure this as `const { type, comment: feedbackText } = body;`. The `comment: feedbackText` is getting the property from `body` and also assigning it to a new variable `feedbackText`. This is not needed but just showing another way to do this.

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,100 @@
...
+  const feedbackText = body.comment;
+  // timestamp: date
+  const currentDate = new Date();
+  const currentDayOfMonth = currentDate.getDate();
+  const currentMonth = currentDate.getMonth(); // Be careful! January is 0, not 1
```
EdwinGuzman, line 24: Is this the reverse in ruby?

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,100 @@
...
+  const currentYear = currentDate.getFullYear();
+  const dateString =
+    currentYear + "-" + (currentMonth + 1) + "-" + currentDayOfMonth;
+  // timestamp: time
+  const hh = currentDate.getUTCHours();
+  const mm = currentDate.getUTCMinutes();
+  const ss = currentDate.getSeconds();
+  const timeString = hh + ":" + mm + ":" + ss;
+  const timestamp = dateString + " " + timeString;
```
EdwinGuzman, line 33: The timestamp code could extracted out into a utility function such as `getTimestamp`. It seems it's also a custom timestamp for DC that is more readable for staff to read when viewing feedback? Maybe `getCustomTimestamp`?

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,100 @@
...
+            userVersion,
+          ],
+        ],
+      },
+    });
+
+    console.log("googleResponse is:");
+    console.log(googleResponse);
+    return response.status(200).json({
+      data: googleResponse.data,
```
EdwinGuzman, line 93: Do you want to return this data to the client call? Maybe an object with `{ text: "successful form submission" }` is enough because you're not doing anything with this returned data in the client call.

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,100 @@
...
+    });
+  } catch (e) {
+    console.log("error: ", e);
+    return response.status(500).send({ message: "something went wrong" });
```
EdwinGuzman, line 97: Continuing the comment above, `message` is a better property name than `text`, and also add in the message that something went wrong submitting/sending the form values to the google spreadsheet. This will be logged so a message that will help you debug when something goes wrong.

EdwinGuzman (changes requested):
Added feedback.

I know it doesn't work on vercel right now but I would err on the side of pushing this onto QA AWS and testing it there. That will be a better gauge of whether this works or not because the production site will be on AWS and not on vercel. This works for us locally so it's a Vercel configuration issue which is something we should fix, but it's not crucial.

src/pages/api/feedback.ts
avertrees: what do you mean? like month|date|year?

EdwinGuzman: I was asking if January is indexed 1 in ruby's date object.

avertrees: Ruby has a Time object that does the formatting out of the box https://ruby-doc.org/core-2.6.3/Time.html

src/pages/_app.tsx
```diff
@@ -1,16 +1,57 @@
...
       <Head>
         <meta name="viewport" content="width=device-width, initial-scale=1" />
       </Head>
       <DSProvider>
         <Component {...pageProps} />
+        <FeedbackBox
+          showCategoryField
+          onSubmit={onSubmit}
+          isOpen={isOpen}
+          onClose={onClose}
```
EdwinGuzman, line 50: You need a custom onClose function that calls `onClose()` internally as well as `setView("form")` to reset the form after you make a submission and the form is closed. Right now, when you open it after submitting feedback, it will render the previous confirmation or error view which is now what we want.

src/pages/_app.tsx
```diff
@@ -1,16 +1,57 @@
...
+  DSProvider,
+  useFeedbackBox,
+} from "@nypl/design-system-react-components";
+import React from "react";
 
 export default function App({ Component, pageProps }: AppProps) {
+  const [view, setView] = React.useState("form");
+  const { onOpen, isOpen, onClose, FeedbackBox } = useFeedbackBox();
+  const onSubmit = async (values) => {
+    const response = await fetch("/api/feedback", {
```
EdwinGuzman, line 14: Do either async/await or the then/catch syntax but not both. Either is fine but the async/await syntax has become more conventional (callback functions can be annoying).

With async/await, wrap this in a try/catch statement:
```
    try {
      const response = await fetch("/api/feedback", {
        method: "POST",
        headers: {
          Accept: "application/json",
          "Content-Type": "application/json",
        },
        body: JSON.stringify(values),
      });

      console.log("response is: ", response);

      if (response.ok) {
        // ...
        setView("confirmation");
      } else {
        setView("error");
      }
    } catch (error) {
      // Reject the promise according to your application.
      // And then call:
      console.log("error is: ", error);
      setView("error");
    }
```

EdwinGuzman (commented):
Two updates should be made. What is interesting is that the feedback form seems to close and then reopen when submitting it. This doesn't happen elsewhere and I can't figure out why it's happening here... I tried a few approaches but keeps flickering close/open when a submission is made. I can take a better look next week.

7emansell (approved):

EdwinGuzman (approved):

PR #49, opened 2024-01-04
first "release" to production

avertrees:
deploy qa to production as v0.1.0


appConfig.ts
```diff
@@ -0,0 +1,5 @@
+const appConfig = {
+  DC_URL: process.env.DC_URL,
```
EdwinGuzman, line 2: You might have to add `API_URL` and `AUTH_TOKEN` here.

.env.example
```diff
@@ -1,5 +1,6 @@
 # Repo api:
-API_URL=
-AUTH_TOKEN=
+API_URL= https://api.repo.nypl.org
+DC_URL=https://qa-digitalcollections.nypl.org
+AUTH_TOKEN= # Get credentials from a teammate!
```
EdwinGuzman, line 4: I think you need to define these in the Dockerfile: https://github.com/NYPL/digital-collections/blob/main/Dockerfile#L18C4-L18C4

I think task definition passes the env var down when docker is building the app but Docker needs to read and pass it down to the app. You can try without it first and then with it.

EdwinGuzman (approved):
Approving. I think you need to make the env variable updates but the way to find out may be by deploying this first and finding it works or not.

appConfig.ts
avertrees: is this file only used in the production environment?

EdwinGuzman: It's used in all environments.

CHANGELOG.md
```diff
@@ -7,16 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Updated featured item link in Hero Component to change on reload using the featured item imageID. (DR-2547)
+- Updated travis deployer keys after rotation (DR-2722)
 
 ### Added
+
 - Added spider block script (DR-2518)
 - Created Dockerfile and docker-compose.yml files. (DR-2446)
 - Created Next.js 13 app using NYPL Reservoir and TS 4.9.5. (DR-2444)
 - Set up travis script to build and deploy to QA (DR-2448)
+- Added global footer (DR-2591)
```
7emansell, line 30: I haven't been adding to the changelog... should I add in all my tickets before this goes to prod

7emansell (approved):
Approved but ditto on env variables, and let me know about the changelog

CHANGELOG.md
avertrees: yes lol you can add them to this release branch or make a new branch and make a PR to merge into main, I don't have a preference

PR #50, opened 2024-01-05
DR-2715 Logging to cloudwatch

7emansell:
## Ticket:

- JIRA ticket [DR-2715](https://jira.nypl.org/browse/DR-2715)

- Adds the winston logger (serverside logs)
- Creates folder for log files

## How has this been tested?

Locally, in Docker

### Checklist:


7emansell:
Some notes: 
1. This is very much a starting point– if we want the logs to be printed differently, or stored somewhere else, that's easy to change now. Leaving the WIP label until we're happy with that configuration
2. Noting that these are only serverside logs, so only errors from `getServerSideProps` and API routes can use this logger anyways
3. Don't know if it's within the scope of the ticket to go through and add `logger.error(..)` in catch blocks, but again, very few places where we can do that anyways?
4. Also, no way to know for sure that this will work with Cloudwatch until it's deployed to QA, but since it logs to Docker just fine, I assume it will

EdwinGuzman (approved):
This looks good for a starting point like you said and that we should deploy this to QA to verify the logs in Cloudwatch.

I was expecting only server log messages for this feature and Cloudwatch. I think it's worth adding an error log in the `apiCall` function catch block. This is a good point that not all await functions have proper error handling but that work (along with adding more error logs) should be separate.

EdwinGuzman:
@7emansell yes.

7emansell:
@EdwinGuzman It cannot go there as far as I understand, only in `getServerSideProps`

EdwinGuzman:
That's not what I expected but I understand. Let's move forward with this and we can rework the error handling so we can log messages through `getServerSideProps`.

avertrees (approved):

avertrees:
looks good to me! one thought - we might want to add dc.log to .gitignore?

PR #53, opened 2024-01-12
DON'T MERGE: For performance testing: comment out any calls to repo api and return default

avertrees:
Keeping this here to compare with skeleton loading


src/utils/utils.ts
```diff
@@ -133,10 +134,14 @@ export const apiCall = async (apiUrl: string) => {
 
 export const getItemDataFromImageID = async (imageID: string) => {
   const apiUri = await getAPIUri("local_image_id", imageID);
-  const data = await apiCall(apiUri.apiUri);
+  // const data = await apiCall(apiUri.apiUri);
```
EdwinGuzman, line 137: This still makes a call to `getAPIUri`, but I think it's minor considering that most API calls have been replaced.

EdwinGuzman (commented):
This example looks good. This _seems_ and feels faster than the main branch (or qa/prod) that use API calls... but is it enough to say that it's all the API calls that make the app slow?

EdwinGuzman:
Can you try using images.nypl.org instead of iiif?

EdwinGuzman:
With the logs in the `main` branch now, this PR should be updated and additional logs would be helpful to compare loading times for the version making API  calls and this version.

PR #55, opened 2024-01-19
DR-2715: Fixing vercel logging error

7emansell:
## Ticket:

Fixes JIRA ticket [DR-2715](https://jira.nypl.org/browse/DR-2715) (particularly its Vercel preview).

- Creates an API route (serverless) that allows Vercel to create the log directory at runtime
- Log directory does not persist
- Needs refactoring since the endpoint is hardcoded

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.1.0] 2024-01-08
+
 ### Created
 
 - Created a campaign hero component and other child components for the featured item data. (DR-2464)
+- Created featured content component (DSD-1566, DR-2592)
 
 ### Update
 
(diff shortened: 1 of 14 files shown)


logger.js
```diff
@@ -51,6 +51,16 @@ const formatter = printf((info) => {
   return JSON.stringify(logObject);
 });
 
+// API workaround for creating log directory in Vercel
+import axios from "axios";
+
+const logDirCreationEndpoint =
+  "https://digital-collections-git-feature-dr-2715logging-nypl.vercel.app/api/createLogDir";
```
EdwinGuzman, line 58: This should work if it's `/api/createLogDir`.

EdwinGuzman:
The previous PR/feature worked on AWS and Cloudwatch but not on Vercel? Did Vercel throw errors somewhere or how did this issue come up?

7emansell:
@EdwinGuzman exactly– logger worked locally as well as on qa-dcfrontend, and logged to Cloudwatch as expected. On Vercel, the build succeeds, so nothing was flagged when I pushed it up, but when you actually view the preview it fails with a 500 error (I guess I hadn't thought to check the preview because I wasn't making any visual changes). There's not loads of documentation but seemingly Vercel needs to use a serverless function to read or write to the file system. This is a possible workaround Ali and I discovered on Friday

logger.js
7emansell: As in `const logDirCreationEndpoint = '/api/createLogDir'` ?

PR #58, opened 2024-01-26
DR-2741: Meta tags QA additions

7emansell:
## Ticket:

Fixes JIRA ticket [DR-2741](https://jira.nypl.org/browse/DR-2741).

- Adds a couple missing meta tags

## How has this been tested?

Locally

### Checklist:

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -48,7 +48,6 @@ export default function App({ Component, pageProps }: AppProps) {
           name="google-site-verification"
           content="_shbOK1otHA_eFNFgJsOwITZrWQwRg4wr8nmrJPNVDM"
         />
-        <meta name="csrf-param" content="authenticity_token" />
         <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
       </Head>
       <DSProvider>
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -44,6 +44,16 @@ export default function App({ Component, pageProps }: AppProps) {
         />
         <meta name="twitter:site" content="@nypl" />
         <meta name="twitter:creator" content="@nypl" />
+        <meta
+          name="google-site-verification"
+          content="_shbOK1otHA_eFNFgJsOwITZrWQwRg4wr8nmrJPNVDM"
+        />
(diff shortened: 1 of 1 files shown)


src/pages/_app.tsx
```diff
@@ -44,6 +44,16 @@ export default function App({ Component, pageProps }: AppProps) {
         />
         <meta name="twitter:site" content="@nypl" />
         <meta name="twitter:creator" content="@nypl" />
+        <meta
+          name="google-site-verification"
+          content="_shbOK1otHA_eFNFgJsOwITZrWQwRg4wr8nmrJPNVDM"
```
EdwinGuzman, line 49: I _think_ this is okay. I don't know enough about how Google verifies sites and if you can reuse the same hash on multiple sites that have different domain names. At some point, this app will go on "digitalcollections.nypl.org" so by then it won't be a problem and it's likely fine to keep this here now.

src/pages/_app.tsx
```diff
@@ -44,6 +44,16 @@ export default function App({ Component, pageProps }: AppProps) {
...
+        />
+        <meta name="csrf-param" content="authenticity_token" />
+        <meta
+          name="csrf-token"
+          content="YGD90UPAnu6p7nAVRLc7kYIQ9hBGSuLSphWaMZdj/r/pv9PTpDmiEQwEpy0Cv6CN7emeN6vjQOprMKv/+q3WeA=="
+        />
```
EdwinGuzman, line 55: The CSRF token is dynamic and different every time a patron visits the site. So this should not be hard coded. At the moment for just the homepage, the app does not need a CSRF token.

This will be useful when validating a form or when a user is signed in and needs to perform authenticated actions, but right now this is ok to leave out. 

It'd be good to know how this token is being generated and how it's used for future reference.

EdwinGuzman (changes requested):
Remove the CSRF tags but otherwise looks good.

src/pages/_app.tsx
7emansell: Also remove the csrf-param tag?

EdwinGuzman: Yes, thanks.

EdwinGuzman (approved):

PR #64, opened 2024-02-09
add logs for adobe env variables

avertrees:
log environment variables used in Adobe Analytics implementation

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -52,7 +52,6 @@ export default function App({ Component, pageProps }: AppProps) {
 
   // Track page view events to Adobe Analytics
   useEffect(() => {
-    console.log("ADOBE_EMBED_URL: ", process.env.ADOBE_EMBED_URL);
     console.log("appConfig.adobeEmbedUrl: ", appConfig.adobeEmbedUrl);
     console.log("appConfig.environment", appConfig.environment);
     trackVirtualPageView(router.asPath);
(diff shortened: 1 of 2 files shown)


src/pages/_app.tsx
```diff
@@ -52,6 +52,9 @@ export default function App({ Component, pageProps }: AppProps) {
 
   // Track page view events to Adobe Analytics
   useEffect(() => {
+    console.log("ADOBE_EMBED_URL: ", process.env.ADOBE_EMBED_URL);
```
EdwinGuzman, line 55: Ah, so you can't access `process.env` on the client side (useEffect runs on the client) because that's a server side only global variable (in node). So this _should_ be accessed through appConfig instead. If it's not working then update `ADOBE_EMBED_URL` to `NEXT_PUBLIC_ADOBE_EMBED_URL` everywhere. It's considered "unsafe" but in this case, we do want this to be sent to the client so it's okay to do this.

avertrees: can I add this line to the utils file, since that runs on the server?

EdwinGuzman: Yes, exactly, that's the first place where the env var will be read.

avertrees: cool, donezo

EdwinGuzman: Did you push anything new?

avertrees: i forgot to push it up oops

avertrees: it's there now

EdwinGuzman (approved):

PR #66, opened 2024-02-12
remove old travis keys

avertrees:
remove old keys that were not deleted after the last rotation of travis keys, see https://github.com/NYPL/digital-collections/pull/60/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485

wondering if this might be tripping up the travis builds? I'd imagine Travis knows to use the latter of the two, but worth trying this first.


croyfish:
Ah yeah, sorry about this. I was falsely assuming that the comments were on the same lines as the keys. It should work as expected the next time around.

PR #70, opened 2024-02-14
update environment variable readme section

avertrees:
update "environment variables" section in readme to reflect Edwin's findings from yesterday and our plan for managing ENV vars at this current moment

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,21 +48,27 @@ A quick note on environment variables
 
 Use `.env.local` for local development.
 
-### Server Side Environment Variables
+### Local development
 
-If a variable is required ONLY on the SERVER side (ie. in an api endpoint), then the variable can be declared in `.env.local.`
+All variables should be declared in `.env.local` for local development. For local development, we don't have to worry about Server and Client env vars and where they come from.
 
-SERVER side methods in Next.js read environments at RUN TIME, which means they have access to environment variables on the server - these can be found by running `env` or referenced within the app as `process.env.ENV`. These variables can/should be managed in QA and Production environments by updating the Task Definition and/or Cloud Formation Templates.
+### QA/Production environments
+
+#### Server Side Environment Variables
+
+If a variable is required ONLY on the SERVER side (ie. in an api endpoint), then the variable can be declared in the AWS Task Definition.
+
+SERVER side methods in Next.js read environments at RUN TIME, which means they have access to environment variables on the server - these can be found by running `env` in the bash terminal or referenced within the app as `process.env.ENV`. These variables can/should be managed in QA and Production environments by updating the Task Definition and/or Cloud Formation Templates.
 
 SERVER side references of environment variables can use `process.env.ENV`
 
-### Client Side Environment Variables
+#### Client Side Environment Variables
 
 If a variable is required ONLY on the CLIENT, the environment variable needs to be declared in `appConfig.ts`.
 
 CLIENT side methods in Next.js only have access to environment variables that are declared at BUILD TIME, which means they DO NOT have access to environment variables on the server and thus do not have access to `process.env.ENV`. These variables should be managed in QA and Production environments by either hard coding them in `appConfig.ts` AND/OR declaring them in the Dockerfile and travis.yml.
 
-Currently, the only CLIENT side environment variable that we use is APP_ENV. APP_ENV drives the logic to choose between several env variables that are hard coded in `appConfig.ts`.
+Currently, the only CLIENT side environment variable that we use is `APP_ENV`. APP_ENV drives the logic to choose between several env variables that are hard coded in `appConfig.ts`.
 
 Example:
 
--- a/README.md
+++ b/README.md
@@ -48,9 +48,43 @@ A quick note on environment variables
 
 Use `.env.local` for local development.
 
+### Server Side Environment Variables
+
+If a variable is required ONLY on the SERVER side (ie. in an api endpoint), then the variable can be declared in `.env.local.`
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,9 +48,43 @@ A quick note on environment variables
 
 Use `.env.local` for local development.
 
+### Server Side Environment Variables
+
+If a variable is required ONLY on the SERVER side (ie. in an api endpoint), then the variable can be declared in `.env.local.`
```
EdwinGuzman, line 53: All variables can/should be declared in `.env.local` for local development. For local development, we don't have the confusing issue about where env vars are coming from and how they are used since the app won't technically be built for local development.

Perhaps emphasize that `.env.local` is specifically a convention for local development and it's safe to put any and all env variable values there. This file is not committed to Github.

README.md
```diff
@@ -48,9 +48,43 @@ A quick note on environment variables
...
+
+SERVER side methods in Next.js read environments at RUN TIME, which means they have access to environment variables on the server - these can be found by running `env` or referenced within the app as `process.env.ENV`. These variables can/should be managed in QA and Production environments by updating the Task Definition and/or Cloud Formation Templates.
```
EdwinGuzman, line 55: "by running `env` in a bash terminal  or..."

README.md
```diff
@@ -48,9 +48,43 @@ A quick note on environment variables
...
 SERVER side references of environment variables can use `process.env.ENV`
 
-If you are referencing an environment variable on the CLIENT, the environment variable needs to be added to `nextConfig.env` in `next.config.js.` Please use the process.env.ENV reference for the variable.
+### Client Side Environment Variables
+
+If a variable is required ONLY on the CLIENT, the environment variable needs to be declared in `appConfig.ts`.
+
+CLIENT side methods in Next.js only have access to environment variables that are declared at BUILD TIME, which means they DO NOT have access to environment variables on the server and thus do not have access to `process.env.ENV`. These variables should be managed in QA and Production environments by either hard coding them in `appConfig.ts` AND/OR declaring them in the Dockerfile and travis.yml.
+
+Currently, the only CLIENT side environment variable that we use is APP_ENV. APP_ENV drives the logic to choose between several env variables that are hard coded in `appConfig.ts`.
```
EdwinGuzman, line 65: backticks on APP_ENV

README.md
```diff
@@ -48,9 +48,43 @@ A quick note on environment variables
...
+SERVER side methods in Next.js read environments at RUN TIME, which means they have access to environment variables on the server - these can be found by running `env` or referenced within the app as `process.env.ENV`. These variables can/should be managed in QA and Production environments by updating the Task Definition and/or Cloud Formation Templates.
+
 SERVER side references of environment variables can use `process.env.ENV`
 
-If you are referencing an environment variable on the CLIENT, the environment variable needs to be added to `nextConfig.env` in `next.config.js.` Please use the process.env.ENV reference for the variable.
+### Client Side Environment Variables
+
+If a variable is required ONLY on the CLIENT, the environment variable needs to be declared in `appConfig.ts`.
+
+CLIENT side methods in Next.js only have access to environment variables that are declared at BUILD TIME, which means they DO NOT have access to environment variables on the server and thus do not have access to `process.env.ENV`. These variables should be managed in QA and Production environments by either hard coding them in `appConfig.ts` AND/OR declaring them in the Dockerfile and travis.yml.
```
EdwinGuzman, line 63: This is partly true.

Another way to declare client-side variables is by prefixing them with `NEXT_PUBLIC_`. Then they _are_ accessible in the app as `process.env.NEXT_PUBLIC_[VAR_NAME]`.

But, they **must** be declared at build time like you mentioned.

This pattern of using `APP_ENV` and the hard coded values in `appConfig.ts` works well but you _could_ set client side env vars through docker/travis. For example, you can set `NEXT_PUBLIC_ADOBE_URL="prod-url"` in travis.yml and send it to the dockerfile so it's there at build time for the app. Then in the app, you can use `process.env.NEXT_PUBLIC_ADOBE_URL` on both the client and server.

You don't have to mention this if the decided pattern is to use `appConfig.ts` and the `APP_ENV` variables. This convention/pattern works well and I'm happy to support this, but we should choose one and we should be clear on the alternatives in case questions come up.

README.md
```diff
@@ -48,9 +48,49 @@ A quick note on environment variables
...
 SERVER side references of environment variables can use `process.env.ENV`
 
-If you are referencing an environment variable on the CLIENT, the environment variable needs to be added to `nextConfig.env` in `next.config.js.` Please use the process.env.ENV reference for the variable.
+#### Client Side Environment Variables
+
+If a variable is required ONLY on the CLIENT, the environment variable needs to be declared in `appConfig.ts`.
+
+CLIENT side methods in Next.js only have access to environment variables that are declared at BUILD TIME, which means they DO NOT have access to environment variables on the server and thus do not have access to `process.env.ENV`. These variables should be managed in QA and Production environments by either hard coding them in `appConfig.ts` AND/OR declaring them in the Dockerfile and travis.yml.
+
+Currently, the only CLIENT side environment variable that we use is `APP_ENV`. APP_ENV drives the logic to choose between several env variables that are hard coded in `appConfig.ts`.
```
EdwinGuzman, line 71: Diego recently encountered some issues where even `APP_ENV` was not working. This may need to be updated to `NEXT_PUBLIC_APP_ENV`. If you don't encounter issues then keep it as is but just pointing this out now.

EdwinGuzman (approved):
I think eventually this may need to be updated to include more information on using `NEXT_PUBLIC_` specific env vars. Right now, the vars are working as expected and if this makes sense to others, Emma and Chris, then this is good to go.

Ask Chris for a review because he's working on adding New Relic through env vars.

PR #72, opened 2024-02-21
DR-2772 add newrelic agent

chrisb87:
## Ticket:

- JIRA ticket [DR-2772](https://jira.nypl.org/browse/DR-2772)

- Installs and configures the NewRelic agent
- Names the service "Facelift" (is this what we want?)
- Adds the NewRelic license key using an environment variable
- Does _not_ install NewRelic client side instrumentation, only server side

## How has this been tested?

After running this code locally, I observed a new service named "Facelift" is now listed in our NewRelic dashboard.

- n/a

### Checklist:


newrelic.js
```diff
@@ -0,0 +1,53 @@
...
+ * New Relic agent configuration.
+ *
+ * See lib/config/default.js in the agent distribution for a more complete
+ * description of configuration variables and their potential values.
+ */
+exports.config = {
+  /**
+   * Array of application names.
+   */
+  app_name: ["Facelift"],
```
EdwinGuzman, line 12: I think this should be based on the `APP_ENV` environment variable so the New Relic dashboard can track both QA and prod.

newrelic.js
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Your New Relic license key.
+   */
+  license_key: "51b50512af4dd4c26e70c0d3a1d84c4b371a812a",
```
EdwinGuzman, line 16: This feels safer as an env variable.

package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "NODE_OPTIONS='-r @newrelic/next' next dev",
     "build": "next build",
-    "start": "next start",
+    "start": "NODE_OPTIONS='-r @newrelic/next' next start",
```
EdwinGuzman, line 8: We don't want to include new relic locally so it should be okay to remove from the "dev" script.

For the "start" script, I think we should keep the original `"start": "next start"` command since this gets called automatically on Vercel and I don't think we want to track new relic there. A new one can be added: `"start:newrelic": "NODE_OPTIONS='-r @newrelic/next' next start",` . This new one should be called in the Dockerfile for AWS.

package.json
```diff
@@ -24,6 +24,7 @@
     "trailingComma": "es5"
   },
   "dependencies": {
+    "@newrelic/next": "^0.7.0",
```
EdwinGuzman, line 27: I think you also want to install the non-next new relic package. This is what DRB is doing: https://github.com/NYPL/sfr-bookfinder-front-end/blob/development/package.json#L26C6-L26C14

EdwinGuzman (changes requested):
I left some comments but let me know what you think about putting the qa and prod new relic environments behind the env vars.

Is there a reason why only the server side implementation is needed and not the client side? If so, you can disregard the comment about adding the non-Nextjs npm "newrelic" package. I was talking to @jackiequach about how she implemented NR in Digital Research Books (DRB) and their team ran into issues. If the requirement is only for server side monitoring, then we shouldn't run into the same issues. If we do want the client side implementation, more work is needed and we can meet about it.

newrelic.js
chrisb87: That is, of course, the proper way to do it. :+1:

newrelic.js
chrisb87: Ah, yes of course, that's definitely required to distinguish the two.

package.json
chrisb87: The documentation was unclear on this. They explicitly say for next apps to install the next agent _instead of_ the newrelic agent. They also say this:

`Currently this package is not bundled with the agent, and must be installed as a standalone. However, the package depends on the agent so you will get all the capabilities of the agent when loading this package.`

On the contrary, their example app shows both installed. 

I'll go ahead and install them both.

chrisb87:
> Is there a reason why only the server side implementation is needed and not the client side? 

The quick story is, no it wasn't specified, so I figured we could do the server side first and see if there's more monitoring desired and add on the client side later.

Not sure who's call it is, but do you think we should go ahead and get client side monitoring installed as part of this same PR? I'm happy to do it if so.

package.json
chrisb87: 👍

avertrees:
> Not sure who's call it is, but do you think we should go ahead and get client side monitoring installed as part of this same PR? I'm happy to do it if so.

I am pretty confident that we want client side monitoring installed as well.

EdwinGuzman (commented):
The update looks good and this PR is good as is for server side monitoring if it's appearing in the newrelic dashboard. I agree we should add client-side monitoring but not sure about the timing.

avertrees:
If I remember correctly, the reporting is mainly for Lana to compare load times before and after we implement the skeleton loading. I don't know if we can get the metrics Lana is looking for with just the server side reporting, but perhaps @chrisb87's approach of testing this out with the server side support first makes the most sense? Then we check in with Lana to see if she needs more data, and if so, add the client side later?

avertrees (approved):

EdwinGuzman (approved):
Just add the newrelic key in the travis or dockerfile so that the app has this at build time. I'm not sure if this will work through a task definition. @avertrees thoughts?

avertrees:
> Just add the newrelic key in the travis or dockerfile so that the app has this at build time. I'm not sure if this will work through a task definition. @avertrees thoughts?

I'm assuming server side monitoring will use env vars on the server - so the task definition should be enough. @EdwinGuzman? Right? 

We could add the newrelic key to the travis and dockerfile just to be safe but I'm leaning towards saying that the task definition is enough.

EdwinGuzman:
The env var is set in a js file so yeah, it should be read by the file, but I'm not sure if the app will pick up updated config settings _while_ it's already running. So perhaps better to set it in travis and pass it down to docker.

PR #74, opened 2024-02-26
DR-2787 and DR-2788 skeleton load the homepage

avertrees:
## Ticket:

- JIRA ticket [DR-2787](https://jira.nypl.org/browse/DR-2787) 
- [DR-2788](https://jira.nypl.org/browse/DR-2788)

- Moves Repo API calls out of getServerSideProps and into new internal /api endpoints
- Moves calls to internal /api endpoints out of index.tsx and into [campaignHero.tsx](https://github.com/NYPL/digital-collections/compare/feature/DR-2788/skeleton-loading?expand=1#diff-1b6f9073a530aa8ad0e9f6481acfdc86b5dda84ae3ab6cdb71f916760b54359a) and [homePageMainContent.tsx](https://github.com/NYPL/digital-collections/compare/feature/DR-2788/skeleton-loading?expand=1#diff-f5d5aa2a27ed15089b8a1f873bb3aba63d767dfe9cafbeb7337283bbf9124096)
- Creates 2 new components - [swimLanesLoading.tsx](https://github.com/NYPL/digital-collections/compare/feature/DR-2788/skeleton-loading?expand=1#diff-45f15c7b88b2074dc8e43f25e1df6429dcf5087b7c4a97f9522eaccfec941c48) and [swimLanesLoading.tsx](https://github.com/NYPL/digital-collections/compare/feature/DR-2788/skeleton-loading?expand=1#diff-45f15c7b88b2074dc8e43f25e1df6429dcf5087b7c4a97f9522eaccfec941c48) 
- Render skeleton loader components while component waits for data waits to be returned.

## How has this been tested?

Locally in browser

-

### Checklist:


src/components/swimlanes/swimLanesLoading.tsx
```diff
@@ -0,0 +1,37 @@
+import {
+  SkeletonLoader,
+  SimpleGrid,
+} from "@nypl/design-system-react-components";
+
+const SwimLanesLoading = () => {
+  return (
+    <SimpleGrid columns={4}>
+      <SkeletonLoader imageAspectRatio="landscape" isBordered contentSize={1} />
```
EdwinGuzman, line 9: Add a "title" skeleton loader with `showImage={false}`, `contentSize={0}`, and `headingSize={1}`.

src/components/hero/campaignHero.tsx
```diff
@@ -1,24 +1,52 @@
...
 import CampaignHeroHeading from "./campaignHeroHeading";
+import CampaignHeroLoading from "./campaignHeroLoading";
+import { useEffect, useState } from "react";
 
-const CampaignHero = ({ featuredItem, numberOfDigitizedItems }: any) => {
-  return (
+const CampaignHero = ({ imageID }) => {
+  console.log(imageID);
+  // const [isFeaturedItemLoading, setIsFeaturedItemLoading] = useState(false);
+  const [data, setData] = useState<any>({});
```
EdwinGuzman, line 10: As discussed, this is pretty much the same as the `isFeaturedItemLoading` variable since this shows that the data is retrieve and available. Okay to just use `data` and check of the updated data's property(ies).

src/components/hero/campaignHeroLoading.tsx
```diff
@@ -0,0 +1,10 @@
+import { SkeletonLoader } from "@nypl/design-system-react-components";
+
+const CampaignHeroLoading = () => {
+  return (
+    // the height of the placeholder image is too large but I only see a width param in the storybook docs
```
EdwinGuzman, line 5: As discussed, try wrapping this in a DS `Box` and add a height or max height.

src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -1,13 +1,36 @@
...
 import FeaturedContentComponent from "../featuredContent/featuredContent";
 import SwimLanes from "../swimlanes/swimLanes";
+import { useEffect, useState } from "react";
+import SwimLanesLoading from "../swimlanes/swimLanesLoading";
 
-const HomePageMainContent = ({ randomNumber, lanesWithNumItems }) => {
-  return (
+const HomePageMainContent = () => {
+  // const [isSwimLaneLoading, setIsSwimLaneLoading] = useState(false);
+  const [data, setData] = useState<any>({});
```
EdwinGuzman, line 9: Similar idea here to the featured item check, it's okay to use the availability of `data` and its properties for the loading state.

src/components/swimlanes/swimLanesLoading.tsx
EdwinGuzman: Adding a title skeleton loader title will throw off the `SimpleGrid`'s four columns. So the DOM structure will have to be reworked a bit.

src/components/swimlanes/swimLanesLoading.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <SkeletonLoader imageAspectRatio="landscape" isBordered contentSize={1} />
+      <SkeletonLoader imageAspectRatio="landscape" isBordered contentSize={1} />
+      <SkeletonLoader imageAspectRatio="landscape" isBordered contentSize={1} />
```
EdwinGuzman, line 12: Remove the borders.

src/pages/api/featuredHero.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+
+    return response.status(200).json({
+      featuredItem: featuredItemObject,
+      numberOfDigitizedItems: numDigitizedItems,
+    });
+  }
+};
+export default itemsDataHandler;
+// http://localhost:3000/api/featuredItems
```
EdwinGuzman, line 41: This comment doesn't seem right.

avertrees: it's on the homepagedata endpoint too - I'll update them both. 

side note: not sure how I feel about the names for these endpoints.

src/components/hero/campaignHero.tsx
```diff
@@ -1,24 +1,44 @@
...
+    const fetchData = async () => {
+      const response = await fetch(`/api/featuredHero?imageID=${imageID}`);
+      const responseData = await response.json();
+      console.log(responseData);
+      setData(responseData);
+    };
+
+    console.log("fetching hero data");
+    fetchData();
+  });
```
EdwinGuzman, line 21: This is causing an infinite loop but not really sure why. What works is to add a second parameter, the dependency array with `imageID`. So on line 21, `, [imageID]);`

src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -1,13 +1,32 @@
...
+    const fetchData = async () => {
+      const response = await fetch("/api/homepagedata");
+      const responseData = await response.json();
+      console.log(responseData);
+      setData(responseData);
+    };
+
+    console.log("fetching swim lane data");
+    fetchData();
+  }, []);
```
EdwinGuzman, line 20: In this case, this does have the effect of calling this once because there are no deps.

src/components/hero/campaignHero.tsx
EdwinGuzman: I noticed this on the terminal but you can see this in the browser's console as well.

EdwinGuzman (changes requested):
The biggest update needed is to fix the infinite loop being called for fetching hero data, and fix the tests.

src/components/homePageMainContent/homePageMainContent.tsx
avertrees: this returns an empty array while the other does not.

EdwinGuzman: What returns an empty array? I meant that there are no deps in the array so the empty array, as the argument to the `useEffect`, makes it call the function once.

src/components/hero/hero.test.tsx
```diff
@@ -1,16 +1,43 @@
 import CampaignHero from "@/components/hero/campaignHero";
 import { render, screen } from "@testing-library/react";
 import { props } from "../../../__tests__/data/homepageProps";
```
EdwinGuzman, line 3: This is no longer being used.

src/components/homePageMainContent/homePageMainContent.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+    (global as any).fetch = jest.fn(() =>
+      Promise.resolve({
+        status: 200,
+        json: () => Promise.resolve({ props }),
+      })
+    ) as jest.Mock;
+
+    const { container } = render(<HomePageMainContent />);
+
+    expect(container.classList.contains("chakra-skeleton"));
```
EdwinGuzman, line 17: I would not expect this to pass because once the data is fetched (as in line 11), it would not render the skeleton loader. This is also missing the second part of the test so it's giving a false positive. Add `.toBeTruthy()` at the end to verify that what's inside the expect is true.

This test makes sense but remove the passed props on line 11. This should test that the skeleton loader renders when there is no data returned from the API.

Then another another test where you are passing the props and you check for expected swimlane or featured content data on the page. This follows the same pattern you did for `hero.test.tsx`.

EdwinGuzman (changes requested):
Some updates for the test but looking better.

src/components/hero/hero.test.tsx
```diff
@@ -1,16 +1,41 @@
...
+  //             uuid: "3a2cced0-12ca-0133-7a6d-58d385a7bbd0",
+  //             title:
+  //               "Latona Giving Birth to Apollo and Diana on the Island of Delos",
+  //             href: `https://qa-digitalcollections.nypl.org/items/3a2cced0-12ca-0133-7a6d-58d385a7bbd0`,
+  //           },
+  //           numberOfDigitizedItems: "863,848",
+  //         }),
+  //     })
+  //   ) as jest.Mock;
+  //   render(<CampaignHero imageID="5273165" />);
```
EdwinGuzman, line 35: The latest updates are great. You're still missing this commented out test and a similar one for the swim lanes component. If you're not going to test it right now, fine... but make sure you add/fix this test at some point before prod deployment. Let me know what you plan on doing and I can approve. The comment below is along the lines to fixing this.

avertrees: The commented out test fails because `render(<CampaignHero imageID="5273165" />)` returns the skeleton loader... so I'm not quite sure how to get this to pass. We can discuss this on Monday? Hopefully QA will be unblocked by then.

PR #75, opened 2024-02-28
DR-2772 (bug fix): make NEWRELIC_LICENSE_KEY available during build

chrisb87:
This is a bug fix for https://jira.nypl.org/browse/DR-2772

This PR makes NEWRELIC_LICENSE_KEY available during the build process.

TODO: I am having trouble with my local ruby configuration and am unable to encrypt the NEWRELIC_LICENSE_KEY using the travis gem. I will need someone else to to push that encrypted value to this branch before merging.


Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM node:18-alpine AS production
 #RUN apt-get upgrade -y
 
 ARG APP_ENV
+ARG NEW_RELIC_LICENSE_KEY
```
EdwinGuzman, line 7: Add this as an env similar to `APP_ENV` on line 19.

avertrees: good catch

EdwinGuzman (approved):

PR #81, opened 2024-03-05
DR-791 enable CORS

avertrees:
## Ticket:

- JIRA ticket [DR-2791](https://jira.nypl.org/browse/DR-2791)

- enables CORS to protect internal api endpoints following the example in this blog: https://blog.logrocket.com/using-cors-next-js-handle-cross-origin-requests/

## How has this been tested?

1. In the browser with the dev tools open, open the network tab and go to http://localhost:3000/api/featuredItem or http://localhost:3000/api/homepage
2. Check the response headers, the following CORS headings should be there. 
ACCESS_CONTROL_ALLOW_CREDENTIALS="true"
ACCESS_CONTROL_ALLOW_METHODS="GET,OPTIONS,PATCH,DELETE,POST,PUT"
ACCESS_CONTROL_ALLOW_HEADERS="X-CSRF-Token, X-Requested-With, Accept, Accept-Version, Content-Length, Content-MD5, Content-Type, Date, X-Api-Version"
Note: ACCESS_CONTROL_ALLOW_ORIGIN should be dynamic based on environment, so it should be http://localhost:3000 when run locally.
3. Load http://localhost:3000/api/featuredItems with the network tab open, and the headers should be different. They should be:     
ACCESS_CONTROL_ALLOW_CREDENTIALS="false"
ACCESS_CONTROL_ALLOW_ORIGIN="*"
ACCESS_CONTROL_ALLOW_METHODS="GET,DELETE,PATCH,POST,PUT"
ACCESS_CONTROL_ALLOW_HEADERS="X-CSRF-Token, X-Requested-With, Accept, Accept-Version, Content-Length, Content-MD5, Content-Type, Date, X-Api-Version"

Additionally, I think the expected behavior is for the protected endpoints is the return  a 401/403 error when they are hit using curl.

## Questions, concerns, updates

1. I am struggling to figure out how to test this. I see that the middleware is working, but I'm not sure if the CORS origin is correct or if the request is even getting an origin (req.headers.origin returns undefined and req.headers.get("origin") returns null)
2. Are these CORS headers correct? specifically origin and methods. I know the headers for the protected endpoints use .env and the non-protected endpioints use hard coded headers. They should probably both use .env, just want to make sure we agree and what those headers should be for both.


EdwinGuzman:
We decided to close this as Next.js provides default CORS for internal API routes which we want to keep.

PR #83, opened 2024-03-07
[DR-2801] Use Repo API v2/items/featured endpoint for featured image

kayleiburke:
## Ticket:

- [DR-2801](https://jira.nypl.org/browse/DR-2801): Use Repo API v2/items/featured endpoint in DC Facelift

- Fixes 403 errors that were occurring when trying to load images for the hero component on the home page.  This component now calls the new `v2/items/featured` repo api endpoint, added in [DR-2759](https://jira.nypl.org/browse/DR-2759).  This endpoint selects an image at random and returns it to the hero component, simplifying the front end logic and ensuring that an existing image is always returned.

## How has this been tested?

Hero component is generating correctly when running the app both locally and in [Vercel](https://digital-collections-no16a22n5-nypl.vercel.app/).  On every page refresh, the foreground and background image of the hero component are generating correctly, and the **Featured Item** title and link match the image and successfully redirect the user to the correct Digital Collections page when clicked.

### Example
Screenshot 2024-03-08 at 7 49 45 AM

#### After clicking the link:
Screenshot 2024-03-08 at 7 49 26 AM

- Images can sometimes take a second to load in the [Vercel](https://digital-collections-no16a22n5-nypl.vercel.app/) app, but this is for all images on the page, and not just those for the hero component.  In order to improve load time, we can reduce the resolution of the images being returned (see [this comment](https://github.com/NYPL/digital-collections/pull/83/files#r1517752547) for details), but lower resolution is also less visually appealing for users.

### Checklist:

**Note:** I'm not sure what constitutes as "relevant accessibility documentation"


src/utils/utils.ts
```diff
@@ -140,15 +152,13 @@ export const apiCall = async (apiUrl: string) => {
...
-    `getItemDataFromImageID: About call getAPIUri and apiCall for image id: ${imageID}`
-  );
-  const apiUri = await getAPIUri("local_image_id", imageID);
-  const data = await apiCall(apiUri.apiUri);
+export const getFeaturedImage = async () => {
+  console.log(`getFeaturedImage: About call getAPIUri`);
+  const apiUri = await getAPIUri("featured", "", { random: "true" });
   return {
-    uuid: apiUri.uuid,
-    title: getTitleFromRepoAPIResponseData(data) || "",
```
kayleiburke, line 151: This method is no longer being used, but I'm not sure if we want to keep it in this file, just in case it's ever needed in the future?

src/utils/utils.ts
```diff
@@ -117,8 +121,16 @@ export const getAPIUri = async (identifierType: string, identifier: string) => {
  * @param {string} apiUrl - the url to make a request to
  */
 
-export const apiCall = async (apiUrl: string) => {
+export const apiCall = async (
+  apiUrl: string,
+  urlParam?: { [key: string]: any }
```
kayleiburke, line 126: I added an optional `urlParam` so that that we could add the needed url key/param to the query string when calling the `featured` endpoint.  If the `random` param is not set to `true` in the query string, then numerous results are returned per [code below](https://github.com/NYPL/repository-api/blob/qa/app/controllers/api/v2/items_controller.rb#L105):
Screenshot 2024-03-08 at 7 29 07 AM

**Example:**
Screenshot 2024-03-08 at 7 32 20 AM

croyfish:
This is working great on my local. Am interested to know more about the resolution changes needed to make it consistent, and if those meet the product requirements. Looking through the code changes now.

src/components/hero/campaignHero.tsx
```diff
@@ -19,7 +19,7 @@ const CampaignHero = ({ imageID }) => {
 
   return data?.featuredItem ? (
     <Hero
-      backgroundImageSrc={data.featuredItem.imageSrc}
+      backgroundImageSrc={data.featuredItem.backgroundImageSrc}
```
croyfish, line 22: 👍

src/pages/api/featuredItem.ts
```diff
@@ -20,12 +20,23 @@ const featuredItemDataHandler = async (
...
+        "full",
+        "!700,700",
+        "0"
+      ),
+      foregroundImageSrc: imageURL(
+        dataFromUri.imageID,
+        "full",
+        "!900,900",
+        "0"
+      ),
```
kayleiburke, line 39: The default image size set by `imageURL` is 1600 x 1600, which was causing a `ERR_INCOMPLETE_CHUNKED_ENCODING` error to occur when loading the foreground and/or background images for the hero component, causing one or both of them to be blank:
Screenshot 2024-03-08 at 7 57 48 AM

When playing around with different sizes, the configurations above seemed to yield a good balance between being small enough for images to load quickly and without errors, and being large enough that images are not blurry.

croyfish: Is this a good time to remove https://github.com/NYPL/digital-collections/blob/main/src/data/featureditems.json ?

croyfish:
Is this a good opportunity to remove https://github.com/NYPL/digital-collections/blob/main/src/data/featureditems.json ?

EdwinGuzman:
Yes, the `featureditems.json` file can be removed.

src/utils/utils.ts
EdwinGuzman: Don't think we'll need this so feel free to remove this.

src/utils/utils.ts
EdwinGuzman: Good point. I think we'll only call the endpoint with the `?random=true` query param so this code should expect only one result. This is good to know if this changes in the future.

EdwinGuzman (changes requested):
This looks great! The update itself is working. I also appreciate the typescript type inclusion for the `urlParam` argument.

The only changes I request is to remove code that is no longer necessary like the `getTitleFromRepoAPIResponseData` function you pointed out. There are a few more variables and functions that can be removed:
- in `utils.ts`, it should be safe to remove the functions `setFeaturedItemArray`, `generateRandomImageId`, and `featuredImageID`. Then any place that invoke those functions can be cleaned up.
- check `/src/pages/index.tsx`. The `getServerSideProps` function was used to generate a hero `imageID` and that was passed to the UI components. With your changes, this is no longer needed so you can remove the entire `getServerSideProps` function!
- A few things will break when this is removed, in the same `/src/pages/index.tsx` file, you'd have to remove more `imageID` references. This value is passed to the `CampaignHero` component as a prop. This can also be removed. 
- Last thing, in the `CampaignHero` component, the `?imageID=${imageID}` from the fetch call can be removed since this is no longer applicable.

Let me know if you want to go over these removals together.

src/pages/api/featuredItem.ts
```diff
@@ -20,12 +20,23 @@ const featuredItemDataHandler = async (
       ? featuredImageID(request.query.imageID as string)
       : featuredImageID();
 
-    const dataFromUri = await getItemDataFromImageID(imageID);
+    const dataFromUri = await getFeaturedImage();
```
avertrees, line 23: nitpick: since we're not longer retrieving the featured image data via uri, it might make sense to change the name of this constant to something like "featuredImageData"

avertrees:
Kaylei, this is great!! I have a few nitpicky comments re: naming but those changes don't need to be made before merging this PR.

avertrees (approved):

src/pages/api/featuredItem.ts
kayleiburke: oh yes, good idea!  thank you very much 🙌

src/utils/utils.test.tsx
```diff
@@ -1,177 +0,0 @@
...
-
-// if rotation is not a string
-// if rotation is not supplied
-// if rotation is a string & not default
-
-//// Example usage:
-// let numberWithCommas = addCommas(1234567);
-// console.log(numberWithCommas); // Output: "1,234,567"
-
-import { adobeAnalyticsRouteToPageName } from "../utils/utils";
```
EdwinGuzman, line 160: We still do want to keep this test.

kayleiburke: oh jeez, sorry about that!  I got a little "delete" happy 😅 .  I'll have this resolved shortly.

kayleiburke:
> This looks great! The update itself is working. I also appreciate the typescript type inclusion for the `urlParam` argument.
> ...

@EdwinGuzman I think I made all the changes you requested, but I'm wondering if you wouldn't mind taking a look when you get a chance.  I don't love the fact that I deleted so many tests, as it seems that some of them could simply be modified rather than completely deleted, but I'm not sure how soon this PR needs to be merged.  I can take some extra time to update these tests if needed.  If you could let me know which you prefer, and if you have any other suggestions for changes, I would appreciate it! 🙌

src/pages/api/featuredItem.ts
```diff
@@ -16,20 +15,27 @@ const featuredItemDataHandler = async (
...
+        "full",
+        "!900,900",
+        "0"
+      ),
+      foregroundImageSrc: imageURL(
+        featuredImageData.imageID,
+        "full",
+        "!1100,1100",
+        "0"
+      ),
```
kayleiburke, line 34: Since devops made updates to the memory on the load balancer in front of the IIIF server this morning, images seem to be loading much more quickly/consistently than they were previously.  As a result, I increased the resolution for both the background and foreground images above.  Everything is loading as expected in the Vercel app, but I can't tell if the increased resolution is slowing down the page load slightly.  The latest app is [here](https://digital-collections-756ps0wbe-nypl.vercel.app/), and if you wouldn't mind giving it a try, that would be great!  I can always reduce image resolution back down to what it was before (`700,700` and `900,900`) if needed.

EdwinGuzman (approved):
Looks good, thanks!

avertrees (approved):
looks great! 💯

PR #85, opened 2024-03-08
add IIIF_URL to app config

chrisb87:

PR #87, opened 2024-03-08
Read IIIF_URL setting from app config file instead of .env

kayleiburke:
## Ticket:

- None

- Updates the `imageURL` of the utils file to grab the `IIIF_URL` setting from app config instead of `.env`, thus making use of the changes added in #85:
   Screenshot 2024-03-08 at 2 48 43 PM

## How has this been tested?

- Vercel app now successfully displays images (whereas they were previously showing blank values):
   - **Before:**
      Screenshot 2024-03-08 at 2 24 39 PM
   - **After:** 
      Screenshot 2024-03-08 at 2 47 26 PM

- None

### Checklist:

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -1,4 +1,5 @@
 import featuredItemsData from "../data/featureditems.json";
+import appConfig from "../../appConfig";
 let featuredItemArray = [];
 
 /**
@@ -205,7 +206,6 @@ import {
   ADOBE_ANALYTICS_DC_PREFIX,
   BASE_URL,
 } from "../config/constants";
-import appConfig from "../../appConfig";
 
 /**
  * adobeAnalyticsParam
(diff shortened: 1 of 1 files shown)


src/utils/utils.ts
```diff
@@ -203,6 +205,7 @@ import {
   ADOBE_ANALYTICS_DC_PREFIX,
   BASE_URL,
 } from "../config/constants";
+import appConfig from "../../appConfig";
```
EdwinGuzman, line 208: Can you move this and the other import statement to the top of the file? That's the general convention for importing packages and other files. I know you're following what was already in this file, but it should've been at the top and it was missed before.

kayleiburke: oh yes, my apologies!  My IDE added that there automatically, sorry 😅

avertrees (approved):

PR #96, opened 2024-03-19
DR-2834: Explore further link hover

7emansell:
## Ticket:

Resolves JIRA ticket [DR-2834](https://jira.nypl.org/browse/DR-2834).

- Adds hover underline to Explore further links
- **This requires some overriding of the Link styles (instead of just using the default card heading behavior, as seen in the swim lanes) because the links need to open in a new tab

## How has this been tested?

Locally

### Checklist:

### DIFF
--- a/src/components/exploreFurther/exploreFurther.tsx
+++ b/src/components/exploreFurther/exploreFurther.tsx
@@ -74,6 +74,7 @@ const ExploreFurther = () => {
                 rel="noreferrer noopener"
                 sx={{
                   textDecoration: "none !important",
+                  _hover: { textDecoration: "underline 1px dotted !important" },
                 }}
               >
                 {item.title}


src/components/exploreFurther/exploreFurther.tsx
```diff
@@ -74,6 +74,7 @@ const ExploreFurther = () => {
                 rel="noreferrer noopener"
                 sx={{
                   textDecoration: "none !important",
+                  _hover: { textDecoration: "underline 1px dotted !important" },
```
EdwinGuzman, line 77: This works. Have you tried the `url` prop in the `CardHeading` component? I think that should work but haven't checked. It might allow you to remove lines 76 and 77.

EdwinGuzman: Ah gotcha, I misread that at first.

EdwinGuzman (approved):

EdwinGuzman:
@7emansell would this be easier if the `CardHeading` component was more open to allow more control for the url/link? It's a few levels down from `CardHeading` down to `Link` but it _might_ work. Not sure how worth it is it to do.

7emansell:
> @7emansell would this be easier if the `CardHeading` component was more open to allow more control for the url/link? It's a few levels down from `CardHeading` down to `Link` but it _might_ work. Not sure how worth it is it to do.

Yes definitely but this is a fine work around rn. Not sure how often we're using CardHeading as a Link and then how often we expect to be opening in a new tab...

EdwinGuzman:
That's true but opening in a new tab is also default web behavior so it makes sense to allow it. It also feels like if we update the `CardHeading`, you'd end up doing a lot of work as well to get it work as expected, so this trade off is hard.

avertrees (approved):

PR #98, opened 2024-03-19
DR-2807: About page

7emansell:
## Ticket:

Resolves JIRA ticket [DR-2807](https://jira.nypl.org/browse/DR-2807)

- Adds About page
- Installs jwplayer-react
- Fixes link color in nav

## How has this been tested?

Locally

### Checklist:

### DIFF
--- a/__tests__/pages/aboutpage.test.tsx
+++ b/__tests__/pages/aboutpage.test.tsx
@@ -6,5 +6,5 @@ describe("About page accessibility", () => {
   it("passes axe accessibility test", async () => {
     const { container } = render(<About />);
     expect(await axe(container)).toHaveNoViolations();
-  });
+  }, 15000);
 });
--- a/__tests__/pages/aboutpage.test.tsx
+++ b/__tests__/pages/aboutpage.test.tsx
@@ -1,0 +1,17 @@
+import About from "@/pages/about";
+import { render } from "@testing-library/react";
+import { axe } from "jest-axe";
+
+class ResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+
+describe("About page accessibility", () => {
+  it("passes axe accessibility test", async () => {
+    window.ResizeObserver = ResizeObserver;
+    const { container } = render(<About />);
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
--- a/__tests__/pages/aboutpage.test.tsx
+++ b/__tests__/pages/aboutpage.test.tsx
@@ -2,15 +2,8 @@ import About from "@/pages/about";
 import { render } from "@testing-library/react";
 import { axe } from "jest-axe";
 
-class ResizeObserver {
-  observe() {}
-  unobserve() {}
-  disconnect() {}
-}
-
 describe("About page accessibility", () => {
   it("passes axe accessibility test", async () => {
-    window.ResizeObserver = ResizeObserver;
     const { container } = render(<About />);
     expect(await axe(container)).toHaveNoViolations();
   });
--- a/src/components/navMenu/navMenu.tsx
+++ b/src/components/navMenu/navMenu.tsx
@@ -19,6 +19,7 @@ const NavMenu = ({ render }) => {
         },
         _hover: {
           textDecoration: "none !important",
+          color: "unset",
         },
       }}
     >
--- a/src/components/navMenu/navMenu.tsx
+++ b/src/components/navMenu/navMenu.tsx
@@ -14,12 +14,11 @@ const NavMenu = ({ render }) => {
         fontWeight: "510",
         color: "ui.black",
         marginLeft: "m",
-        ":visited": {
+        _visited: {
           color: "ui.black",
         },
-        ":hover": {
+        _hover: {
           textDecoration: "none !important",
-          color: "ui.link.secondary",
         },
       }}
     >
--- a/src/components/navMenu/navMenu.tsx
+++ b/src/components/navMenu/navMenu.tsx
@@ -19,6 +19,7 @@ const NavMenu = ({ render }) => {
         },
         ":hover": {
           textDecoration: "none !important",
+          color: "ui.link.secondary",
         },
       }}
     >
--- a/src/data/aboutData.tsx
+++ b/src/data/aboutData.tsx
@@ -4,6 +4,7 @@ import {
   List,
   Heading,
   Accordion,
+  Box,
 } from "@nypl/design-system-react-components";
 import JWPlayer from "@jwplayer/jwplayer-react";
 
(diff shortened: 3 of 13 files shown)


7emansell:
Waiting on confirmation about JWPlayer preview and about a broken DPLA link from Lily but otherwise done

src/components/navMenu/navMenu.tsx
```diff
@@ -19,6 +19,7 @@ const NavMenu = ({ render }) => {
         },
         ":hover": {
           textDecoration: "none !important",
+          color: "ui.link.secondary",
```
EdwinGuzman, line 22: I forgot why this needs to be manually added. This style rule is already applied elsewhere 
Screen Shot 2024-03-20 at 1 04 40 PM

src/components/navMenu/navMenu.tsx
```diff
@@ -19,6 +19,7 @@ const NavMenu = ({ render }) => {
         },
         ":hover": {
```
EdwinGuzman, line 20: Btw, the Chakra convention is `_hover`. This does the same thing and is the standard CSS convention.

__tests__/pages/aboutpage.test.tsx
```diff
@@ -0,0 +1,17 @@
+import About from "@/pages/about";
+import { render } from "@testing-library/react";
+import { axe } from "jest-axe";
+
+class ResizeObserver {
```
EdwinGuzman, line 5: Move this to the jest setup file so you don't have to redeclare this in other tests.

src/pages/about/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <>
+      <NotificationBanner />
+      <Header />
+      <Box
+        sx={{
+          maxWidth: "820px",
+          margin: "s",
+          [`@media screen and (min-width: 820px)`]: {
```
EdwinGuzman, line 24: Would using the `TemplateAppContainer` component like the home page help with the static width value? The About page should follow the standard width and now I'm realizing I did not check that detail in the designs before.

src/data/aboutData.tsx
```diff
@@ -0,0 +1,638 @@
...
+  Text,
+  Link,
+  List,
+  Heading,
+  Accordion,
+  Box,
+} from "@nypl/design-system-react-components";
+import JWPlayer from "@jwplayer/jwplayer-react";
+
+const aboutData = [
```
EdwinGuzman, line 11: How about renaming this to include that what is returned are elements? This is not data in the "data" sense like an API data.

src/data/aboutData.tsx
```diff
@@ -0,0 +1,638 @@
...
+      <Heading
+        sx={{ marginTop: "xxl" }}
+        level="h1"
+        size="heading1"
+        text="About NYPL Digital Collections"
+      />
+    ),
+    body: (
+      <Text sx={{ marginBottom: "xxl", marginTop: "s" }}>
+        {" "}
```
EdwinGuzman, line 23: Why this extra space?

src/data/aboutData.tsx
```diff
@@ -0,0 +1,638 @@
...
+    ),
+  },
+  {
+    heading: <Heading level="h2" size="heading4" text="Find items" />,
+    body: (
+      <Box sx={{ marginBottom: "xxl" }}>
+        <Text sx={{ marginTop: "s" }}>
+          {" "}
+          Looking for something? Start with a{" "}
+          <Link href="https://digitalcollections.nypl.org">search</Link> or
```
EdwinGuzman, line 70: Hmm eventually these should be self-referential and be `"/"`, `"/search/index..."`, etc. You could use the `DC_URL` value from appConfig but those urls are also not self-referential. Can you add a TODO comment somewhere so we don't forget? This won't have to be done until this app is entirely complete.

src/data/aboutData.tsx
```diff
@@ -0,0 +1,638 @@
...
+        materials, please see our{" "}
+        <Link href="https://www.nypl.org/help/get-what-you-need/audio-video-files">
+          Audio and Moving Image Preservation and Access Initiative.
+        </Link>
+      </Text>
+    ),
+  },
+  {
+    heading: (
+      <Heading level="h2" size="heading4" text="Copyright status undefined" />
```
EdwinGuzman, line 493: My developer brain thought "undefined" was an error.

EdwinGuzman (changes requested):
Looks great but just pointing out minor things. The JWPlayer seems very easy to work with and I saw the issue about the no preview image shown.

Now is also a good time to ask Clare for an accessibility review before this gets merged.

src/components/navMenu/navMenu.tsx
7emansell: Ah it doesn't I thought I was overriding with the `color` and needed to reset

src/pages/about/index.tsx
7emansell: Yea this was weird. I was assuming (but obviously did not check when reviewing the design 🤦‍♀️) that if the header was on new breakpoints, the width of the content would align. This `820px` width content is in a weird spot between the small tablet breakpoint (`768px`) and the old `lg` breakpoint (`960px`). Bc it's not either, and the page is just this straight up static content, I'm not using the new breakpoints at all as you can see here– all that matters, to create the breakpoint behavior as it is in Figma, is whether the screen width is larger or smaller than the content. Which is all a very long way to say are we gonna keep the content this width knowing that it's not "standard" ?

EdwinGuzman: "820px" is directly from Figma, right? We should point it out to Lily and mention we can keep it but it's not standard and this might change when the new breakpoints are in place globally (and not just in the Header).

This _feels_ off to me but this is up to Lily to test and decide.

EdwinGuzman:
The latest updates look good but there's the jwplayer thumbnail issue and still waiting on accessibility review. Will approve after in case you need to make any code updates.

7emansell:
JWPlayer thumbnail issue fixed– had to put each video into its own playlist on the JWPlayer dashboard for the thumbnails to show up, which was interesting– don't really know where to document that but would like to as I couldn't find it anywhere on the internet. @EdwinGuzman I'll wait for a11y then pass this over to Lily with the width concern noted

EdwinGuzman:
> JWPlayer thumbnail issue fixed– had to put each video into its own playlist on the JWPlayer dashboard for the thumbnails to show up, which was interesting– don't really know where to document that but would like to as I couldn't find it anywhere on the internet. @EdwinGuzman I'll wait for a11y then pass this over to Lily with the width concern noted

This doesn't seem like a dev responsibility to update the JWPlayer dashboard but I understand the need. You could document it in the README but maybe a better place would be confluence for anyone else who has access to the dashboard.

EdwinGuzman (approved):
I think the updates from vqa may be small enough to not require an approval but I can review again. Just hold off on merging this in until we have a plan for QA/release.

avertrees (approved):

avertrees:
Approved - when we launch this we'll want to add functionality to link directly to NYPL Statement on Potentially Harmful Content using some sort of slug or query paramter, ie. /about#nypl_harmful_content_statement

7emansell:
Adding that id, noting that there is the continuing issue of the sticky header actually covering the target element

PR #105, opened 2024-03-29
DR-2894: Quick pre-launch a11y fixes

7emansell:
## Ticket:

Resolves ticket [DR-2894](https://jira.nypl.org/browse/DR-2894).

- Fixes focus outline on hero link
- Removes confusing alt text from FeaturedContent link and image
- Adds SkipNav to header

- These fixes are just to improve accessibility, no visual or functional changes

### Checklist:

### DIFF
--- a/src/components/hero/campaignHeroSubText.tsx
+++ b/src/components/hero/campaignHeroSubText.tsx
@@ -36,7 +36,14 @@ const CampaignHeroSubText = ({ featuredItem }: any) => {
       </Text>
       <HorizontalRule />
       <Box>
-        <Text color="ui.typography.body" mb="0px" noOfLines={2}>
+        <Text
+          color="ui.typography.body"
+          mb="0px"
+          noOfLines={2}
+          __css={{
+            ":focus-within": { overflow: "visible" },
+          }}
+        >
           Featured Image:{" "}
           <DSLink
             color="var(--nypl-colors-ui-link-primary) !important"
--- a/src/components/featuredContent/featuredContent.test.tsx
+++ b/src/components/featuredContent/featuredContent.test.tsx
@@ -13,10 +13,6 @@ describe("Featured Content component renders with expected props", () => {
         "src",
         "/pd_banner.png"
       );
-      expect(within(component).getByRole("img")).toHaveAttribute(
-        "alt",
-        "Public Domain banner"
-      );
 
       const button = within(component).getByTestId("featured-learn-more");
       expect(button).toHaveAttribute("href", "https://publicdomain.nypl.org");
@@ -33,10 +29,6 @@ describe("Featured Content component renders with expected props", () => {
       expect(within(component).getByRole("img")).toHaveAttribute(
         "src",
         "/service-artehouse.jpg"
-      );
-      expect(within(component).getByRole("img")).toHaveAttribute(
-        "alt",
-        "Service Artehouse banner"
       );
 
       const button = within(component).getByTestId("featured-visit-store");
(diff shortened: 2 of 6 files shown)


7emansell:
This was extremely low-hanging fruit, but I'm 90% sure the other two issues flagged on https://jira.nypl.org/browse/DR-2839 will require component-level fixes to `Search` and `CardHeading` media queries. Will pass directly to Clare once I have review.

src/components/hero/campaignHeroSubText.tsx
```diff
@@ -36,7 +36,14 @@ const CampaignHeroSubText = ({ featuredItem }: any) => {
       </Text>
       <HorizontalRule />
       <Box>
-        <Text color="ui.typography.body" mb="0px" noOfLines={2}>
+        <Text
+          color="ui.typography.body"
+          mb="0px"
+          noOfLines={2}
+          __css={{
+            ":focus-within": { overflow: "visible" },
```
7emansell, line 44: Necessary bc of Chakra's otherwise super useful `noOfLines` -> `lineClamp` for this link

EdwinGuzman: I wonder if v3 of the DS would affect this. I won't want this fix to be mixed with the DS v3 update but it'd be worth checking in a separate branch if v3 changes this (because of the Chakra update).

EdwinGuzman:
> This was extremely low-hanging fruit, but I'm 90% sure the other two issues flagged on https://jira.nypl.org/browse/DR-2839 will require component-level fixes to Search and CardHeading media queries.

Agree about `CardHeading` but for issue 5, I think it's the entire fixed header that is the issue.

Clare also made another comment regarding the skip navigation. That should be fixed in this PR, too.  Otherwise, the current fixes look good.

General docs https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/accessibility-guide-skip-navigation--docs
You might be able to add the `[SkipNavigation](https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/components-navigation-skipnavigation--docs)` component.

EdwinGuzman:
The skip nav looks good but I think it should be the first tabbable (sp?) element on the page. So move it above the banner.

When you do press enter on the link, the window scrolls but the focus is lost. I'm not sure if this is an issue but Clare would know more about this.

src/components/hero/campaignHeroSubText.tsx
7emansell: Checked it out and it didn't seem to fix it out of the box. I guess that makes sense bc the `noOfLines` prop is working as expected here, it's just a blunt instrument

7emansell:
@EdwinGuzman Did Clare review the SkipNavigation component originally? Also, focus is lost but when you tab it shows up where you would expect it to (top of main content)

EdwinGuzman:
Yes, she reviewed the component in isolation and within the NYPL Header. Yes, I noticed the focus is lost, that may need an update similar to what you did for the fixed header when you tab on the page.

7emansell:
@EdwinGuzman right, I mean that if this has already been reviewed by Clare then the focus being lost and then regained by tabbing again must be expected behavior?

EdwinGuzman:
Ah, the component has been reviewed but it changes based on the context it's in. In this case, having the fixed header and going to the right HTML element is "new" behavior.

EdwinGuzman (approved):
Dev looks good but ask Clare for a review.

avertrees (approved):

7emansell:
Header not perfect but not a blocker for launch https://jira.nypl.org/browse/DR-2911

PR #106, opened 2024-04-02
DR 2872 puppeteer script for facelift homepage

croyfish:
## Ticket:

- JIRA ticket [Configure passenger settings for Repo API](https://jira.nypl.org/browse/DR-2872)

- Adds a test script to assert that swim lane images load on the homepage

## How has this been tested?

- Install puppeteer with `npm install puppeteer`
- Get on the VPN and accessinate your IP for nypl-dams-dev
- Run the test against qa with `node test/homepageImageLoadingTest.js`. If it does not time out, that means it succeeded. It should print all images loaded and the time taken for the swim lane images to finish loading, including the 12 that require scrolling down the page. This simulates a single user loading the homepage one time.
- Run the test against production with `node test/homepageImageLoadingTest.js --prod`. The above still applies.

- None

### Checklist:

- [ x ] All new and existing tests passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove extra spacing on links (DR-2833)
 - Remove Wallach Prints and Photos from "Explore further" (DR-2871)
 
+### Added
+- Added homepage images loading test script (DR-2872)
+
 ## [0.1.2] 2024-03-20
 
 ### Removed
--- a/README.md
+++ b/README.md
@@ -153,6 +153,23 @@ To run tests, run
 npm run test
 ```
 
+## Test Scripts
+
+To run the homepage swimlane image loading test script against the qa environment, run
+
+```
+npm run hompageImageLoadingTest
+```
+(requires accessination with nypl-dams-dev)
+
+To run the same test script against the production environment, run
+```
+npm run hompageImageLoadingTest --prod
+```
+
+The test waits for the 24 swim lane images to load, including the 12 that load after scrolling, and has succeeded if it does not time out. It simulates a single user loading the homepage one time.
+
+
 ## Getting Started With Docker
 
 First, install ["Docker"](https://www.docker.com/) on your local machine.
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,8 @@
         "jest-axe": "8.0.0",
         "jest-environment-jsdom": "29.7.0",
         "lint-staged": "15.0.1",
-        "prettier": "3.0.3"
+        "prettier": "3.0.3",
+        "puppeteer": "^22.6.2"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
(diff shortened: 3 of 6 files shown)


EdwinGuzman (changes requested):
This is nice! How is this work related to the ticket and passenger settings for Repo API?

Just a couple of changes:

- The description and list of instructions on how to run the script under "how has this been tested?" is great. This should be added to the README so it doesn't get lost and anyone can run the script in the future.
  - Also include that you may need to allow the browser to run tests. I got these popups when I tried to run the script:
Screen Shot 2024-04-02 at 9 42 51 AM

  - This makes me wonder if the tests are running in "headless" mode so they don't need a "real" browser to run.
- The command `node test/homepageImageLoadingTest.js` assumes that you have `puppeteer` installed globally on your machine. Add it as part of the repo through `npm install puppeteer --save-dev`. This will add it to `package.json` and will be installed for development purposes when a dev works in the repo.
 
Screen Shot 2024-04-02 at 9 50 57 AM

Nice to have:
- What do you think about adding the command behind an npm script? We are used to running commands like `npm start`, `npm build`, `npm test`, etc. So you could add the script to the package.json file as `"hompageImageLoadingTest": "node test/homepageImageLoadingTest.js"`. That way, a dev would run `npm run hompageImageLoadingTest`. It's a very minor detail but something like this would keep scripts consistent, e.g. if you're working on the `digital-collections` app, you most likely should use `npm` for scripts.
- I like `puppeteer` but would like to use `cypress` or `playwright` in the future. These are used by other teams for end-to-end testing and scripts and would like to have this be consistent across teams. This is a "nice to have" comment and not something you need to change for this PR but just want to point it out.

croyfish:
> This is nice! How is this work related to the ticket and passenger settings for Repo API?
> 
> Just a couple of changes:
> 
> * The description and list of instructions on how to run the script under "how has this been tested?" is great. This should be added to the README so it doesn't get lost and anyone can run the script in the future.
>   
>   * Also include that you may need to allow the browser to run tests. I got these popups when I tried to run the script:
> 
> Screen Shot 2024-04-02 at 9 42 51 AM
> * This makes me wonder if the tests are running in "headless" mode so they don't need a "real" browser to run.
> * The command `node test/homepageImageLoadingTest.js` assumes that you have `puppeteer` installed globally on your machine. Add it as part of the repo through `npm install puppeteer --save-dev`. This will add it to `package.json` and will be installed for development purposes when a dev works in the repo.
> 
> Screen Shot 2024-04-02 at 9 50 57 AM
> Nice to have:
> 
> * What do you think about adding the command behind an npm script? We are used to running commands like `npm start`, `npm build`, `npm test`, etc. So you could add the script to the package.json file as `"hompageImageLoadingTest": "node test/homepageImageLoadingTest.js"`. That way, a dev would run `npm run hompageImageLoadingTest`. It's a very minor detail but something like this would keep scripts consistent, e.g. if you're working on the `digital-collections` app, you most likely should use `npm` for scripts.
> * I like `puppeteer` but would like to use `cypress` or `playwright` in the future. These are used by other teams for end-to-end testing and scripts and would like to have this be consistent across teams. This is a "nice to have" comment and not something you need to change for this PR but just want to point it out.

Nice suggestions. I can take a look at these today.

EdwinGuzman (approved):
Looks good. Thanks for the update. Another step for another ticket is to make this a Github Action cron job to run daily(?). That's a nice to have.

PR #107, opened 2024-04-03
DR-2893 add NewRelic client side monitoring

chrisb87:
This adds a NewRelic snippet to every page to enable client-side monitoring.

Mostly follows the pattern set by the DRB team, with some small changes to not require the license key to be committed to the repository: https://github.com/NYPL/sfr-bookfinder-front-end/blob/9d7f9060d00e8b345047bd7605b75d953cef73c2/src/lib/newrelic/NewRelic.tsx

Currently this is not working due to a CORS issue when connecting to NewRelic!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


avertrees:
We're going to merge this in to test if the CORS errors only occur locally.

PR #108, opened 2024-04-08
try running swimlanes test with actions

croyfish:
## Ticket:

- JIRA ticket [xxx](url)

## This PR does the following:

-

## Accessibility concerns or updates

-

### Checklist:

- [ ] I have added relevant accessibility documentation for this pull request.
- [ ] All new and existing tests passed.

PR #109, opened 2024-04-09
DR 2912 enable GitHub actions for swimlanes test v2

croyfish:
Fixing a typo in https://github.com/NYPL/digital-collections/pull/108


7emansell (approved):
Lgtm (also looked at #108), just to double check, we're using the Github actions log as where we're storing results of this test?

croyfish:
> Lgtm (also looked at #108), just to double check, we're using the Github actions log as where we're storing results of this test?

I suppose that's true. Do you have a suggestion for how to improve this?

PR #117, opened 2024-04-15
DR-2918 return error message upon google sheet submission error

chrisb87:
https://jira.nypl.org/browse/DR-2918

The application is experiencing an error when submitting an entry to Google sheets. This will just allow the details of the error to be seen in a console.

### DIFF
--- a/src/pages/api/feedback.ts
+++ b/src/pages/api/feedback.ts
@@ -78,8 +78,7 @@ export default async function feedbackFormHandler(
     });
   } catch (e) {
     return response.status(500).send({
-      message:
-        "Something went wrong submitting the feedback form values to the google spreadsheet",
+      message: e.message,
       error: "Internal Server Error",
     });
   }


EdwinGuzman (commented):
This is ok but the message might get lost because it won't be logged or seen anywhere.

In `/src/pages/_app.tsx` there's a call to this internal endpoint and that's wrapped in a try/catch. Also add a console.log in that catch statement to log the message from the error. That's what we will be able to see in the browser's console.

chrisb87:
@EdwinGuzman I'm already seeing the existing generic message, "Something went wrong submitting the feedback form values to the google spreadsheet", in the console. The message is actually being returned by the server side app in response to this call, as a json object, which is then shown in the browser console. So I think this change will reveal the actual error.

EdwinGuzman:
Do you mean you see it in the console in cloudwatch? I don't see that message in the browser console.

chrisb87:
I got it in my browser (Firefox) console!

chrisb87:
Screenshot 2024-04-15 at 8 20 01 AM

EdwinGuzman:
I see, yes, that's the network tab log. The change I mentioned is for the more generic console. It doesn't really matter but it's a convenience thing and also separating the endpoint log with the app log. So try this out on QA to test.

EdwinGuzman (approved):

avertrees (approved):

PR #118, opened 2024-04-16
DR-2918 comment out feedback button

avertrees:
## Ticket:

- JIRA ticket [[DR-2918](https://jira.nypl.org/browse/DR-2918)

- comments out the feedback button. we'll fix it after launch.

## How has this been tested?

- locally or in the vercel preview.


chrisb87 (approved):

EdwinGuzman:
Is the context for removing the `FeedbackBox` component that there's no solution to the 500 error? So it's not New Relic or updated AWS env var values?

7emansell (approved):

PR #120, opened 2024-04-17
DR 2921 Add a test for legacy dc homepage images

croyfish:
## Ticket:

- JIRA ticket [Run puppeteer script on legacy Digital Collections for images.nypl.org](https://jira.nypl.org/browse/DR-2921)

- Added a puppeteer script to test the loading of the swimlanes images on the legacy digital collections homepage.
- Added a new GitHub Actions workflow to also make this script run every 10 minutes

## How has this been tested?

- Do `npm install`
- Do `npm run npm run legacyHomepageImageLoadingTestProd`
- The test should run and should (probably) succeed once the 24 images we expect to load have loaded.
- NB: We'll need to wait until the qa homepage is fixed to do anything with the qa site. That can come in a subsequent PR.

- N/A

### Checklist:

- [ x ] All new and existing tests passed.

### DIFF
--- a/test/legacyHomepageImageLoadingTest.js
+++ b/test/legacyHomepageImageLoadingTest.js
@@ -47,24 +47,6 @@ function parseArgs(args) {
 
   const startTime = new Date();
 
-  // Scroll down the page to trigger image loading
-  //  await page.evaluate(async () => {
-  //    await new Promise((resolve, reject) => {
-  //      let totalHeight = 0;
-  //      const distance = 100;
-  //      const scrollInterval = setInterval(() => {
-  //        const scrollHeight = document.body.scrollHeight;
-  //        window.scrollBy(0, distance);
-  //        totalHeight += distance;
-  //
-  //        if (totalHeight >= scrollHeight) {
-  //          clearInterval(scrollInterval);
-  //          resolve();
-  //        }
-  //      }, 100);
-  //    });
-  //  });
-
   // Wait for all desired images to be present in the DOM
   await page.waitForFunction(() => {
     const desiredImageUrls = [
--- a/test/legacyHomepageImageLoadingTest.js
+++ b/test/legacyHomepageImageLoadingTest.js
@@ -1,0 +1,118 @@
+const puppeteer = require("puppeteer");
+
+function parseArgs(args) {
+  const parsedArgs = {};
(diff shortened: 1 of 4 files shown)


test/legacyHomepageImageLoadingTest.js
```diff
@@ -0,0 +1,100 @@
...
+  page.on("request", (request) => {
+    if (request.resourceType() === "image") {
+      // Log the URL of the image being requested
+      console.log("Image request:", request.url());
+      loadedImageUrls.push(request.url());
+    }
+    request.continue();
+  });
+
+  let homepage = "https://qa-digitalcollections.nypl.org";
```
EdwinGuzman, line 39: Everything before line 39 is the same in `test/homepageImageLoadingTest.js` and `test/homepageFeaturedImageLoadingTest.js. It'd be nice to see this refactored into one test script and the appropriate function (for legacy, swimlane, featured item) called so it's not three different files with similar code. Then, perhaps, the internal npm script can be 

```
    "homepageImageLoadingTestQa": "node test/imageLoadingTest.js --swimlane -- qa",
    "homepageImageLoadingTestProd": "node test/imageLoadingTest.js --swimlane --prod",
    "homepageFeaturedImageLoadingTestQa": "node test/imageLoadingTest.js --featuredImage -- qa",
    "homepageFeaturedImageLoadingTestProd": "node test/imageLoadingTest.js --featuredImage --prod"
    "legacyHomepageImageLoadingTestProd": "node test/imageLoadingTest.js --legacy --prod"
```

EdwinGuzman (approved):
I have a refactoring suggestion but that can (should?) be done later. I think it's more beneficial to merge this now to get some results instead of waiting to refactor this.

Non-blocking comment:
My bigger concern is that this test script lives in this repo. Work for "legacy" DC is now being done in the repo for the new DC. It's a nitpicky opinion but conceptually, I think it makes sense to have the scripts live in the repo they test.

However, I think this is here because it's easier to reuse the same js script and node/npm is already available in this repo whereas it's not for "legacy" DC. Is that right?

test/legacyHomepageImageLoadingTest.js
croyfish: I love this. If we want these tests to be maintained for the long term, we'll definitely do it.

PR #125, opened 2024-04-25
DR-2936: Point hero and header links to About page

7emansell:
## Ticket:

Resolves JIRA ticket [DR-2936](https://jira.nypl.org/browse/DR-2936)

- Points "Learn more" link in hero to the harmful content section of About page
- Points "What is public domain?" link in header to the public domain/copyright section of About page

## How has this been tested?

Locally

- Anchor links point to the correct place on the page, but often the text they're pointing to is covered by the sticky header. This is expected and will be addressed in [this ticket](https://jira.nypl.org/browse/DR-2911), it is not considered a blocker to launch by Clare

### Checklist:

### DIFF
--- a/src/components/hero/campaignHeroSubText.tsx
+++ b/src/components/hero/campaignHeroSubText.tsx
@@ -27,7 +27,7 @@ const CampaignHeroSubText = ({ featuredItem }: any) => {
               color: "var(--nypl-colors-ui-link-tertiary) !important",
             },
           }}
-          href="https://digitalcollections.nypl.org/about#nypl_harmful_content_statement"
+          href="/about#nypl_harmful_content_statement"
           aria-label="Learn more about harmful content"
           target="_blank"
         >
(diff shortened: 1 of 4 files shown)


src/components/hero/campaignHeroSubText.tsx
```diff
@@ -27,7 +27,7 @@ const CampaignHeroSubText = ({ featuredItem }: any) => {
               color: "var(--nypl-colors-ui-link-tertiary) !important",
             },
           }}
-          href="https://digitalcollections.nypl.org/about#nypl_harmful_content_statement"
+          href="/about#nypl_harmful_content_statement"
           aria-label="Learn more about harmful content"
           target="_blank"
```
avertrees, line 32: We can remove the `target="_blank"` in this link - Lana thinks this should open in the same tab on DC Facelift because the behavior in the old app is for this link to open in the same tab.

Probably a bug I introduced.

avertrees:
Looks good to me. Just one request to update the changelog and then this is good to go!

EdwinGuzman (approved):

PR #128, opened 2024-04-30
Hotfix/dr 2944 remove cache

avertrees:
## Ticket:

- JIRA ticket https://jira.nypl.org/browse/DR-2944

- removes code that caches of the v2/items/featured?random=true repo api response and thus breaks the featured item in prod 

-

### Checklist:


7emansell (approved):
Lgtm– noting that removing this locally doesn't have any effect bc it already doesn't do what this line intends in dev/QA environment, so we'll have to see what happens in prod

EdwinGuzman (approved):

PR #143, opened 2024-05-20
No ref/bug fixes

avertrees:
## Ticket:

- JIRA ticket [Fallback image for Featured Item](https://jira.nypl.org/browse/DR-2925)
- JIRA ticket [Update DS version on DC Facelift to v3.0](https://jira.nypl.org/browse/DR-2924)

comparing https://digital-collections-git-release-018-nypl.vercel.app/  to https://new-digitalcollections.nypl.org/, this PR:
- updates the "see more" links in the swim lanes to be blue (instead of black) and display a dotted line (instead of a solid line).
- updates the links in the "Explore Further" card title to have a visited state after clicking.

## How has this been tested?

locally

- 

## Other concerns or updates

- the "see more" links - I am having issues adding some padding/a margin to the dotted line below the links on hover to match the style that exists on prod.

### Checklist:

### DIFF
--- a/src/components/exploreFurther/exploreFurther.tsx
+++ b/src/components/exploreFurther/exploreFurther.tsx
@@ -73,6 +73,10 @@ const ExploreFurther = () => {
                 target="_blank"
                 rel="noreferrer noopener"
                 hasVisitedState
+                __css={{
+                  textDecoration: "none !important",
+                  _hover: { textDecoration: "underline 1px dotted !important" },
+                }}
               >
                 {item.title}
               </Link>
--- a/src/components/exploreFurther/exploreFurther.tsx
+++ b/src/components/exploreFurther/exploreFurther.tsx
@@ -72,7 +72,7 @@ const ExploreFurther = () => {
                 href={item.url}
                 target="_blank"
                 rel="noreferrer noopener"
-                hasVisitedState={true}
+                hasVisitedState
               >
                 {item.title}
               </Link>
--- a/src/components/exploreFurther/exploreFurther.tsx
+++ b/src/components/exploreFurther/exploreFurther.tsx
@@ -72,10 +72,7 @@ const ExploreFurther = () => {
                 href={item.url}
                 target="_blank"
                 rel="noreferrer noopener"
-                sx={{
-                  textDecoration: "none !important",
-                  _hover: { textDecoration: "underline 1px dotted !important" },
-                }}
+                hasVisitedState={true}
               >
                 {item.title}
               </Link>
--- a/src/components/featuredContent/featuredContent.tsx
+++ b/src/components/featuredContent/featuredContent.tsx
@@ -10,7 +10,6 @@ import { FeaturedContentData } from "@/types/FeaturedContentData";
 
 const FeaturedContentComponent = ({ randomNumber }) => {
   const data: FeaturedContentData = featuredContentData[randomNumber];
-  // console.log()
   return (
     <FeaturedContent
       data-testid={`featured-content-${randomNumber}`}
--- a/src/components/featuredContent/featuredContent.tsx
+++ b/src/components/featuredContent/featuredContent.tsx
@@ -10,6 +10,7 @@ import { FeaturedContentData } from "@/types/FeaturedContentData";
 
 const FeaturedContentComponent = ({ randomNumber }) => {
   const data: FeaturedContentData = featuredContentData[randomNumber];
+  // console.log()
   return (
     <FeaturedContent
       data-testid={`featured-content-${randomNumber}`}
--- a/src/components/hero/campaignHero.tsx
+++ b/src/components/hero/campaignHero.tsx
@@ -23,7 +23,7 @@ const CampaignHero = () => {
         response = await fetch(`/api/featuredItem`);
         responseData = await response.json();
       } catch (e) {
-        console.log("error: ", e);
+        console.log("CampaignHero error: ", e);
         console.log("using fallback featured item");
         responseData = defaultFeaturedItemResponse;
       }
--- a/src/components/hero/campaignHero.tsx
+++ b/src/components/hero/campaignHero.tsx
@@ -17,13 +17,21 @@ const CampaignHero = () => {
 
   useEffect(() => {
     const fetchData = async () => {
-      const response = await fetch(`/api/featuredItem`);
-      const responseData = await response.json();
+      let response;
+      let responseData: FeaturedItemData;
+      try {
+        response = await fetch(`/api/featuredItem`);
+        responseData = await response.json();
+      } catch (e) {
+        console.log("error: ", e);
+        console.log("using fallback featured item");
+        responseData = defaultFeaturedItemResponse;
+      }
       setData(responseData);
     };
 
     fetchData();
-  }, []);
+  }, [defaultFeaturedItemResponse]);
 
   const handleError = (e: any) => {
     console.log(e);
--- a/src/components/swimlanes/swimLanes.tsx
+++ b/src/components/swimlanes/swimLanes.tsx
@@ -35,7 +35,6 @@ const SwimLanes = ({ lanesWithNumItems }) => {
             color: "ui.blue",
             fontWeight: "500 !important",
             alignItems: "center",
-            // marginBottom: "1px",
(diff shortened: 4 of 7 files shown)


src/components/exploreFurther/exploreFurther.tsx
```diff
@@ -72,10 +72,7 @@ const ExploreFurther = () => {
                 href={item.url}
                 target="_blank"
                 rel="noreferrer noopener"
-                sx={{
-                  textDecoration: "none !important",
-                  _hover: { textDecoration: "underline 1px dotted !important" },
-                }}
+                hasVisitedState={true}
```
EdwinGuzman, line 75: Nitpicky and not a necessary change but more of an FYI: passing the prop automatically sets it to true for boolean props. So in this can, `hasVisitedState={true}` is the same as `hasVisitedState`.

src/components/featuredContent/featuredContent.tsx
```diff
@@ -10,6 +10,7 @@ import { FeaturedContentData } from "@/types/FeaturedContentData";
 
 const FeaturedContentComponent = ({ randomNumber }) => {
   const data: FeaturedContentData = featuredContentData[randomNumber];
+  // console.log()
```
EdwinGuzman, line 13: This can be removed.

src/components/hero/campaignHero.tsx
```diff
@@ -17,13 +17,21 @@ const CampaignHero = () => {
...
     const fetchData = async () => {
-      const response = await fetch(`/api/featuredItem`);
-      const responseData = await response.json();
+      let response;
+      let responseData: FeaturedItemData;
+      try {
+        response = await fetch(`/api/featuredItem`);
+        responseData = await response.json();
+      } catch (e) {
+        console.log("error: ", e);
```
EdwinGuzman, line 26: To track the error in the console, update the message to `"CampaignHero error: ", e)`

src/components/swimlanes/swimLanes.tsx
```diff
@@ -29,10 +29,14 @@ const SwimLanes = ({ lanesWithNumItems }) => {
...
           aria-label={`See more ${lane.title.toLowerCase()}`}
+          hasVisitedState={false}
           sx={{
             display: { sm: "none", md: "inline" },
-            fontWeight: "500",
+            color: "ui.blue",
+            fontWeight: "500 !important",
             alignItems: "center",
+            // marginBottom: "1px",
+            _hover: { textDecoration: "underline 1px dotted !important" },
```
EdwinGuzman, line 39: This is so odd. This should the default but it's not working. I tried making this `sx={{}}` to test and it throws the styles off. This is a bug in the DS but luckily you basically don't need any of this code.

Update this by:
- add `fontWeight="medium"` as a style prop
- remove the entire `sx` block.

You should have the desired result. I'll make a bug ticket for the empty sx value affecting styles though.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -109,6 +114,10 @@ const SwimLanes = ({ lanesWithNumItems }) => {
           "& svg": {
             marginTop: "1px",
           },
+          color: "ui.blue",
+          _hover: {
+            textDecoration: "underline 1px dotted !important",
+          },
```
EdwinGuzman, line 120: For this block, some styles are needed. Replace lines 108-120 from  `sx={...}`  to the following:

```
__css={{
          display: { sm: "flex", md: "none" },
          fontWeight: "medium",
          justifyContent: "flex-end",
          marginTop: "s",
}}
```

Related to the issue above, it looks like the `sx` prop is not working as intended. Since there are specific styles that are required, use `__css` instead.

src/components/swimlanes/swimLanes.tsx
avertrees: this fixes the text color and weight but doesn't fix the alignment issue of the text and the arrow.

EdwinGuzman: Is the default alignment considered "wrong"?

avertrees: i believe so, I'm just going off what is in prod dc but I'll ask lily

avertrees: got confirmation from lily that the default alignment is incorrect

EdwinGuzman: Okay, then I'd prefer if this is fixed in the DS so you don't have to add an override. Any thoughts @bigfishdesign13 ?

src/components/swimlanes/swimLanes.tsx
EdwinGuzman: yes, also, the `>` svg icon's fill color is black and not blue.... For some reason, the `Link` is breaking in this context and I can't tell why.

src/components/exploreFurther/exploreFurther.tsx
```diff
@@ -72,10 +72,7 @@ const ExploreFurther = () => {
                 href={item.url}
                 target="_blank"
                 rel="noreferrer noopener"
-                sx={{
-                  textDecoration: "none !important",
-                  _hover: { textDecoration: "underline 1px dotted !important" },
-                }}
```
EdwinGuzman, line 78: Seems like this is still needed, correct?

src/components/swimlanes/swimLanes.tsx
```diff
@@ -78,13 +81,7 @@ const SwimLanes = ({ lanesWithNumItems }) => {
                 <Text
                   id={`item-count-${lane.slug}-${index}`}
                   size="subtitle2"
-                  sx={{
-                    fontWeight: "400",
-                    display: "none",
-                    [`@media screen and (min-width: 600px)`]: {
-                      display: "inline",
-                    },
-                  }}
```
EdwinGuzman, line 87: This is needed because it hides this text on mobile and displays it on desktop. If this was intentional then ok, I just don't see where this was stated as an update.

For the font-weight, however, it was initially set as "400" so you could update it to `fontWeight: "regular"`.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -29,10 +29,13 @@ const SwimLanes = ({ lanesWithNumItems }) => {
           type="standalone"
           href={`${DC_URL}/collections/lane/${lane.slug}`}
           aria-label={`See more ${lane.title.toLowerCase()}`}
+          hasVisitedState={false}
           sx={{
             display: { sm: "none", md: "inline" },
-            fontWeight: "500",
+            color: "ui.blue",
```
EdwinGuzman, line 35: This should be `ui.primary.link`.

The bigger concern is that the styles for this "See more" link should be the same as the other "See more" link. This link is the desktop one since it is hidden on mobile (display: "none"). Use the same pattern as the other link and use `__css` to have the same styles and pattern.

src/pages/api/homepage.ts
```diff
@@ -27,7 +27,7 @@ const homepageHandler = async (
     });
 
     // 24 hour cache
-    response.setHeader("Cache-Control", "s-maxage=86400");
+    // response.setHeader("Cache-Control", "s-maxage=86400"); //removing bc the featured content is not
```
EdwinGuzman, line 30: Nitpick: complete the sentence to make it clear why it's removed. Delete instead of comment out as well.

src/components/swimlanes/swimLanes.tsx
```diff
@@ -100,15 +97,13 @@ const SwimLanes = ({ lanesWithNumItems }) => {
...
           display: { sm: "flex", md: "none" },
-          fontWeight: "500",
+          fontWeight: "medium",
           justifyContent: "flex-end",
           marginTop: "s",
-          alignItems: "center",
-          "& svg": {
-            marginTop: "1px",
-          },
+          alignItems: "center", //idk why this doesn't work
```
EdwinGuzman, line 106: I _think_ that's the alignment was done with `marginTop` on the SVG itself.

EdwinGuzman (approved):
Looks good! I can take a look at the Link issue within this certain context in June but it looks like the overrides are working well.

If you merge this branch in, then DR-2925 can be re-reviewed because the changes Bridgette is testing is not in that branch or Vercel preview until this is merged.

PR #144, opened 2024-05-20
Spike: Layout/breadcrumbs in Pages router

7emansell:
## Ticket:

Investigates [DR-2966](https://jira.nypl.org/browse/DR-2966).

- Creates collections and divisions directory structure
- Adds `Layout` component that contains header, DS provider, feedback, etc
- Adds `Layout` to all pages
- Creates a breadcrumbs context provider
- Adds breadcrumbs context to `/divisions/[slug]`, `/collections/[slug]`, `/divisions`, `/collections`

### Checklist:

### DIFF
--- a/src/pages/collections/[id].tsx
+++ b/src/pages/collections/[id].tsx
@@ -1,24 +1,36 @@
+import { useContext, useEffect } from "react";
+import { useRouter } from "next/router";
 import Layout from "@/components/layout/layout";
-
 import { Heading } from "@nypl/design-system-react-components";
-import Head from "next/head";
+import { breadcrumbsContext } from "../breadcrumbsContext";
 
-export default function Collection() {
+const Collection = () => {
+  const { breadcrumbs, setBreadcrumbs } = useContext(breadcrumbsContext);
+  useEffect(() => {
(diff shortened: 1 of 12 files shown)


src/pages/collections/[id].tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <Layout activePage="collection" breadcrumbs={breadcrumbs}>
+      <Heading>
+        <>Collection</>
+      </Heading>
+    </Layout>
+  );
+};
+
+// Imagine an API call somewhere in here
```
Comment (EdwinGuzman) [inline src/pages/collections/[id].tsx:20]: Have you tried passing this mocked API-fetched data to the component? How would this change if you use this `breadcrumbs` array as a prop in `Collection`, i.e. `const Collection = ({ breadcrumbs }) => {....`.

You could use the last object in the array as the value that is passed to `setBreadcrumbs`.

Comment (7emansell) [reply src/pages/collections/[id].tsx:20]: Right, sorry for the confusion of trying to represent two approaches within the same file but yes imagining that as a prop to `Collections`. If that's the case though, I wouldn't use the context at all

PR #154, opened 2024-06-18
DR-3011: Migrating API endpoints to App router

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3011](https://newyorkpubliclibrary.atlassian.net/browse/DR-3011).

- Moves `pages/api` endpoints over to App Router `app/api/.../route.tsx` files
- Updates endpoints to use `NextRequest` and `NextResponse`
- Creates separate config for API tests (`npm run test:API`)
- Removes wrapping `src` folder

## How has this been tested?

Locally

### Checklist:


app/api/feedback/route.tsx
```diff
@@ -0,0 +1,74 @@
...
+      },
+      { status: 200 }
+    );
+  } catch (e) {
+    return NextResponse.json(
+      {
+        message: e.message,
+        error: "server error",
+      },
+      { status: 500 }
```
EdwinGuzman, line 71: A more appropriate code is 400 since technically the server itself did not error out but there was an error with the google sheets request.

EdwinGuzman: I know it's not related to this PR so you can update this later as well.

EdwinGuzman (commented):
Code looks good, but the main hero image is to be cached and does not render a new one on every page refresh. It might be related to the App router's caching strategy which I think is quite strict.

EdwinGuzman (changes requested):
Some updates:
- update the changelog
- perhaps you can have one "main" test npm script that calls both scripts? is that too much?
- update/add documentation in the readme to include the new npm script that should be run for testing.

README.md
```diff
@@ -399,16 +373,24 @@ Note: Do not need to worry about Quality or Format for this application.
 
 The Digital Collections app runs unit tests with Jest and React Testing Library.
 
+We have two groups of tests: pages/components have unit tests in the `jsdom` environment, and the API endpoints are tested in the `node` environment.
```
EdwinGuzman, line 376: Can you add that the pages/components and `jsdom` tests are for the react/client side tests and the API/`node` env tests are for the functional ones and mostly due to how the app is structured in Next (because we _have_ to do this in the new Next.js App router architecture, right?)

EdwinGuzman:
Looks good. The caching still happens on Vercel and locally. Is this something we noticed before or only in dev? (will this be a problem in the qa-new?)

EdwinGuzman (approved):
Great!

avertrees:
looks good!

avertrees (approved):

PR #155, opened 2024-06-20
DR-3036: Card column breakpoints

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3036](https://newyorkpubliclibrary.atlassian.net/browse/DR-3036)

- Adds new column breakpoints to the home page swimlanes

### Checklist:

### DIFF
--- a/app/components/swimlanes/swimLanes.tsx
+++ b/app/components/swimlanes/swimLanes.tsx
@@ -19,7 +19,18 @@ import { DC_URL } from "../../config/constants";
 import useBreakpoints from "../../hooks/useBreakpoints";
 
 const SwimLanes = ({ lanesWithNumItems }) => {
-  const { isLargerThanLargeTablet } = useBreakpoints();
+  const {
+    isLargerThanLargeTablet,
+    isLargerThanSmallTablet,
+    isLargerThanLargeMobile,
+  } = useBreakpoints();
+  const numColumns = isLargerThanLargeTablet
+    ? 4
+    : isLargerThanSmallTablet
+    ? 2
+    : isLargerThanLargeMobile
+    ? 2
+    : 2;
   return lanesWithNumItems.map((lane, key) => (
     <Box className={styles.lane} data-testid={lane.slug} mt="xxl" key={key}>
       <Flex alignItems="baseline">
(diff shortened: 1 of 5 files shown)


EdwinGuzman (changes requested):
Only change is to try to figure out the flicker situation for the desktop view.

app/components/swimlanes/swimLanes.tsx
```diff
@@ -19,7 +19,12 @@ import { DC_URL } from "../../config/constants";
 import useBreakpoints from "../../hooks/useBreakpoints";
 
 const SwimLanes = ({ lanesWithNumItems }) => {
-  const { isLargerThanLargeTablet } = useBreakpoints();
+  const { isLargerThanLargeTablet, isLargerThanLargeMobile } = useBreakpoints();
+  const getNumColumns = () => {
+    if (isLargerThanLargeTablet) return 4;
+    if (isLargerThanLargeMobile) return 2;
+    return 1;
```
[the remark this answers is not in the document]
7emansell: Passing it down as a prop is the only thing that resolves the flicker (not setting it as state in a `useEffect`, or reversing the logic) but it does kind of seem like overkill

app/components/swimlanes/swimLanes.tsx
```diff
@@ -16,9 +16,9 @@ import {
 import styles from "./Swimlanes.module.css";
 import { imageURL } from "../../utils/utils";
 import { DC_URL } from "../../config/constants";
-import useBreakpoints from "../../hooks/useBreakpoints";
+import useBreakpoints from "app/hooks/useBreakpoints";
 
-const SwimLanes = ({ lanesWithNumItems }) => {
+const SwimLanes = ({ numColumns, lanesWithNumItems }) => {
   const { isLargerThanLargeTablet } = useBreakpoints();
```
EdwinGuzman, line 22: You don't need to import `useBreakpoints` here anymore.

EdwinGuzman (approved):
Minor, remove the unnecessary import in `swimLanes.tsx`.

This works better but I get that it's not ideal and the cost is low. This is something I think Marty and I should get working properly in the DS and allow components to be more flexible with responsive styles. And also to try to move away from the `useNYPLBreakpoint`/`useBreakpoint` hook because we've noticed it causes a noticeable flicker.

7emansell:
Would it be better to do it as media queries entirely? @EdwinGuzman

app/components/swimlanes/swimLanes.tsx
7emansell: It's used in another place

EdwinGuzman: Ah, the tooltip.

EdwinGuzman:
Yes, I think it's a good case where CSS can and should be able to handle the media queries and responsive styles. It _could_ work with a CSS module but not sure if you want to introduce that now. I'd prefer if devs can add the chakra responsive style object but `SimpleGrid` does not currently allow that.

I'm okay with this approach if you are.

avertrees (approved):

PR #158, opened 2024-06-25
DR-3022: Page layout component and metadata

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3022](https://newyorkpubliclibrary.atlassian.net/browse/DR-3022).

- Creates custom (client side) page layout component with `activePage`, `breadcrumbs` and `DSProvider`, `FeedbackBox`, etc.
   - Moves Adobe Analytics into this as well
- Moves home and about page contents into client components
- Adds `PageLayout` to all pages' corresponding client component
- Adds Next's `Metadata` to all pages (bc now they're server components!)
   - Dynamically generating metadata title tag from url on `/collections/lane/[slug]` and `/divisions/[slug]`– also using this title for the breadcrumbs. Once we are actually fetching page data, this won't be how we generate the breadcrumbs. Or maybe it'll be some hybrid of these, whatever performs better
   - Statically generating title tag on `/collections` and `/divisions`
- Adds router mock to home and about pages' tests

## How has this been tested?

Locally (necessary to see breadcrumbs)

### Checklist:

### DIFF
--- a/__tests__/pages/aboutpage.test.tsx
+++ b/__tests__/pages/aboutpage.test.tsx
@@ -1,7 +1,12 @@
-import About from "../../app/about/page";
 import { render } from "@testing-library/react";
+import About from "app/about/page";
 import { axe } from "jest-axe";
 import React from "react";
+
+jest.mock("next/navigation", () => ({
+  useRouter: jest.fn(),
+  usePathname: jest.fn(),
+}));
 
 describe("About page accessibility", () => {
   it("passes axe accessibility test", async () => {
--- a/__tests__/pages/aboutpage.test.tsx
+++ b/__tests__/pages/aboutpage.test.tsx
@@ -12,5 +12,5 @@ describe("About page accessibility", () => {
   it("passes axe accessibility test", async () => {
     const { container } = render(<About />);
     expect(await axe(container)).toHaveNoViolations();
-  }, 15000);
+  }, 60000);
 });
--- a/app/about/page.tsx
+++ b/app/about/page.tsx
@@ -1,10 +1,10 @@
 import AboutPage from "app/components/aboutPage/aboutPage";
+import { Metadata } from "next";
+
+export const metadata: Metadata = {
+  title: "About NYPL Digital Collections",
+};
 
 export default function About() {
-  return (
-    <>
-      <title>About NYPL Digital Collections</title>
-      <AboutPage />
-    </>
-  );
+  return <AboutPage />;
 }
--- a/app/about/page.tsx
+++ b/app/about/page.tsx
@@ -1,35 +1,10 @@
-"use client";
-import { Box } from "@nypl/design-system-react-components";
-import aboutPageElements from "../data/aboutPageElements";
-import React from "react";
+import AboutPage from "app/components/aboutPage/aboutPage";
 
 export default function About() {
-  function createSection(heading: React.JSX.Element, body: React.JSX.Element) {
-    return (
-      <>
-        {heading}
-        {body}
-      </>
-    );
-  }
-
   return (
     <>
       <title>About NYPL Digital Collections</title>
-      <Box
-        id="mainContent"
-        sx={{
-          maxWidth: "820px",
-          margin: "s",
-          [`@media screen and (min-width: 820px)`]: {
-            margin: "auto",
-          },
-        }}
-      >
-        {aboutPageElements.map((section) =>
-          createSection(section.heading, section.body)
-        )}
-      </Box>
+      <AboutPage />
     </>
   );
 }
--- a/app/about/page.tsx
+++ b/app/about/page.tsx
@@ -1,4 +1,4 @@
-import AboutPage from "app/components/aboutPage/aboutPage";
+import AboutPage from "@/components/pages/aboutPage/aboutPage";
 import { Metadata } from "next";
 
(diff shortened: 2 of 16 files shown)


app/about/page.tsx
```diff
@@ -1,35 +1,10 @@
-"use client";
-import { Box } from "@nypl/design-system-react-components";
-import aboutPageElements from "../data/aboutPageElements";
-import React from "react";
+import AboutPage from "app/components/aboutPage/aboutPage";
+import { Metadata } from "next";
```
avertrees, line 2: wait - I thought Metadata only worked on server components. Does this work with client components too or does this mean we're using server components now? (eyes emoji)

7emansell: read PR description <3

avertrees: I did and that's why I left a comment but I'm still confused! More surprised, actually. This feels like a big deal?

avertrees: I thought we couldn't use the DS with server components

avertrees: nvm, jumped the gun a bit. I see how this works now

avertrees (approved):

__tests__/pages/aboutpage.test.tsx
```diff
@@ -1,11 +1,16 @@
...
+  useRouter: jest.fn(),
+  usePathname: jest.fn(),
+}));
+
 describe("About page accessibility", () => {
   it("passes axe accessibility test", async () => {
     const { container } = render(<About />);
     expect(await axe(container)).toHaveNoViolations();
-  }, 15000);
+  }, 60000);
```
EdwinGuzman, line 15: Oiy, that's a big timeout increase.

EdwinGuzman (commented):
How do you feel about making a directory under `components` named `pages` for the about and home page files? Later on it can include the other page files. I can see it becoming a big folder with subfolders for each page. It might or might not be an actual problem but if it does, you can always make the change later.

app/collections/lane/[slug]/page.tsx
```diff
@@ -1,10 +1,36 @@
...
+import PageLayout from "app/components/pageLayout/pageLayout";
 
-export default function Lane() {
+type LaneProps = {
+  params: { slug: string };
+};
+
+export async function generateMetadata(
+  { params }: LaneProps,
+  parent: ResolvingMetadata
```
Comment (EdwinGuzman) [inline app/collections/lane/[slug]/page.tsx:11]: Do you need the parent arg in the `generateMetadata` functions?

app/components/pageLayout/pageLayout.tsx
```diff
@@ -0,0 +1,110 @@
...
+}: PropsWithChildren<PageLayoutProps>) => {
+  const pathname = usePathname();
+  // Track page view events to Adobe Analytics
+  useEffect(() => {
+    /*
+    "asPath has been removed because the concept of as has been removed from the new router."
+    https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration
+    */
+    console.log("pathname", pathname);
+    const route = pathname ? pathname : "";
```
EdwinGuzman, line 36: You should also be able to do `pathname || ""`

EdwinGuzman: Or perhaps `??` instead.

app/components/pageLayout/pageLayout.tsx
```diff
@@ -0,0 +1,110 @@
...
+  breadcrumbs,
+}: PropsWithChildren<PageLayoutProps>) => {
+  const pathname = usePathname();
+  // Track page view events to Adobe Analytics
+  useEffect(() => {
+    /*
+    "asPath has been removed because the concept of as has been removed from the new router."
+    https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration
+    */
+    console.log("pathname", pathname);
```
EdwinGuzman, line 35: Remove logs

app/components/pageLayout/pageLayout.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const route = pathname ? pathname : "";
+    trackVirtualPageView(route);
+  });
+  const [view, setView] = React.useState("form");
+  const { onOpen, isOpen, onClose, FeedbackBox } = useFeedbackBox();
```
EdwinGuzman, line 40: Not part of this PR but `onOpen` is not being used.

EdwinGuzman (approved):
Great to see app router features!

As an aside: the DS needs to make an update to the `Breadcrumbs` component to allow the Next.js `Link` component instead of a regular link. It's currently making full page refreshes which is not a problem but it'd be nice to use the in-app router.

Aside 2: I see this warning in the console
`VM26917:106 NYPL Reservoir FeaturedContent: A value for 'imageProps.alt' is required. `  but it's also in prod so don't fix this now. Create a ticket to clean up the warnings that get logged.

EdwinGuzman:
Ticket for breadcrumbs routing https://newyorkpubliclibrary.atlassian.net/browse/DSD-1797

__tests__/pages/aboutpage.test.tsx
7emansell: They don't take this long locally but I just want the test hook to not fail bc of timeout

PR #160, opened 2024-06-26
DR-3027 create collection card component

avertrees:
## Ticket:

- JIRA ticket [DR-3027](https://newyorkpubliclibrary.atlassian.net/browse/DR-3027)

- creates a CollectionCard component using the card in the swimLanes component
- creates a CollectionCard type interface and CollectionCardModel class (this may be a little redundant, thinking of refactoring later)
- creates LaneData type interface 

## How has this been tested?

locally by making sure cards load as expected

- zilch

### Checklist:

[DR-3027]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


7emansell (commented):
Overall looks great, minor comments! Also I think there will be some style changes needed (particularly the `LargerThanLargeTablet` prop) once the card column breakpoints (#155) is merged in so maybe wait on that/ merge it in? It's still in VQA so I'm not sure how close that is to what we'll actually have

app/types/CollectionCard.ts
```diff
@@ -0,0 +1,12 @@
+export interface CollectionCardData {
```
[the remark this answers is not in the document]
avertrees: True.. think it would be more clear but I was just following the pattern already in place with the following types: FeaturedItemData, ExploreFurtherData, and FeaturedContentData . If we add ...Type to the new types, I think we should also add them the existing ones - do you agree?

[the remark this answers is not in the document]
avertrees: also wondering if we should have `props` directory in the types folder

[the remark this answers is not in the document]
EdwinGuzman: Try the `props` directory? It's a reversible small update so you can always try it to see if it's helpful.

Personally, I like to keep the prop type/interface in the same file as the component for readability. It's the convention in the DS and maybe other projects.

app/components/cards/collectionCard.tsx
```diff
@@ -0,0 +1,73 @@
+"use client";
+import React from "react";
+import {
+  Card,
+  CardHeading,
+  Text,
+  CardContent,
+  Tooltip,
+} from "@nypl/design-system-react-components";
+import styles from "../swimlanes/Swimlanes.module.css";
```
EdwinGuzman, line 10: Ah, start creating separate CSS module files for each component. So there should be a new one for `CollectionCard` to better scope styles to each component.

app/types/CollectionCard.ts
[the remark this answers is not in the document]
7emansell: I agree we should add ...Type to everything. `props` directory is up to you if you think it'd be good @avertrees

app/components/cards/collectionCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </Text>
+      </CardContent>
+    </Card>
+  );
+};
+
+export default CollectionCard;
+{
+  /*
+<SimpleGrid
```
EdwinGuzman, line 79: Remove the comments.

app/components/cards/collectionCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+        )}
+      </CardHeading>
+      <CardContent>
+        <Text
+          id={`item-count-${slug}-${index}`}
+          size="subtitle2"
+          fontWeight="medium"
+          __css={{
+            display: "none",
+            [`@media screen and (min-width: 400px)`]: {
```
EdwinGuzman, line 63: Use the breakpoint values found in `app/utils/breakpoints.ts`. In this case I think this should be `lgMobile`.

EdwinGuzman: The bigger problem is that `CollectionCard` is composed to replaced the `Card` component in `swimLanes.tsx` and so far it's coming along well. But when you get to replacing `Card` in `exploreFurther.tsx` with `CollectionCard`, it won't be a direct swap because the heading and content are _slightly_ different.

We can chat about some approaches tomorrow.

app/components/cards/collectionCard.tsx
avertrees: oop

app/components/cards/collectionCard.tsx
avertrees: I don't think CollectionCard should be used to replace the Card component in exploreFurther - I wasn't planning on swapping them out either but if we want to make a new ticket for that we can.

avertrees: I think CollectionCard should represent a Collection - the Explore Further section does not display Collections

EdwinGuzman (commented):
Latest round of updates look good but create a new CSS modules file for `CollectionCard` component.

EdwinGuzman (approved):

PR #161, opened 2024-06-26
DR-3028 create url structure for phase 3

avertrees:
## Ticket:

- JIRA ticket [DR-3028](https://newyorkpubliclibrary.atlassian.net/browse/DR-3028)

creates url structure for the following pages:
 - `/search/index`
 - `/collections/[slug]`

also accepts query parameters and renders query param on the body

## How has this been tested?

locally

-

### Checklist:

[DR-3028]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/collections/[slug]/page.tsx
+++ b/app/collections/[slug]/page.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { Metadata } from "next";
 import PageLayout from "app/components/pageLayout/pageLayout";
-import Collections from "app/components/collections/collections";
+import CollectionsPage from "app/components/collections/collections";
 import { useSearchParams } from "next/navigation";
 
 type CollectionProps = {
@@ -30,7 +30,7 @@ export default function Lane({ params }: CollectionProps) {
         },
       ]}
     >
-      <Collections slug={params.slug} />
+      <CollectionsPage slug={params.slug} />
     </PageLayout>
   );
 }
--- a/app/collections/[slug]/page.tsx
+++ b/app/collections/[slug]/page.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+import { Metadata } from "next";
+import PageLayout from "app/components/pageLayout/pageLayout";
+import Collections from "app/components/collections/collections";
+import { useSearchParams } from "next/navigation";
+
+type CollectionProps = {
+  params: { slug: string };
+};
+
+export async function generateMetadata({
+  params,
+}: CollectionProps): Promise<Metadata> {
+  const slug = params.slug;
+  return {
+    title: `${slug} - NYPL Digital Collections`,
+  };
+}
+
+export default function Lane({ params }: CollectionProps) {
+  return (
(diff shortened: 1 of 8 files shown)


7emansell (approved):
Looks good to me!!

app/collections/[slug]/page.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
Comment (EdwinGuzman) [inline app/collections/[slug]/page.tsx:1]: There's currently an `app/collections/lane/[slug]/page.tsx` file. What's the difference between that route and this route?

app/collections/[slug]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+      breadcrumbs={[
+        { text: "Home", url: "/" },
+        { text: "All Collections", url: "/collections" },
+        {
+          text: `${params.slug}`,
+          url: `/collections/${params.slug}`,
+        },
+      ]}
+    >
+      <Collections slug={params.slug} />
```
Comment (EdwinGuzman) [inline app/collections/[slug]/page.tsx:33]: NItpick: I'd rename this as `CollectionsPage` so it's not confusing if there's a component that is a list of collections called `Collections`. This current component includes the search and other elements on the page so that's why I'm thinking `CollectionsPage` might be better.

app/components/search/body.tsx
```diff
@@ -0,0 +1,14 @@
+import { useSearchParams } from "next/navigation";
+
+const SearchBody = () => {
```
EdwinGuzman, line 3: Nitpick: What do you think about "SearchContent" rather than "SearchBody"?

app/components/search/results.tsx
```diff
@@ -0,0 +1,18 @@
...
+import SearchBody from "./body";
+
+function SearchBarFallback() {
+  return <>placeholder</>;
+}
+
+const SearchResults = ({}) => {
+  return (
+    <>
+      <Suspense fallback={<SearchBarFallback />}>
```
EdwinGuzman, line 12: What was the issue with using suspense before and how did you resolve it?

app/components/search/body.tsx
avertrees: fine with me

app/components/search/results.tsx
avertrees: `useSearchParams()` needs to be wrapped in a `Suspense`, which means the  code that uses `useSearchParams()` needs to be in it's own component and wrapped by the `Suspense` tags. there's a config we could use to get around this but it's not recommended by Next.js and will not be supported in future versions. 

see: https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

EdwinGuzman (approved):

PR #162, opened 2024-06-26
DR-3029 create /items/:uuid page

avertrees:
## Ticket:

- JIRA ticket [DR-3029](https://newyorkpubliclibrary.atlassian.net/browse/DR-3029)

create url structure for `/items/:uuid` pages
- conditionally renders a Clover IIIF component on each page, based on the `typeOfResource` returned by Repo API 
- Single capture renders the Clover openseadragon viewer (doesn't need manifests)
- Multi capture renders Clover IIIF Viewer using sample metadata
- Audio renders Clover Audio viewer example using sample metadata
- Video renders Clover Video viewer example using sample metadata
- PDF/Book renders the Book viewer while I try to figure out how the PDF example is supposed to work

## How has this been tested?

Locally and in vercel (need to run `npm install` locally)

Examples: 

Single Capture: 
/items/510d47dc-8a69-a3d9-e040-e00a18064a99

Multi Capture: 
/items/70bcc910-d5c9-0134-fbe5-00505686d14e

Video:
/items/a026eb76-e762-46a1-8edf-591f4b419f0f

Audio:
/items/2b4b1010-c32b-0133-b90f-60f81dd2b63c

PDF: 
/items/8fcbf960-fed9-0130-73f0-58d385a7bbd0

- N/A rn

## Open Questions

Is there a better way to get the Item model accessible to the Next Metadata object and the component?

### Checklist:

[DR-3029]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman:
Nitpick: include in your PR description that you need to run `npm install` since there's a new package.

This is great! I haven't reviewed the code yet but why not remove the redirect that's in place just to show it off and demo internally on the vercel instance?

app/components/items/clover/audio/viewer.tsx
```diff
@@ -0,0 +1,28 @@
+"use client";
```
EdwinGuzman, line 1: I haven't checked the clover-iiif docs but this has to be a client component, right?

app/components/items/item.tsx
```diff
@@ -0,0 +1,68 @@
+import { getItemData } from "@/utils/utils";
+import { ItemModel } from "@/models/item";
+import ImageViewer from "./clover/image/image";
+import CloverImageViewer from "./clover/image/viewer";
+import AudioViewer from "./clover/audio/viewer";
+import VideoViewer from "./clover/video/viewer";
+import BookViewer from "./clover/book/viewer";
+import PDFViewer from "./clover/pdf/viewer";
+
+const getItem = async (uuid) => {
```
EdwinGuzman, line 10: Looks like this function isn't being used in this file (and instead it's a prop which is great).

app/items/[uuid]/page.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ItemModel } from "@/models/item";
+
+type ItemProps = {
+  params: {
+    uuid: string;
+    item: ItemModel;
+  };
+};
+
+const getItem = async (uuid) => {
```
Comment (EdwinGuzman) [inline app/items/[uuid]/page.tsx:15]: Regarding your question in the description, looks like it's a common problem and request: https://github.com/vercel/next.js/discussions/50080

It doesn't seem like a big problem though but it's unfortunate.

app/items/[uuid]/page.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const item = await getItemData(uuid);
+  return item;
+};
+
+export async function generateMetadata({
+  params,
+}: ItemProps): Promise<Metadata> {
+  const data = await getItem(params.uuid);
+  const item = new ItemModel(data);
```
Comment (EdwinGuzman) [inline app/items/[uuid]/page.tsx:24]: I think it's okay to move this line inside of the `getItem` function and return it. The `getItemData` function itself is already doing the fetching so `getItem` is really a wrapper but not doing anything else. So you could have done, with the same result, 
```
const data = await getItemData(params.uuid);
const item = new ItemModal(data);
```

If you do move line 24 inside of `getItem`, it could be renamed to reflect what it's doing such as `getItemModel`.

app/utils/utils.ts
```diff
@@ -98,7 +109,38 @@ export const apiCall = async (
...
+    } else {
+      return undefined;
+    }
+  } catch (error) {
+    return undefined;
+  }
+};
+
+/**
+ * Returns Repo API response WITH request data.
```
EdwinGuzman, line 128: Why do you need to do this?

I'm not against this, but you can refactor these two functions so that the bulk of the work is not repeated. The base function can have `return data`. Then the other function can call the base function and return `data.nyplAPI.response`.

app/models/item.ts
```diff
@@ -0,0 +1,22 @@
...
+  capture: any;
+  typeOfResource: string;
+  title: string;
+  isPDF: boolean;
+  isBook: boolean;
+  isSingleCapture: boolean;
+
+  constructor(data: any) {
+    this.uuid = data.nyplAPI.request.uuid;
+    this.mods = data.nyplAPI.response.mods;
```
EdwinGuzman, line 13: This answers my question regarding having two API functions, you need both `request` and `response`. Very nitpicky but you can do `const { nyplAPI } = data;` just to not repeat `data.nyplAPI`, or even
`const { nyplAPI: { request, response } } = data;` so you can use `request` and `response` directly.

app/components/items/item.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const item = await getItemData(uuid);
+  return item;
+};
+
+const Item = ({ item }) => {
+  // const data = await getItem(uuid);
+  // const item = new ItemModel(data);
+
+  const viewer = () => {
```
EdwinGuzman, line 19: This doesn't need to be a function. If it's a `let` variable, in each `case` you can do something such as
```
viewer  = (<> .... </>)
return;
````
(just because cases should return even undefined). And then on line 65 `<>{viewer}</>` so you don't have to invoke a function call.

Comment (EdwinGuzman) [reply app/components/items/item.tsx:19]: This is also partly related to the React component life cycle and on every rerender, the value will be assigned and rendered so you don't have to make a function call.

**Review by EdwinGuzman — CHANGES_REQUESTED** [review]
Minor updates and questions

File: app/utils/utils.ts (hunk 5 above)
Comment (avertrees) [reply app/utils/utils.ts:128]: one returns the whole api response and the other returns  `data.nyplAPI.response.` - wanted the request object from the original response to get the uuid and the pagination information.

**Comment by avertrees** [comment]
> This is great! I haven't reviewed the code yet but why not remove the redirect that's in place just to show it off and demo internally on the vercel instance?

@EdwinGuzman by remove the redirect you mean change the `APP_ENV` in the Vercel to be `development`, right?

**Comment by EdwinGuzman** [comment]
> change the APP_ENV in the Vercel to be development, right?

Not exactly. If you do this, then all the new routes will be exposed. I was thinking you could not add the route in the middleware file to only expose this specific route. But if there's no harm in exposing all routes in vercel, then yes, you can change `APP_ENV` to `development`.

File: app/components/items/clover/audio/viewer.tsx (hunk 1 above)
Comment (avertrees) [reply app/components/items/clover/audio/viewer.tsx:1]: i didn't see it mentioned in the clover-iiif docs, but I ran into issues trying to render it in the app router without "use client"

File: app/models/item.ts (hunk 6 above)
Comment (avertrees) [reply app/models/item.ts:13]: mmm the code above is not working for me?

Comment (avertrees) [reply app/models/item.ts:13]: nvm... got it to work

File: app/components/items/item.tsx (hunk 7 above)
Comment (avertrees) [reply app/components/items/item.tsx:19]: I'm having trouble rendering the component with: 

`viewer = (<> .... </>)
 return;`
 
 but if I do this in each case, the component renders and I can remove line 65
 
`viewer = (<> .... </>)
 return viewer;`

File: app/utils/utils.ts (hunk 5 above)
Comment (EdwinGuzman) [reply app/utils/utils.ts:128]: Right, that's ok. But you're doing the base work twice. I think you can keep `RepoAPICall` as is since that's where the `fetch` call is being made and let it return the base `data` response object. In `apiCall`, you don't need most of the code. You can do

```
export const apiCall = async (
  apiUrl: string,
  urlParam?: { [key: string]: any }
) => {
  const data = await RepoAPICall(apiUrl, urlParam);

  return data?.nyplAPI?.response;
};
```

- In your current `apiCall` function, you don't need to wrap `await RepoAPICall` in a try/catch or check for status because that's already done in the `RepoAPICall` since that function calls fetch. The error handling occurs in that function and if there is an error, it will return undefined. That's why in this function, you can just check for undefined values with the `?.` optional chaining.
- `RepoAPICall` returns the `data` object and `apiCall` returns the `response` property from the `nyplAPI` property. The names of the functions could be clearer.

File: app/components/items/item.tsx [hunk 8]
```diff
@@ -0,0 +1,58 @@
...
+      //also PDF
+      viewer = (
+        <>
+          <h2> Book: {item.title} </h2>
+          <BookViewer />
+        </>
+      );
+      return viewer;
+    default:
+      return <h1>No type of resource match</h1>;
```
EdwinGuzman, line 54: Just an FYI: this is not a necessary update for this PR but in the future, there should only be one `h1` on the page and I don't think it should come from this component. I think that should be set at the page level so this component can just render the viewer or an error message without having to worry about the heading hierarchy of the page.

EdwinGuzman:
The only outstanding update is in `utils.ts` https://github.com/NYPL/digital-collections/pull/162/files#r1658707883

app/utils/utils.ts
EdwinGuzman: Yes, that makes sense.

EdwinGuzman (approved):

PR #163, opened 2024-06-28
DR-2867: Enable strict typing

7emansell:
## Ticket:

Resolves JIRA ticket [DR-2867](https://newyorkpubliclibrary.atlassian.net/browse/DR-2867)

- Turns on `strict` flag for Typescript
- Adds types accordingly 
- Moves component prop interfaces into component files (this is maybe not the pattern we want though? Could also be in a `/props` folder

## How has this been tested?

Locally

### Checklist:

[DR-2867]: https://newyorkpubliclibrary.atlassian.net/browse/DR-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/data/swimlaneProps.tsx
+++ b/__tests__/data/swimlaneProps.tsx
@@ -1,5 +1,7 @@
-export const props = {
-  randomNumber: 1,
+import { SwimLanesProps } from "@/components/swimlanes/swimLanes";
+import type { CollectionItem } from "@/components/swimlanes/swimLanes";
+
+export const props: SwimLanesProps = {
   lanesWithNumItems: [
     {
       title: "Test collections 1",
@@ -34,7 +36,7 @@ export const props = {
           image_id: "58591658",
           numItems: 55,
         },
-      ],
+      ] as CollectionItem[],
     },
     {
       title: "Test collections 2",
--- a/__tests__/data/swimlaneProps.tsx
+++ b/__tests__/data/swimlaneProps.tsx
@@ -36,7 +36,7 @@ export const props: SwimLanesProps = {
           image_id: "58591658",
           numItems: 55,
         },
-      ] as CollectionItem[],
+      ],
     },
     {
       title: "Test collections 2",
--- a/__tests__/data/swimlaneProps.tsx
+++ b/__tests__/data/swimlaneProps.tsx
@@ -1,5 +1,4 @@
 import { SwimLanesProps } from "@/components/swimlanes/swimLanes";
-import type { CollectionItem } from "@/components/swimlanes/swimLanes";
 
 export const props: SwimLanesProps = {
   lanesWithNumItems: [
--- a/app/api/feedback/route.tsx
+++ b/app/api/feedback/route.tsx
@@ -66,13 +66,19 @@ export const POST = async (request: NextRequest) => {
     );
   } catch (e: unknown) {
     if (e instanceof Error) {
-      return NextResponse.json({
-        message: e.message,
-      });
+      return NextResponse.json(
+        {
+          message: e.message,
+        },
+        { status: 400 }
+      );
     } else {
-      return NextResponse.json({
-        message: "unknown error occurred",
-      });
+      return NextResponse.json(
+        {
+          message: "unknown error occurred",
+        },
+        { status: 400 }
+      );
     }
   }
 };
--- a/app/api/feedback/route.tsx
+++ b/app/api/feedback/route.tsx
@@ -13,7 +13,9 @@ export const POST = async (request: NextRequest) => {
     const origin = request.headers.get("origin") || "";
     const page = referer.replace(origin, "");
     const ipAddress = request.headers.get("x-forwarded-for") || "";
-    const userAgentParser = new UAParser(request.headers.get("user-agent"));
+    const userAgentParser = new UAParser(
+      request.headers.get("user-agent") ?? undefined
+    );
     const userAgent = userAgentParser.getResult();
     const userPlatform = userAgent.device.model || "";
     const userBrowser = userAgent.browser.name || "";
@@ -62,13 +64,15 @@ export const POST = async (request: NextRequest) => {
       },
       { status: 200 }
     );
-  } catch (e) {
-    return NextResponse.json(
-      {
+  } catch (e: unknown) {
+    if (e instanceof Error) {
+      return NextResponse.json({
         message: e.message,
-        error: "server error",
-      },
-      { status: 400 }
-    );
+      });
+    } else {
+      return NextResponse.json({
+        message: "unknown error occurred",
+      });
+    }
   }
 };
--- a/app/components/hero/campaignHeroSubText.tsx
+++ b/app/components/hero/campaignHeroSubText.tsx
@@ -1,4 +1,3 @@
-import { FeaturedContentData } from "@/types/FeaturedContentData";
 import useBreakpoints from "../../hooks/useBreakpoints";
 import {
   Link as DSLink,
--- a/app/components/hero/campaignHeroSubText.tsx
+++ b/app/components/hero/campaignHeroSubText.tsx
@@ -1,3 +1,4 @@
+import { FeaturedContentData } from "@/types/FeaturedContentData";
 import useBreakpoints from "../../hooks/useBreakpoints";
 import {
   Link as DSLink,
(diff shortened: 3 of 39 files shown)


app/utils/utils.ts
```diff
@@ -221,6 +232,7 @@ export const adobeAnalyticsRouteToPageName = (route = "", queryParams = "") => {
  * Tracks a virtual page view to Adobe Analytics on page navigation.
  */
 export const trackVirtualPageView = (pathname = "") => {
+  // @ts-ignore
   const adobeDataLayer = window["adobeDataLayer"] || [];
```
7emansell, line 236: This one's really gnarly– tried `any`, etc, tried `interface AdobeDataLayerEvent {
  event?: string;
  page_name?: string;
  site_section?: string;
}`... open to ideas

app/components/swimlanes/swimLanes.tsx
```diff
@@ -18,9 +18,26 @@ import { imageURL } from "../../utils/utils";
 import { DC_URL } from "../../config/constants";
 import useBreakpoints from "../../hooks/useBreakpoints";
 
-const SwimLanes = ({ lanesWithNumItems }) => {
+export type CollectionItem = {
```
7emansell, line 21: This sort of duplicates/overlaps with Alessandra's `CollectionCard` work from #160, so needs to be melded together

app/components/search/search.tsx
```diff
@@ -11,7 +11,10 @@ const Search = () => {
   const [keywords, setKeywords] = useState("");
   const [publicDomainOnly, setPublicDomainOnly] = useState(false);
 
-  const handleSubmit = (event, router) => {
+  const handleSubmit = (
+    event: React.FormEvent<HTMLFormElement>,
+    router: ReturnType<typeof useRouter>
```
7emansell, line 16: first time I've done this with the `ReturnType` but it seems to work. I don't think this is a good thing to use a lot since it's kind of unclear, but there's no type for `AppRouterInstance`

app/api/feedback/route.tsx
```diff
@@ -13,7 +13,9 @@ export const POST = async (request: NextRequest) => {
     const origin = request.headers.get("origin") || "";
     const page = referer.replace(origin, "");
     const ipAddress = request.headers.get("x-forwarded-for") || "";
-    const userAgentParser = new UAParser(request.headers.get("user-agent"));
+    const userAgentParser = new UAParser(
+      request.headers.get("user-agent") ?? undefined
```
EdwinGuzman, line 17: Is this because `.get("user-agent")` can be an empty string? So you should make it undefined by default?

app/api/feedback/route.tsx
```diff
@@ -62,13 +64,15 @@ export const POST = async (request: NextRequest) => {
       },
       { status: 200 }
     );
-  } catch (e) {
-    return NextResponse.json(
-      {
+  } catch (e: unknown) {
+    if (e instanceof Error) {
+      return NextResponse.json({
```
EdwinGuzman, line 69: Can you do `NextResponse.status(400).json....`? I think that's the syntax. Otherwise, it's losing the status code in the response.

__tests__/data/swimlaneProps.tsx
```diff
@@ -1,5 +1,7 @@
-export const props = {
-  randomNumber: 1,
+import { SwimLanesProps } from "@/components/swimlanes/swimLanes";
+import type { CollectionItem } from "@/components/swimlanes/swimLanes";
```
EdwinGuzman, line 2: This is not used in the file

app/components/hero/campaignHeroSubText.tsx
```diff
@@ -8,7 +8,18 @@ import {
 } from "@nypl/design-system-react-components";
 import React from "react";
 
-const CampaignHeroSubText = ({ featuredItem }: any) => {
+interface CampaignHeroSubtextProps {
+  featuredItem: {
```
EdwinGuzman, line 12: I was going to say to use `FeaturedItemData` but that one contains `numberOfDigitizedItems`. This feels more correct and perhaps what's in `app/types/FeaturedItemData/FeaturedItemData.ts` should be renamed to be slightly more generic.

app/components/pageLayout/pageLayout.tsx
```diff
@@ -14,10 +14,11 @@ import { ADOBE_EMBED_URL } from "app/config/constants";
 import Script from "next/script";
 import { trackVirtualPageView } from "app/utils/utils";
 import { usePathname } from "next/navigation";
+import { BreadcrumbsDataProps } from "@nypl/design-system-react-components/dist/src/components/Breadcrumbs/Breadcrumbs";
```
EdwinGuzman, line 17: use the existing import statement.

app/components/search/search.tsx
EdwinGuzman: Yeah, you'd have to think about what's going on to understand it.

What you can also do is remove `router` as an argument. I don't think I noticed this before but it's not wrong to use the router from the global scope so you don't pass it to the function.

app/components/swimlanes/swimLanes.tsx
```diff
@@ -128,6 +145,7 @@ const SwimLanes = ({ lanesWithNumItems }) => {
       </Link>
     </Box>
   ));
+  return <>{lanes}</>;
```
EdwinGuzman, line 148: Any issue with React children?

app/utils/utils.ts
EdwinGuzman: honestly, I don't want to get too involved with Adobe and getting the type right for it. So this is fine, but can you add a comment above the `ts-ignore` and say something along the lines that we're just following Adobe's instructions and they don't provide ts support.

package.json
```diff
@@ -55,6 +55,7 @@
     "@types/jest": "29.5.5",
     "@types/jest-axe": "3.5.6",
     "@types/newrelic": "^9.14.3",
+    "@types/ua-parser-js": "0.7.39",
```
EdwinGuzman, line 58: Is this actively or when it's compiled?

EdwinGuzman (commented):
Looks good with some minor changes and questions. I would not worry about the comment in `campaignHeroSubText` regarding `FeaturedItemData` if you don't have time for it.

app/api/feedback/route.tsx
7emansell: I think it's that `.get("user-agent")` can be `null` (not `""`) and UAParser can only handle `undefined`

app/components/hero/campaignHeroSubText.tsx
7emansell: Maybe `HeroData`? I do find it a little confusing that we have `FeaturedItem` everywhere and then at some point you have to clock that that's actually the data going into the `CampaignHero` component, it'd be nice to have it be consistent everywhere. What do you think @avertrees

app/components/pageLayout/pageLayout.tsx
7emansell: Is there one? I did not even notice this import wasn't the standard one, but it looks like we don't export those props from the DS for some reason

app/components/swimlanes/swimLanes.tsx
7emansell: Yes it's mad that it's returning an array

package.json
7emansell: Just when it's compiled

app/components/pageLayout/pageLayout.tsx
EdwinGuzman: Ugh, I see why you did that. I'll work on that in the DS. So this is ok.

app/components/hero/campaignHeroSubText.tsx
EdwinGuzman: `HeroData` sounds good. We can work on that next week.

EdwinGuzman: Now that this has been updated with `main`, should the name be updated? @avertrees

app/components/swimlanes/swimLanes.tsx
```diff
@@ -20,10 +20,27 @@ import CollectionCard from "../cards/collectionCard";
 import type { CollectionCardDataType } from "../../types/CollectionCard";
 import { CollectionCardModel } from "../../models/collectionCard";
 
-const SwimLanes = ({ numColumns, lanesWithNumItems }) => {
-  const { isLargerThanLargeTablet } = useBreakpoints();
+// export type CollectionItem = {
```
EdwinGuzman, line 23: Delete if not needed anymore.

EdwinGuzman (approved):
I think this could use an update later to replace `any`s with proper types and fix type/interface names, but this is still a big step forward especially considering that the ts strict flag is set to true. So approving now and the two issues I mentioned can be fixed either in another round or while refactoring/developing new features -- this will become the standard now to add proper types in each PR.

avertrees:
made a new ticket to update the "any" types @EdwinGuzman

avertrees (approved):

PR #169, opened 2024-07-17
Feature/DR-3048/changelog endpoint

avertrees:
## Ticket:

- JIRA ticket [DR-3048](https://newyorkpubliclibrary.atlassian.net/browse/DR-3048)

- creates `/api/changelog` endpoint that returns the contents of the changelog.md file

## How has this been tested?

by hitting https://digital-collections-git-feature-dr-3048changelog-endpoint-nypl.vercel.app/api/changelog

-

### Checklist:

[DR-3048]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


7emansell (approved):

EdwinGuzman (commented):
This works but the output is difficult to read. I don't know how it renders in legacy DC but I think it should be readable for QA (or devs).

Screenshot 2024-07-18 at 11 03 15 AM

app/api/changelog/route.tsx
```diff
@@ -0,0 +1,14 @@
+import { NextResponse } from "next/server";
+import fs from "fs";
+import path from "path";
+
+export const GET = async () => {
+  const filePath = path.join(process.cwd(), "public", "CHANGELOG.md");
+  const fileContent = fs.readFileSync(filePath, "utf8");
+  // const mdToJSON = fileContent.toJSON();
+  const jsonMD = JSON.stringify(fileContent);
+  console.log("jsonMD is: ", jsonMD);
```
EdwinGuzman, line 10: Remove the console and the comment above if you don't need it.

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "next build && cp ./CHANGELOG.md ./public/changelog.md",
```
EdwinGuzman, line 7: This works because it runs on _build_ time. I wouldn't change how this script works, but this leads to my question, why do you need to commit `public/CHANGELOG.MD` to the repo?

I don't think it's necessary to commit this file to git.

`npm run builds` runs automatically in vercel and in docker when the app builds, so the changelog file will be copied over correctly. If you want to keep it committed in the repo, then you'd have to run `npm run build` before making PRs to keep both changelog files synced and typically you don't run `npm run build` locally.

If you do want/need it in git that's okay but just curious why it's there.

EdwinGuzman (approved):

package.json
avertrees: The file is there because, in the past (with ruby environments), I've ran into issues copying contents to files if the file didn't already exist. I kept it in as a precaution. 

Is the concern that the` /public/CHANGELOG.MD` file in the _repo_ won't reflect the changes that are live in vercel and aws because the changes created when running `npm run build` or `docker build` would only update the  file in the container?

EdwinGuzman: Yes, that's right. Both files can be out of sync in the repo and in vercel (or aws) if you _only_ update the main changelog file at the root of the repo. You and Emma (or anyone else, really) would have to remember to run `npm run build` before committing and pushing code so that they're synced. It's not common to have to run that command before pushing code.

Try this in vercel and aws without committing the `public/CHANGELOG.md` file and let me know what do you think.

EdwinGuzman (approved):

PR #170, opened 2024-07-19
DR-3089: Sticky header update

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3089](https://newyorkpubliclibrary.atlassian.net/browse/DR-3089).

On < 480px breakpoint: 
- Removes row stacking from search bar input and button
- Removes "Search" text from search bar button

On < 768px breakpoint: 
- Removes header and logo on downward scroll, adds them on upward scroll
- Removes scroll offset

## How has this been tested?

Locally

- Should take up less space on zoomed 1280 breakpoint or on mobile breakpoint = easier to read and navigate

### Checklist:

[DR-3089]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/search/search.tsx
```diff
@@ -65,11 +65,20 @@ const Search = () => {
             placeholder: "Search keyword(s)",
           }}
           sx={{
-            [`@media screen and (min-width: ${headerBreakpoints.lgMobile})`]: {
-              flexFlow: "column nowrap",
-            },
-            [`@media screen and (min-width: ${headerBreakpoints.smTablet})`]: {
-              flexFlow: "row nowrap",
+            flexFlow: "row nowrap",
```
7emansell, line 68: The below code is weird but I think necessary to get around the component styling, particularly the existence of the "Search" text

EdwinGuzman: Yeah, the `SearchBar` doesn't give you much room for this type of custom styling. It seems reasonable but why are textTransform and borderRadius needed? Two two don't do much, especially textTransform since the text is moved outside of the visible screen and is already capitalized.

The only update I'd ask is to update this to use `min-width` instead of `max-width` to be more consistent with the mobile-first approach.

EdwinGuzman (changes requested):
This works well but, if possible, switch to `min-width` for mobile-first approach. What you have should be the default and then when it hits the first breakpoint, you'd make the text visible for all bigger widths.

src/components/search/search.tsx
7emansell: `textTransform` was actually not needed– holdover from a different approach. `borderRadius` is necessary, needs to be explicitly set I think because the `textIndent` overrides the borders of the button.  I know we are aiming for a mobile-first approach but this is a situation where it's really counterintuitive, since the styling is meant to undo what's already there. Using `min-width` here adds 10-15 unnecessary lines of code, basically just trying to restore the search bar to its own default styling @EdwinGuzman

EdwinGuzman: Great, thanks for the explanation and trying it out!

EdwinGuzman (approved):
Just remove the `textTransform` if it's not needed.

7emansell:
Wishing I had bothered to simply google my question before going down this 30 commit rabbit hole but you live and learn... Not `generateMetadata` but `generateViewport`, and `next/headers` are more powerful than I realized https://github.com/vercel/next.js/discussions/60661 @EdwinGuzman

src/components/header/header.tsx
```diff
@@ -73,6 +76,8 @@ const Header = () => {
         </HStack>
         <HStack
           sx={{
+            //display: isScrolled ? "flex" : isScrollingUp ? "flex" : "none",
```
EdwinGuzman, line 79: Remove the commented out code .

src/components/pageLayout/pageLayout.tsx
```diff
@@ -6,7 +6,7 @@ import {
   SkipNavigation,
   useFeedbackBox,
 } from "@nypl/design-system-react-components";
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
```
EdwinGuzman, line 9: Don't think you need `useState` here.

src/hooks/useScrolled.ts
```diff
@@ -5,7 +5,7 @@ export function useScrolled(elementId: string) {
   const [isScrolled, setIsScrolled] = useState(true);
 
   useEffect(() => {
-    let DELAY_STICKY_TRANSITION = window.innerWidth < 600 ? 700 : 300;
+    let DELAY_STICKY_TRANSITION = window.innerWidth < 767 ? 0 : 300;
```
EdwinGuzman, line 8: Is it possible to use the values in `breakpoints.ts`? I know those are string values with the added "px" but just for consistency. This is not a blocker for this PR though.

EdwinGuzman (approved):
The solution provided in the thread is interesting and it's an abstraction from Next.js and the App Router. I'm glad you did the investigation to fix the issue because it allows you to get into the nitty gritty of this tech.

src/hooks/useScrolled.ts
7emansell: This inspired me to just make those breakpoint values numbers.. I *think* that makes more sense overall

avertrees (approved):

EdwinGuzman (approved):
Re-approving for the last update

PR #172, opened 2024-07-29
DR-3019: Move deployment to Github Actions

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3019](https://newyorkpubliclibrary.atlassian.net/browse/DR-3019)

- Creates deploy to QA and deploy to prod workflows
- Removes Travis config

## How has this been tested?

- Only way to test is by deploying so need to set up time to do that (esp prod)

### Checklist:

[DR-3019]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
-    "typescript.tsdk": "node_modules/typescript/lib",
-    "editor.formatOnSave": true,
```
EdwinGuzman, line 3: :(

7emansell: Wait why would you need my copy of this file I'm confused

.github/workflows/deploy_qa.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Log in to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Build and push Docker image to Amazon ECR
+        env:
+          APP_ENV: qa
+          NEW_RELIC_LICENSE_KEY: ${{ secrets.NEW_RELIC_LICENSE_KEY }}
+          ECR_URL: 685731035297.dkr.ecr.us-east-1.amazonaws.com/dc-frontend:qa-latest
```
EdwinGuzman, line 37: I usually see the final ECR url as 
```
ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
ECR_REPOSITORY: dc-frontend
```

and then 

```
docker push $ECR_REGISTRY/$ECR_REPOSITORY:qa-latest
```

I _think_ this should also work. You can leave this it is, but if there are any issues deploying, this is a place to start.

EdwinGuzman (commented):
Just minor comments but looks good!

The only way to test this is to deploy it to QA and see an update. What I'm doing for testing in the Remediation Team is to console log a message when the app loads that says, "App v2.0.0" or even "using github actions". You can do something similar to make it easier to know the switch to github moved, and then remove it later.

This might be something useful in general for dev and qa to check for testing, e.g. deploying a new prod update and checking that the log message renders the expected version. Not needed for this PR but the above can lead to this if it's useful.

.vscode/settings.json
EdwinGuzman: I don't think it's a bad idea to "force" others to make prettier work when they save. They might not have it enabled already. I also somehow lose it entirely so I can recreate it, nbd.

This could be an enhancement in another PR for the readme to tell about devs this useful file and settings.

.github/workflows/deploy_qa.yml
```diff
@@ -0,0 +1,45 @@
...
+          aws-region: us-east-1
+
+      - name: Log in to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Build and push Docker image to Amazon ECR
+        env:
+          APP_ENV: qa
+          NEW_RELIC_LICENSE_KEY: ${{ secrets.NEW_RELIC_LICENSE_KEY }}
```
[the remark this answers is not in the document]
7emansell: Yes just added

.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,46 @@
+# Deploys to nypl-dams-prod.
+name: deploy production
+
+on:
+  push:
```
EdwinGuzman, line 5: Just a note: once we know this is working, this should probably change to `pull_request`. It'll force you and Alessandra to make pull requests for deployments but you should both decide if that's a good workflow.

EdwinGuzman (approved):

avertrees (approved):

PR #173, opened 2024-07-31
remove query params from redirects of phase 3 pages

avertrees:
## Ticket:

- JIRA ticket https://newyorkpubliclibrary.atlassian.net/browse/DR-3028

- removes the query params from the redirect of the phase 3 pages, it's a bit redundant as it's another layer of conditional redirection on top of the middleware file.

## How has this been tested?

Locally by changing APP_ENV to 'qa'. Can change the APP_ENV back to `qa` in vercel to test if we need.

-

### Checklist:

### DIFF
--- a/app/collections/[slug]/page.tsx
+++ b/app/collections/[slug]/page.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { Metadata } from "next";
 import PageLayout from "../../../src/components/pageLayout/pageLayout";
 import CollectionsPage from "src/components/collections/collections";
-import { useSearchParams } from "next/navigation";
 
 type CollectionProps = {
   params: { slug: string };
@@ -17,7 +16,7 @@ export async function generateMetadata({
   };
 }
 
-export default function Lane({ params }: CollectionProps) {
+export default function Collections({ params }: CollectionProps) {
   return (
     <PageLayout
       activePage="lane"
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -50,11 +50,6 @@ export default function RootLayout({
     <html lang="en">
       <head>
         <meta httpEquiv="X-UA-Compatible" content="IE=edge,chrome=1" />
-        <script
-          async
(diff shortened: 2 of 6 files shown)


7emansell (changes requested):
I'm still seeing what Bridgette pointed out [here](https://newyorkpubliclibrary.atlassian.net/browse/DR-3028?focusedCommentId=114784) (ex., http://localhost:3000/collections/women-of-distinction-remarkable-in-works-and-invincible-in-character#/?tab=about redirects to http://localhost:3000/#/?tab=about, not http://localhost:3000/). 
I just tried a bunch of stuff, including directly setting the `window.location` and none of it worked to remove the hash, sending me down this path: https://stackoverflow.com/questions/72306039/nextjs-router-redirect-hash-routes-to-non-hash-routes . The `router.replace` you're using in `/search/results` is not what it's hitting, it's hitting the serverside middleware, so that's where the change needs to be made, but the hash is only available clientside so it's gnarly . I'll keep digging!!

avertrees:
@7emansell did you change the APP_ENV to 'qa' locally?

EdwinGuzman:
It's fine to move on without this for now. We'll chat more about this later.

EdwinGuzman:
@7emansell , we tried a couple of things and maybe the update is not the most ideal but it works and will eventually be removed when this page is released.
The point about `#/?...` not being removed was indeed gnarly. I think this is, or was, a ruby of rails routing convention for hashes. I know it was definitely something I struggled with in the Angular days.

Talking to Ali, _if_ the feature that created the `#/?...` hashes needs to be implemented in DCF, I will argue against keeping this convention since it's not commonly used. So at that point it won't be a one-to-one feature parity.

EdwinGuzman (approved):
Can you just add a comment in `homePage.tsx` above the router call to explain why it's there? And also a note about removing that code later in phase 3.

7emansell (approved):

PR #177, opened 2024-08-06
DR-3102: Scroll offset for header

7emansell:
## Ticket:

- JIRA ticket [DR-3102](https://newyorkpubliclibrary.atlassian.net/browse/DR-3102)

- Adds `scroll-padding-top` to offset the height of the sticky header 
- ** What I planned to do is use a hook to get the height of the header to set the offset, but the `scroll-padding-top` value should really be set on `<html>`, so then we're into the server side of the app (`layout.tsx`) and I'm operating with the mission of keeping as much of DCF server side as possible. The simplest solution then seemed to be just to statically set the height of the offset using media queries, which is less elegant and less specific, so there's not always a consistent amount of space between the header and the target element. I think as long as the target element is visible this is perfect accessibility-wise but it'll be up to VQA. TLDR it's CSS media queries with no JS at all

## How has this been tested?

Locally– try `/#mainContent` on the home page, or `/about#public_domain` on various breakpoints. 

### Checklist:

[DR-3102]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/layout.tsx
```diff
@@ -1,6 +1,7 @@
 import Script from "next/script";
 import React from "react";
 import { Metadata } from "next";
+import "./globals.css";
```
EdwinGuzman, line 4: This works well so feature-wise, approved.

Two things with this file though. Can it be scss instead of plain css? That way you might be able to use css vars from the DS for the min-width breakpoint values. If not possible because of the order the files load, no big deal.

My bigger concern are all default rules in the `global.css` files. They came from the default nextjs starter kit and they weren't used before so you can clean up and remove the default and only keep the new rules you added.

EdwinGuzman (changes requested):
I don't think you need to update the changelog for this since it's part of the larger header revamp ticket, unless if you want to highlight this update.

app/layout.tsx
7emansell: All unused defaults removed– I think they were getting overridden by the DS Provider so I didn't notice. Also, I'm not sure the values I'm using here are actually vars in the DS since these are the new breakpoints... am I missing a set of spacing variables somewhere

EdwinGuzman: Yes, they're not in the DS. Sorry, I meant from the internally defined breakpoints file.

EdwinGuzman: Actually, pulling in the js won't work and you could create CSS variables at the root level in `global.css`. Regardless, you'll have to update it whenever the new breakpoint values come out. I don't expect they'll come out until v4 of the DS which might be a while. So for now this is fine.

EdwinGuzman (approved):
Can you just add a TODO comment to update the breakpoint values to DS values in the future? Most likely for DS v4.

app/layout.tsx
7emansell: I'm a lot less familiar with Sass than standard CSS but it seems like I would still need to re-declare those breakpoint variables for Sass (like in a `_variables.scss` file or just at the top of `globals`), and even if this wasn't the case I'm not sure it feels worth it to add the whole package just for this use. Is that implementation what you were picturing ?

avertrees (approved):

PR #180, opened 2024-08-13
DR-3099: Render collections page (and move /src into /app)

7emansell:
## Ticket:

Resolves [DR-3099](https://newyorkpubliclibrary.atlassian.net/browse/DR-3099)

- Renders the collections page with dummy data
- Adds sort queries to URL from menu selection (both search and sort don't actually work yet)
- Moves /src into /app according to [Next standards](https://nextjs.org/docs/app/building-your-application/configuring/src-directory)**

** I am very sorry this wasn't a separate PR, I tried to have it be, and I am sorry in advance that I am asking you to review a 100-file PR 😬
The files that have actually changed their contents re. the collections page and specifically need review are:
1. [ `components/search/content.tsx`](https://github.com/NYPL/digital-collections/blob/DR-3099/feature/render-collections-page/app/src/components/search/content.tsx) 
2. [`components/pages/collectionsPage/collectionsPage.tsx`](https://github.com/NYPL/digital-collections/blob/DR-3099/feature/render-collections-page/app/src/components/pages/collectionsPage/collectionsPage.tsx).

### Checklist:

[DR-3099]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/search/content.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const SearchContent = () => {
+  const queryParams = useSearchParams();
+  const query = queryParams.toString();
+  const numColumns = useNumColumns();
+  const { isLargerThanLargeTablet } = useBreakpoints();
+
+  const router = useRouter();
+
+  const createQueryString = (name, value) => {
```
7emansell, line 24: This will need a much fuller implementation– just a starting point to show how we can add sort/filter queries alongside search queries

avertrees: oh, I was under the impression that this page would support a max of two query params: one for search and one for filter.

kind of how the url on the old site already behaves: https://digitalcollections.nypl.org/collections?utf8=%E2%9C%93&collection_keywords=cat&sort=title 

but we should definitely double check with Lana.

app/src/utils/utils.ts
```diff
@@ -216,6 +215,7 @@ import {
   ADOBE_ANALYTICS_DC_PREFIX,
   BASE_URL,
 } from "../config/constants";
+import { ENV_KEY } from "../types/EnvironmentType";
```
avertrees, line 218: any reason this got moved to line 218?

avertrees (approved):
looks good to me!

app/src/components/search/content.tsx
7emansell: The sort and filter is so broken on old DC that i think this is definitely a good opportunity to reevaluate

app/collections/[slug]/page.tsx
```diff
@@ -28,8 +27,6 @@ export default function Collections({ params }: CollectionProps) {
           url: `/collections/${params.slug}`,
         },
       ]}
-    >
-      <CollectionsPage slug={params.slug} />
```
Comment (EdwinGuzman) [inline app/collections/[slug]/page.tsx:32]: Do you still need this page/file?

app/src/components/search/content.tsx
EdwinGuzman: Wouldn't it need three? The search, the sort, and then the order?

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const numColumns = useNumColumns();
+  return (
+    <>
+      <Box
+        sx={{
+          display: "flex",
+          flexDirection: "column",
+          "> h1": {
+            marginBottom: 0,
+          },
```
EdwinGuzman, line 34: Why not add this style directly as a style prop to the `Heading` component?

EdwinGuzman:
This PR is already merge so no worries about addressing comments now but for a future PR.

app/collections/[slug]/page.tsx
Comment (7emansell) [reply app/collections/[slug]/page.tsx:32]: this is the collections landing page which is actually not part of phase 2, so this is just scaffolding.

app/src/components/search/content.tsx
7emansell: By order do you mean ascending/descending ? The implementation of that will have to change here because of how we're displaying the options– on old DC the order doesn't even appear in the URL, and it's pretty unclear what the up/down arrows even mean, but here the order is included in the menu options. Again, I didn't mean this to be the final implementation of the sort, I think that's something that will have to wait on the finished endpoint and more decisions about URL parity

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
7emansell: Sorry holdover from `Heading`s displaying as an `hgroup`. Will fix at next chance

app/src/components/search/content.tsx
EdwinGuzman: Ah yes, that's right. I keep thinking of legacy DC's query params. This would break parity but I think that's a good thing in this case.

PR #182, opened 2024-08-14
DR-3123: Adding containsOnSiteMaterials flag to models and card component

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3123](https://newyorkpubliclibrary.atlassian.net/browse/DR-3123)

- Adds `containsOnSiteMaterials` field to `Collection` and `CollectionCard` models, mocks
- Adds conditional status badge to `collectionCard` component

## How has this been tested?

Locally– `/divisions` or `/divisions/[slug]` pages

- Tooltips only on truncated titles now

### Checklist:

[DR-3123]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/cards/collectionCard.tsx
```diff
@@ -45,6 +46,12 @@ const CollectionCard = ({
         className={styles.cardTitle}
         noOfLines={3}
       >
+        {collection.containsOnSiteMaterials && (
```
EdwinGuzman, line 49: Add a test for this update.

app/src/components/cards/collectionCard.tsx
```diff
@@ -45,6 +46,12 @@ const CollectionCard = ({
...
+          <StatusBadge type="informative">
+            {" "}
```
EdwinGuzman, line 51: Why is the extra space needed?

avertrees:
I'm noticing some inconsistent padding between the thumbnail and the status badge - quick look at the dev tools looks likeit might be caused by the thumbnails
Screenshot 2024-08-15 at 9 08 30 AM

7emansell:
> I'm noticing some inconsistent spacing between the thumbnail and the status badge

that exists on all the cards on prod as well (even without badge, the spacing is inconsistent between title and thumbnail so that the card will match the height of the other cards in the row) so I think I'll wait for vqa to point it out before trying to resolve @avertrees

EdwinGuzman:
Looks like the longer titles move the buttom half of the Card content up, so the space is reduced between the image and the text. I initially thought it was because the title text was vertically aligned. If Lily does flag this, @7emansell can you then bring this up to Marty? The `Card` might need some stricter rules.

app/src/components/cards/collectionCard.tsx
```diff
@@ -7,10 +7,13 @@ import {
   CardContent,
   Tooltip,
   Link,
+  StatusBadge,
+  Heading,
 } from "@nypl/design-system-react-components";
 import styles from "./Card.module.css";
 import { headerBreakpoints } from "../../utils/breakpoints";
 import { CollectionCardDataType } from "../../types/CollectionCardDataType";
+import { TextDecoderStream } from "stream/web";
```
EdwinGuzman, line 16: Hmm be careful with this.

avertrees (approved):

7emansell:
Final tooltip implementation now in

__tests__/pages/aboutpage.test.tsx
```diff
@@ -8,7 +8,7 @@ jest.mock("next/navigation", () => ({
   usePathname: jest.fn(),
 }));
 
-describe("About page accessibility", () => {
+describe.skip("About page accessibility", () => {
```
EdwinGuzman, line 11: Why?

app/src/components/cards/collectionCard.tsx
```diff
@@ -6,56 +6,67 @@ import {
...
+              alt: "",
+              id: `no-image-${id}`,
+              isLazy: true,
+              aspectRatio: "twoByOne",
+              src: "/noImage.png",
+            }
+      }
     >
+      <CardContent>
+        {collection.containsOnSiteMaterials && (
```
EdwinGuzman, line 53: If it's not a layout problem, include `CardContent` in this conditional render logic. Small thing but one less div on the page.

app/src/components/cards/itemCard.tsx
```diff
@@ -9,9 +9,9 @@ import {
 import styles from "./Card.module.css";
 
 const ItemCard = ({ id, isLargerThanLargeTablet, item }) => {
-  return (
+  const truncatedTitle = item.title.length > 80; // Pretty much random
```
EdwinGuzman, line 12: Declare `80` as a constant variables to not have magic numbers, even if it's random.

EdwinGuzman (commented):
Minor updates.

The pattern of conditionally wrapping the `Card` in a `Tooltip` happens in two different files. They're very similar and I'm tempted to have a wrapper component that handles the logic but it might be an unnecessary abstraction:

```
// collectionCard.tsx
<CardWithTooltip
  renderTooltip={isLargerThanLargeTablet && truncatedTitle}
  offset={...}
  content={...}
>
  {card}
</CardWithTooltip>
```

This way, itemCard and collectionCard don't need to import Tooltip, but the tradeoff is importing this new component. Just a thought.

EdwinGuzman (approved):
Meant to approve while I'm out. The things to update are small.

app/src/components/cards/collectionCard.tsx
7emansell: Layout problem for the `Card` :(

7emansell:
> Minor updates.
> 
> The pattern of conditionally wrapping the `Card` in a `Tooltip` happens in two different files. They're very similar and I'm tempted to have a wrapper component that handles the logic but it might be an unnecessary abstraction:

Especially with the logic needed to make the `Tooltip` responsive, I think this is a good idea, but maybe this can be a refactor down the line since `ItemCard` is not reallyyy part of this phase anyways and this might end up being more complicated than it's worth.

avertrees (approved):
we reviewed most of the changes today, think this looks good. we can discuss refactoring the card component in with the work that will be done for https://newyorkpubliclibrary.atlassian.net/browse/DR-3125

PR #184, opened 2024-08-19
Feature/dr 3016/move api calls to server

avertrees:
Move API calls used by homepage to server and remove skeleton loader tests. Also updated the public changelog by accident after running npm build, but hey at least it's up-to-date in the repo now.

Note: The data is returned by the server before the CSS loads on the page, which results in the individual Collection Cards loading at the full width of the container while the page is "loading." Weird new behavior.


app/src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -5,22 +5,10 @@ import SwimLanes from "../swimlanes/swimLanes";
...
-  useEffect(() => {
-    const fetchData = async () => {
-      const response = await fetch("/api/homepage", {
-        method: "GET",
-        cache: "no-store",
-      });
-      const responseData = await response.json();
-      setData(responseData);
-    };
+const HomePageMainContent = (data) => {
```
EdwinGuzman, line 8: These are props and you need to destructure them, so this should be `({ data })`.

EdwinGuzman: The check on line 13 doesn't work because it's actually `data.data.lanesWithNumItems` in the way you have it set. So if you destructure the prop value, then the logic check will be correct.

app/src/components/pages/homePage/homePage.tsx
```diff
@@ -7,7 +7,10 @@ import HomePageMainContent from "../../homePageMainContent/homePageMainContent";
 import PageLayout from "../../pageLayout/pageLayout";
 import { useRouter } from "next/navigation";
 
-export default function HomePage() {
+export default function HomePage(data) {
```
EdwinGuzman, line 10: Same here, this should be `({ data })`

app/src/utils/api.ts
```diff
@@ -0,0 +1,43 @@
...
+      return {
+        ...collection,
+        numItems: uuidtoItemCountMap[collection.uuid] || "0",
+      };
+    });
+    return { ...lane, collections: updatedCollections };
+  });
+
+  const newResponse = { randomNumber, lanesWithNumItems: updatedLanes };
+  return JSON.parse(JSON.stringify(newResponse));
```
EdwinGuzman, line 29: Why do you need to stringify and then parse the result?

app/src/utils/api.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export const getHomePageData = async () => {
+  const data = await homePageData();
+  console.log("data in utils/api.ts is :", data);
+  if (data.lanesWithNumItems[0]) {
+    return data;
+  } else {
+    console.log("There is an error getting the home page data: ", data);
+    throw new Error(
```
EdwinGuzman, line 39: If you throw an error, the whole page will crash. Instead, you want to gracefully degrade and still show _something_. This is returning swimlane data which is static and should not fail. What would fail is fetching the item number count but that fetch function already handles the error state. So in this case this is redundant and not what we need. 

I don't think you need this entire function at all and you can export `homepageData` and just call that in `app/page.tsx`.

app/src/utils/utils.ts
```diff
@@ -5,6 +5,7 @@ import {
   BASE_URL,
 } from "../config/constants";
 import { ENV_KEY } from "../types/EnvironmentType";
+// import { useRouter } from "next/navigation";
```
EdwinGuzman, line 8: Just remove this.

EdwinGuzman (commented):
It's coming along well.
1. You need to follow a similar pattern for the featuredItem hero data.
2. The test is failing because it's trying to render and test the `Home` component which is now a server component calling the fetch function. Go one step down and instead test `HomePage`

app/src/components/homePageMainContent/homePageMainContent.tsx
avertrees: ah okay, that makes sense

app/src/utils/api.ts
avertrees: I was testing something in a panic, please ignore this

__tests__/api/featuredItem.tsx
```diff
@@ -1,28 +1,21 @@
-import CampaignHero from "../../app/src/components/hero/campaignHero";
+import CampaignHero from "../../app/src/components/featuredItem/campaignHero";
 import React from "react";
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { mockHomePageMainContent } from "__tests__/__mocks__/data/mockHomePageMainContent";
```
EdwinGuzman, line 4: You're not using this so remove it.

__tests__/api/featuredItem.tsx
EdwinGuzman, line 1: Rename this to `featuredItem.test.tsx`.

__tests__/pages/homepage.test.tsx
```diff
@@ -1,22 +1,17 @@
 import { render } from "@testing-library/react";
-import Home from "../../app/page";
 import { axe } from "jest-axe";
 import React from "react";
+import { mockHomePageMainContent } from "__tests__/__mocks__/data/mockHomePageMainContent";
```
EdwinGuzman, line 4: 👍

app/src/components/featuredItem/hero.test.tsx
EdwinGuzman, line 1: Rename this to `campaignHero.test.tsx` to match the component that is being tested.

__tests__/api/featuredItem.tsx
EdwinGuzman: Also, the location of this file seems odd. It's not an API endpoint test but a component test so it could probably move into the `app/src/components/featuredItem` directory. 

Just note that this file and `hero.test.tsx` are _both_ testing `CampaignHero` and the tests could be merged!

app/src/utils/api.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { LaneDataType } from "../../src/types/Lane";
+import { ENV_KEY } from "../../src/types/EnvironmentType";
+import {
+  getNumDigitizedItems,
+  getFeaturedImage,
+  imageURL,
+} from "../utils/utils";
+import appConfig from "../appConfig";
+
+export const getHomePageData = async () => {
```
EdwinGuzman, line 12: Add a ticket in the backlog to write tests for `getHomePageData` and `getFeaturedItemData`. Thanks!

public/CHANGELOG.md
```diff
@@ -7,9 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Updated
```
EdwinGuzman, line 10: Updating two changelogs can be a pain, so add a note in the readme and the github pull request template so it's not missed if this is the approach moving forward.

app/src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -5,22 +5,8 @@ import SwimLanes from "../swimlanes/swimLanes";
 import SwimLanesLoading from "../swimlanes/swimLanesLoading";
 import { useNumColumns } from "../../hooks/useNumColumns";
 
-const HomePageMainContent = () => {
+const HomePageMainContent = ({ data }) => {
   const numColumns = useNumColumns();
```
EdwinGuzman, line 9: In order to "fix" the issue where the swimlanes initially render as one column and then it renders as four columns, you can replace line 9 with (and import `useBreakpoints`).

```
const { isLargerThanLargeTablet, isLargerThanLargeMobile } = useBreakpoints();
  const numColumns = isLargerThanLargeTablet
    ? 4
    : isLargerThanLargeMobile
    ? 2
    : 4; 
```

The only problem is that it won't render as a single column on mobile which I think is an okay trade-off for the moment. You can create a ticket and assign it to me -- I want to see if we can set the columns more responsively through the DS. @7emansell if you want to take a look at this instead, then that's also ok,let me know!

EdwinGuzman (changes requested):
Thanks, the patterns in this PR are definitely moving the codebase in the right direction. The page loads quickly and it appears faster because there are no loading skeleton states.

There are some changes that should be made but none to the actual update. What's still missing but should be three separate tickets and PRs:

- adding tests for the new API util functions
- fixing the swimlane _desktop_ initial rendering of the swimlane columns
- I see the following error in the console
Screenshot 2024-08-22 at 1 08 21 PM

It has to do with `Homepage.tsx` specifically
```
const router = useRouter();
router?.replace("/", undefined);
```

I know this was added for the redirect but the code and error seems to slow down the page. When I remove it, it loads better. Can you create another ticket to update this? I don't know yet what to do with this but let's try to brainstorm ideas, the three of us together.

public/CHANGELOG.md
avertrees: I wouldn't say this is an approach I'd like to continue moving forward. I accidentally ran "npm run build" while working on this and updated the file. I figured it wouldn't hurt to include the updates but think it's overkill to update it with _every_ pr.

avertrees:
> Thanks, the patterns in this PR are definitely moving the codebase in the right direction. The page loads quickly and it appears faster because there are no loading skeleton states.
> 
> There are some changes that should be made but none to the actual update. What's still missing but should be three separate tickets and PRs:
> 
> * adding tests for the new API util functions
> * fixing the swimlane _desktop_ initial rendering of the swimlane columns
> * I see the following error in the console
> 
> Screenshot 2024-08-22 at 1 08 21 PM
> It has to do with `Homepage.tsx` specifically
> 
> ```
> const router = useRouter();
> router?.replace("/", undefined);
> ```
> 
> I know this was added for the redirect but the code and error seems to slow down the page. When I remove it, it loads better. Can you create another ticket to update this? I don't know yet what to do with this but let's try to brainstorm ideas, the three of us together.

@EdwinGuzman this redirect has to do with removing the query parameters from the url when we redirect the pages under development

__tests__/api/featuredItem.tsx
avertrees: I'm going to merge the tests from featuredItem.test.tsx into hero.test.tsx  and delete this file

app/src/utils/api.ts
avertrees: https://newyorkpubliclibrary.atlassian.net/browse/DR-3170

app/src/components/homePageMainContent/homePageMainContent.tsx
avertrees: is that an okay trade-off for the moment? I feel it might be best to keep the initial render as one column and create another ticket to fix that than change the way the entire page renders on mobile.

EdwinGuzman:
> IIRC this redirect has to do with removing the query parameters from the url when we redirect the pages under development

Yes, but now it seems it's causing issues and throwing an error, unless it was always there?

EdwinGuzman (commented):
Everything has been addressed. Let me know what you want to do with the swimlane column issue and then I'll approve (and wait for the other review).

app/src/components/homePageMainContent/homePageMainContent.tsx
7emansell: The desktop loading state has passed VQA on mobile (slipped past rather than cleared with flying colors maybe)– I think the bigger concern was with just indicating that there was content to see there, but ultimately I think we definitely want a responsive approach to this and it'll be more noticeable on phase 2 pages than it has been on the homepage. Lmk when you make the ticket @avertrees

7emansell (approved):
approved pending loading state decision (will the work be done in this PR or something else more specific)

avertrees:
> > IIRC this redirect has to do with removing the query parameters from the url when we redirect the pages under development
> 
> Yes, but now it seems it's causing issues and throwing an error, unless it was always there? I'm suggesting finding a different approach for this feature that doesn't throw an error.

do you think that's in the scope of this PR or should be addressed separately?

app/src/components/homePageMainContent/homePageMainContent.tsx
avertrees: Perhaps I am testing or implementing this wrong locally, but when I try the code above it does not **_only_** render the desktop (4 column) _loading_ state on mobile, but also renders the cards in the desktop (4 column) view after the styling and data have loaded on mobile. If this is true, I'm not sure this would pass VQA.

I pushed the change up. @7emansell @EdwinGuzman Let me know if I did something wrong or if you see what I'm seeing when testing the mobile views. https://digital-collections-git-feature-dr-3016move-api-cal-61ed34-nypl.vercel.app/

Screenshot 2024-08-23 at 9 22 29 AM

EdwinGuzman: I'm seeing the same. I knew it would happen and pointed it out as a potential issue that would not pass vqa, but I think it's an okay trade-off for now since desktop is more of the primary use-case. You can remove the code you added in the last commit and leave it so it renders one column and then renders four columns once the page loads. 

Regardless, the logic from the `useNumColumns` function needs to be updated so it works on mobile or desktop. And this should be addressed before the next release.

EdwinGuzman:
> do you think that's in the scope of this PR or should be addressed separately?

It should be addressed separately (along with 1. adding tests for the new API util functions and 2. fixing the swimlane desktop initial rendering of the swimlane columns).

EdwinGuzman (approved):

PR #189, opened 2024-08-29
DR-3101: Connect Divisions page to Repo API

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3101](https://newyorkpubliclibrary.atlassian.net/browse/DR-3101). 

- Connects divisions page to Repo API `/divisions` endpoint
- Removes "swim" lanes, now they are collection lanes (that can have a `isSwimLane` flag)

## How has this been tested?

Locally

### Checklist:

[DR-3101]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/cards/collectionCard.tsx
```diff
@@ -76,7 +76,9 @@ const CollectionCard = ({
             },
           }}
         >
-          {`${collection.numItems} items`}
+          {collection.numberOfDigitizedItems > 1
```
EdwinGuzman, line 79: Another common pattern is 
```
`${collection.numberOfDigitizedItems} item${collection.numberOfDigitizedItems > 1 ? "s" : "" }
```

They both give you the same result but if you need to pluralize other words in the future, you could make a function and make the number variable more generic.

app/src/components/collectionLanes/collectionLanes.tsx
```diff
@@ -8,28 +8,31 @@ import {
...
-import { SwimLanesProps } from "../../types/props/SwimLanesProps";
 
-const SwimLanes = ({ numColumns, lanesWithNumItems }: SwimLanesProps) => {
+const CollectionLanes = ({ numColumns, lanesWithNumItems, isSwimLane }) => {
   const { isLargerThanLargeTablet } = useBreakpoints();
-  const lanes = lanesWithNumItems.map((lane, key) => (
+  const seeMoreLink = (lane, isSwimLane) =>
+    isSwimLane
+      ? `${DC_URL}/collections/lane/${lane.slug}`
+      : `${DC_URL}/divisions/${lane.slug}`;
```
EdwinGuzman, line 23: I think this will work better if the "see more" link is passed as a prop instead of `isSwimLane`:

```
<CollectionLanes
  numColumns={numColumns}
  lanesWithNumItems={data?.data?.divisions}
  seeMoreLink={`${DC_URL}/divisions/${lane.slug}`}
/>
```

- `CollectionLanes` does not need to know what the link it should render is,
- the logic is outside of the component,
- it makes it more generic. If there's a new "see more" link in the future, you'd have to update the logic to render another URL. (This may not happen in this case in DC Facelift but consider this as a general pattern to simplify components).

EdwinGuzman (commented):
As we chatted offline, QA Repo /divisions isn't accessible right now and that's ok for now. Just reviewed what I could.

public/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Updated
 - Updated swim lanes and featured item features to use server side api calls (DR-3016)
+- Added `containsOnSiteMaterials` field to collection card model and component (DR-3123)
```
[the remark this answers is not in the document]
7emansell: its #182 and just line specific diff so it'll come into main before this

EdwinGuzman:
> Right now I have a literally a two liner relying on RepoApiCall for error handling (which may be enough?) it just seems too simple, maybe because we don't have a 500 page yet.

There is really no error handling going on. In `RepoApiCall`, if the response status is not 200 or if there is a 500 error from Repo API, the function returns `undefined`. It's catching the errors but not throwing any additional errors, which is good.

- You should log the errors (to cloudwatch since that function is a server function) in `RepoApiCall`.
- You need to decide what to do when it returns `undefined` and it's passed to `DivisionsPage`.  Right now, I just see a blank page but the heading still renders. 
- As a quick test in `DivisionsPage`, you can conditionally render `CollectionLanes` and if `data?.divisions` is undefined, then render a "sorry! no data, try again" (but better) message.

Now is probably the best time to work on the error handling and displaying no data to the user since the endpoint is still being worked on.

But... considering you said there's no 500 page, I think it's better to still render the page the user wanted to see and render a message if there's no data. The DCF 500 page should be returned when the app itself has an error, like a syntax error or runtime error.

7emansell:
@EdwinGuzman that makes sense, but then do we want to have each page handle its own error display (like how I have it here), or to have one error page that any failed data fetch would display like [this](https://nextjs.org/docs/app/api-reference/file-conventions/error)?

EdwinGuzman:
The error boundary is a very nice way to handle errors. I have not used it before so I can't recommend it, but it seems like you can (and maybe should) use it to have one error page for any failed data fetch. If you want to look into it, do it but it's out of scope for this PR.

7emansell:
Decided offline that this PR will not do error handling since we want a broader approach for all pages, will be done here https://newyorkpubliclibrary.atlassian.net/browse/DR-3175 cc @EdwinGuzman @avertrees

PR #193, opened 2024-09-05
DR-3098 connect /divisions/[slug] page to api, update page to match designs

avertrees:
## Ticket:

- JIRA ticket [FE | Connect division landing page to Repo API endpoint](https://newyorkpubliclibrary.atlassian.net/browse/DR-3098)

- Create `collectionsTable` component (render collection cards in grid format, plan to later rename to just be a table that supports refactored Card component that can support both Item and Collection cards)
- `collectionsTable` supports pagination and query parameters (using https://nextjs.org/learn/dashboard-app/adding-search-and-pagination as a reference)
- Pagination adds query parameter and #collection-table to the url so that the page reloads at the correct section
- Server components, ie `divisions/[slug]/page.tsx` get access to query parameters by using the "searchParams" prop, which is apparently a special name in Next.js https://nextjs.org/docs/app/api-reference/file-conventions/page
- cleanup utils file by moving Repo API related methods to `/utils/api.ts`
- update links to point to local urls if APP_ENV is development (easier for navigating the pages locally)
- create `/lanes` directory
- move `collectionLanes` to `/lanes` directory
- create `itemLane` component & directory

## Things to create:
- Table component ticket (for rendering Cards)
- Lane component ticket
- Card component ticket 

## Open Questions / Concerns / Things I noticed

- Should probably create custom props for each component instead of just passing down { data }
- Some thumbnails don’t have the correct rights and we should probably load the broken thumbnail image as the fallback image in this case, similar to how we load the fallback image in the featuredItem
- Schomburg Center for Research in Black Culture, Moving Image and Recorded Sound Division only has one collection in the /divisions/slug endpoint and is not rendering anything at all on the frontend
- Having pagination at the bottom of the page is lowkey annoying? Solution could be to render 
- I had issues passing numColumns and isLargerThanBreakpoint down to child components but I think this was an issue with how I was implementing this
- The Division `title` can come from either the slug or the Repo API data (it's 'name' in the API response), idk if it's better to use the slug or the data?
-  if collections have the same number of items, they are not currently sorted alphabetically by title (data issue with the endpoint).

-

### Checklist:


EdwinGuzman:
I did not review all of the code yet but general comments we talked offline about:

- This link should not span the entire page's width: 
Screenshot 2024-09-09 at 11 31 22 AM
- Fix the `data.collections.map` issue that happens on bad data. You mentioned the endpoint was stale but this is should also be addressed on the frontend.
Screenshot 2024-09-09 at 11 29 56 AM
- Rename `CollectionsTable` to `CollectionsGrid`.
- Use a ref to focus on the heading after the page is updated. This ensures focus is visibly updated and the user can navigate with a keyboard once the data is updated.
- I'll work on a quick DS release candidate to fix the fallback image props for the `Card` component.
- `CollectionsGrid` can be reusable but that can be a focus later on.
- Likewise, the logic in component can be moved to a reusable hook.

app/src/components/cards/collectionCard.tsx
```diff
@@ -64,6 +64,12 @@ const CollectionCard = ({
               id: `image-${slug}-${id}`,
               isLazy: true,
               aspectRatio: "twoByOne",
+              fallbackSrc: "/noImage.png",
+              onError: (_event) =>
+                console.log(
+                  "Card image failed to load, fallback image loaded instead. ImageURL: ",
```
EdwinGuzman, line 70: Nit: use string interpolation.

app/src/components/cards/itemCard.tsx
```diff
@@ -40,6 +41,12 @@ const ItemCard = ({ id, isLargerThanLargeTablet, item }) => {
         id: `image-${id}`,
         isLazy: true,
         aspectRatio: "twoByOne",
+        fallbackSrc: "/noImage.png",
+        onError: (_event) =>
+          console.log(
```
EdwinGuzman, line 46: Use `console.warn` so it's more prominent in the console. (also but nitpick: use string interpolation).

app/src/components/lanes/itemLane/itemLane.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Box>
+        <Flex alignItems="baseline">
+          <Heading level="h2" size="heading3">
+            {`Items in the ${divisionName}`}
+          </Heading>
+          <Spacer />
+          <Link
+            id={`row-see-more-items-${divisionName}`}
+            type="standalone"
+            href={`${DC_URL}/search/index?filters[divisionFullname_mtxt_s][]=${titleToDCParam(
```
EdwinGuzman, line 33: That's an interesting looking URL. Out of curiosity and not blockers: would you happen to know why it's `+` instead of the browser' `%20` for spaces? Or why an URL encoded title can't be passed?

app/src/components/pages/divisionPage/divisionPage.tsx
```diff
@@ -1,37 +1,28 @@
...
+import React from "react";
+import { ItemLane } from "../../lanes/itemLane/itemLane";
+import { CollectionsGrid } from "../../tables/collectionsGrid";
+import { useState } from "react";
 
-export default function DivisionPage() {
+export default function DivisionPage({ data }: any) {
   const params = useParams();
-  const { isLargerThanLargeTablet } = useBreakpoints();
+  const queryParams = useSearchParams();
```
EdwinGuzman, line 19: `queryParams` is not being used.

app/src/components/pages/divisionPage/divisionPage.tsx
```diff
@@ -61,101 +52,22 @@ export default function DivisionPage() {
...
-              collection={collectionModel}
-              isLargerThanLargeTablet={isLargerThanLargeTablet}
-            />
-          );
-        })}
-      </SimpleGrid>
+      <ItemLane data={data} />
+
+      <HorizontalRule sx={{ marginTop: "xxl", marginBottom: "xxl" }} />
+      <CollectionsGrid data={data} updateCollections={updateCollections} />
```
EdwinGuzman, line 70: The `updateCollections` function is not being used and the prop is not declared in `CollectionsGrid`. Remove it.

app/src/components/pages/divisionsPage/divisionsPage.tsx
```diff
@@ -23,29 +24,31 @@ export default function DivisionsPage() {
...
-      <HorizontalRule sx={{ marginTop: "xxl", marginBottom: "xxl" }} />
-      <SwimLanes
-        numColumns={numColumns}
-        lanesWithNumItems={mockSwimLanes.lanesWithNumItems}
-      />
+      {data?.divisions && data.divisions.length > 0 ? (
+        <>
+          <Box
+            sx={{
+              maxWidth: "715px",
```
EdwinGuzman, line 31: Is this kind of arbitary?

app/src/utils/utils.ts
```diff
@@ -4,8 +4,7 @@ import {
 } from "../config/constants";
 import { ENV_KEY } from "../types/EnvironmentType";
 import appConfig from "../../../appConfig";
-import defaultFeaturedItems from "../data/defaultFeaturedItemData";
-
+import { CARDS_PER_PAGE } from "../config/constants";
```
EdwinGuzman, line 7: This is not being used

app/divisions/page.tsx
```diff
@@ -1,11 +1,20 @@
...
 };
 
-export default function Divisions() {
-  return <DivisionsPage />;
+export default async function Divisions() {
+  let data;
+  try {
+    data = await RepoAPICall(`${process.env.API_URL}/api/v2/divisions`);
+  } catch (error) {
+    logger.error("Failed to fetch divisions data", error);
```
EdwinGuzman, line 16: NBD _right now_ but I'm pretty sure it will log "undefined". So eventually the proper error should be thrown in `RepoAPICall`.

app/src/components/tables/collectionsGrid.tsx
EdwinGuzman, line 1: Don't think you need `tables` in the directory anymore.

app/src/models/collectionCard.ts
```diff
@@ -1,23 +1,29 @@
...
     this.uuid = data.uuid;
     this.title = data.title;
-    this.url = data.url;
-    this.imageID = data.image_id;
-    this.imageURL = imageURL(data.image_id, "full", "288,", "0");
-    this.numItems = data.numItems || 0;
-    this.containsOnSiteMaterials = data.containsOnSiteMaterials || false;
+    this.url =
+      process.env.APP_ENV === "development"
+        ? `/collections/${stringToSlug(data.title)}#/?tab=about`
```
EdwinGuzman, line 20: The `#/?tab=about` is temporary?...

avertrees: it's the default in the old app...

app/src/components/pages/divisionsPage/divisionsPage.tsx
7emansell: Arbitrary but on the design

app/src/components/lanes/itemLane/itemLane.tsx
avertrees: given that the filter is `divisionFullname_mtxt_s`, I'm assuming it's a solr thing. perhaps solr use use both `AND` and `+`, but I can ask Kris if we need to know. I'm just matching what I see in the legacy dc lol

app/src/components/pages/divisionPage/divisionPage.tsx
avertrees: more scalpels to remove from the patient, sorry about that.

app/src/components/tables/collectionsGrid.tsx
avertrees: should it be `grids`?

app/src/models/collectionCard.ts
avertrees: https://digitalcollections.nypl.org/collections/16ad5350-c52e-012f-aecf-58d385a7bc34 works, but legacy DC always appends the `#/?tab=about`

7emansell (approved):

app/src/components/pages/divisionPage/divisionPage.tsx
```diff
@@ -35,6 +24,8 @@ export default function DivisionPage() {
     setIsLoaded(true);
   }, []);
 
+  const updateCollections = () => {};
```
EdwinGuzman, line 27: This is not being used, remove it.

app/src/components/pages/divisionPage/divisionPage.tsx
```diff
@@ -64,102 +55,32 @@ export default function DivisionPage() {
...
+          <span> Contact info and more </span>
         </Link>
       </Box>
+
+      <HorizontalRule sx={{ marginTop: "xxl", marginBottom: "xxl" }} />
+
+      {isLoaded ? (
+        <ItemLane data={data} />
+      ) : (
+        <>
```
EdwinGuzman, line 74: You don't need the fragment for a single child component.

app/src/models/collectionCard.ts
EdwinGuzman: We talked about this offline but it should be kept to link to legacy appropriately, but once the `/collection/:slug` page is worked on in DCF, it should be removed. We also need to take into account whether there will be a tabs component in the new design that has an "About" tab.

app/src/utils/api.ts
```diff
@@ -188,9 +188,12 @@ export const RepoAPICall = async (
       const data = await response.json();
       return data;
     } else {
+      console.log(`Response from Repo API ${apiUrl} was not a 200`);
       return undefined;
     }
   } catch (error) {
+    console.log(`error fetching Repo API ${apiUrl}`);
```
EdwinGuzman, line 195: You can remove the logs. I think most of this is updated in the error handling PR.

EdwinGuzman (changes requested):
Just some minor updates. Bigger updates I'm flagging in DR-3194.

avertrees:
@EdwinGuzman do you mind if the feedback requested in this PR gets addressed in the PR for https://newyorkpubliclibrary.atlassian.net/browse/DR-3194

PR #194, opened 2024-09-09
DR-3171: Lane skeleton loading

7emansell:
## Ticket:

Resolves [DR-3171](https://newyorkpubliclibrary.atlassian.net/browse/DR-3171)

- Adds lane loading to `/divisions`, `/division/[slug]` and establishes pattern for manual loading state
- Removes `useNumColumns` hook in favor of media queries in a custom `DCSimpleGrid` component

Locally: `/divisions` since this is branched off code really hitting Repo API, or, on Vercel: `/divisions/hello`

### Checklist:

[DR-3171]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman (commented):
This works nicely and resolves the issue with the initial big card displaying before the styling kicks in, but I think the underlying issue should be resolved. I think you should try to implement a local custom `SimpleGrid` that is more flexible and can adjust responsively to the number of columns it should render.

I agree the loading doesn't have to be perfect, but it is jarring when three skeleton cards display initially in a row and then four appear when js kicks in. If the above suggestion doesn't work out, at a minimum display four skeleton cards. The `useEffect` pattern makes sense but, this might be a nuisance, perhaps it should go higher in the DOM tree.

7emansell:
> This works nicely and resolves the issue with the initial big card displaying before the styling kicks in, but I think the underlying issue should be resolved. I think you should try to implement a local custom `SimpleGrid` that is more flexible and can adjust responsively to the number of columns it should render.
> 
> I agree the loading doesn't have to be perfect, but it is jarring when three skeleton cards display initially in a row and then four appear when js kicks in. If the above suggestion doesn't work out, at a minimum display four skeleton cards. The `useEffect` pattern makes sense but, this might be a nuisance, perhaps it should go higher in the DOM tree.

A custom `SimpleGrid` is a good idea, I'll do that. Looks straightforward in the DS. Just another reason to be excited for the DS to update to new breakpoints :) Also, `useEffect` is clientside, so I think it's actually at the highest possible branch of the DOM tree.. @EdwinGuzman

EdwinGuzman:
> Also, useEffect is clientside, so I think it's actually at the highest possible branch of the DOM tree..

Good point. I'm not entirely used to the rsc and client component delineation yet.

7emansell:
Off the /divisions endpoint which is not going to QA

PR #197, opened 2024-09-11
DR-3175: More error handling and server error page

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3175](https://newyorkpubliclibrary.atlassian.net/browse/DR-3175).

- Adds/refactors more error handling in `utils/api.ts`
   - Instead of returning undefined for bad requests, now helpers throw an error
- Creates error boundary and clientside page (`error.tsx`) on `/app` level for server errors
   - Very boilerplate design [in anticipation of DRB designs](https://newyorkpubliclibrary.atlassian.net/browse/DR-3203)
- 404s are now caught on the page level, which is a pattern that needs to be consistent throughout the app as we hook more pages up to Repo API

Try setting incorrect Repo API credentials in your `.env`, or just `throw new Error("fake error")` on any of the server components. Note that the message on the error page reflects the type of error

### Checklist:

[DR-3175]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/divisions/[slug]/page.tsx
```diff
@@ -21,6 +22,9 @@ export async function generateMetadata({
 
 export default async function Division({ params, searchParams }) {
   const data = await getDivisionData(params.slug, searchParams.page);
+  if (data?.headers?.code === "404") {
```
Comment (EdwinGuzman) [inline app/divisions/[slug]/page.tsx:25]: We talked about this but to be clear, this is because Repo API's 200 response returns the 404 _within_ the data content, correct? Can you add this info as a comment above this? It's not exactly a standard API response so it's worth mentioning it here.

app/src/utils/utils.test.tsx
```diff
@@ -50,25 +48,24 @@ describe("apiPOSTCall()", () => {
...
         )
-      ).toBe("success");
+      ).toStrictEqual({ nyplAPI: { response: "success" } });
     });
   });
 });
 
 describe("getItemsCountFromUUIDs()", () => {
-  it("should return an empty object with a bad request", async () => {
+  it("should return an error with a bad request", async () => {
```
EdwinGuzman, line 57: nit: "throw" an error rather than "return"

app/src/utils/api.ts
```diff
@@ -107,13 +108,11 @@ export const getNumDigitizedItems = async () => {
 export const getItemsCountFromUUIDs = async (uuids: string[]) => {
   const apiUrl = `${process.env.API_URL}/api/v2/items/counts`;
   const response = await apiPOSTCall(apiUrl, { uuids });
-
-  if (!response?.counts?.count?.length) {
+  const counts = response.nyplAPI.response.counts;
```
EdwinGuzman, line 111: nit: `const { counts } = response.nyplAPI.response;`

app/src/utils/api.ts
```diff
@@ -188,13 +187,13 @@ export const RepoAPICall = async (
       const data = await response.json();
       return data;
     } else {
-      console.log(`Response from Repo API ${apiUrl} was not a 200`);
-      return undefined;
+      throw new Error(
+        `${response.status}: ${response.statusText || "3xx/4xx error"}`
```
EdwinGuzman, line 191: Can you prefix the messages with the name of the function? So in this case, "RepoAPICall: ....". I see the errors in the terminal console and the stack trace is helpful but this might make it easier to read both in the console and in cloudwatch if these are being logged there (since the stack trace is not).

EdwinGuzman: There are four throws that should have the prefix of their function name.

app/src/utils/api.ts
```diff
@@ -216,21 +215,28 @@ export const apiPOSTCall = async (apiUrl: string, postData: any) => {
...
   } catch (error) {
-    return undefined;
+    console.error(error);
+    throw new Error(error.message);
   }
 };
 
 export const getDivisionData = async (
-  slug: string,
+  slug?: string,
```
EdwinGuzman, line 231: Switch this around so that the optional argument is last in the list - general optional argument convention. But also, since you're giving them all defaults, they are all optional. This is called in `app/divisions/page.tsx` without any arguments.

Another pattern can also be to have these arguments passed as an object, such as:

```
const data = await getDivisionData({ slug: params.slug, pageNum: searchParams.page });
```

You'd still need the defaults and then the consumer can put the args in any order.

EdwinGuzman (changes requested):
Minor updates. This nicely catches the 401 failure. I added a wrong `AUTH_TOKEN` and got 
Screenshot 2024-09-19 at 12 20 38 PM

This could be better:
Screenshot 2024-09-19 at 12 20 54 PM

I updated `API_URL` to `https://ap.repo.nypl.org`. I think a better error message here is a 500. So there might be a fetch call that is not being caught or updated.

7emansell:
Discussed offline but recording here: new designs for the error pages (the catch all `error.js` and the 404) are coming, and won't display the specific error message at all, but I'm still passing the error object to `error.js` to be logged.

EdwinGuzman (commented):
Update looks good but do you need official designs before merging in since it might have different styles?

EdwinGuzman (approved):

avertrees (approved):

PR #200, opened 2024-09-20
DR-3101 (continued): VQA on Divisions page

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3101](https://newyorkpubliclibrary.atlassian.net/browse/DR-3101)

- Adjusts lane title widths and link behavior to better handle long titles
- Also updates DS to `v3.3.2`

### Checklist:

[DR-3101]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/collectionLanes/collectionLanes.tsx
+++ b/app/src/components/collectionLanes/collectionLanes.tsx
@@ -56,7 +56,6 @@ const CollectionLanes = ({ lanesWithNumItems, seeMoreLink }) => {
               },
             color: "ui.link.primary",
             fontWeight: "medium",
-            alignItems: "end",
             alignSelf: "end",
             _hover: {
               textDecoration: "underline 1px dotted !important",
--- a/app/src/components/collectionLanes/collectionLanes.tsx
+++ b/app/src/components/collectionLanes/collectionLanes.tsx
@@ -42,6 +42,7 @@ const CollectionLanes = ({ lanesWithNumItems, seeMoreLink }) => {
           hasVisitedState
           isUnderlined={false}
           __css={{
+            paddingLeft: "l",
             whiteSpace: "nowrap",
             justifyContent: "flex-end",
             display: "none",
--- a/app/src/components/collectionLanes/collectionLanes.tsx
+++ b/app/src/components/collectionLanes/collectionLanes.tsx
@@ -61,6 +61,7 @@ const CollectionLanes = ({ lanesWithNumItems, seeMoreLink }) => {
             _hover: {
               textDecoration: "underline 1px dotted !important",
               textUnderlineOffset: "2px",
+              "> svg": { fill: "ui.link.secondary" },
             },
             _visited: { "> svg": { fill: "ui.link.tertiary" } },
           }}
--- a/app/src/components/collectionLanes/collectionLanes.tsx
+++ b/app/src/components/collectionLanes/collectionLanes.tsx
@@ -20,7 +20,7 @@ const CollectionLanes = ({ lanesWithNumItems, seeMoreLink }) => {
   const { isLargerThanLargeTablet } = useBreakpoints();
   const lanes = lanesWithNumItems?.map((lane, key) => (
     <Box className={styles.lane} data-testid={lane.slug} mt="xxl" key={key}>
-      <Flex alignItems={"center"}>
+      <Flex alignItems="baseline">
         <Heading
           id={`row-heading-${lane.slug}`}
           sx={{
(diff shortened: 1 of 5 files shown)


app/src/components/pages/divisionsPage/divisionsPage.tsx
```diff
@@ -46,9 +46,9 @@ export default function DivisionsPage({ data }) {
             />
           ) : (
             <>
-              <CollectionLanesLoading />,
-              <CollectionLanesLoading />,
-              <CollectionLanesLoading />
+              {[...Array(36)].map((_, index) => (
```
EdwinGuzman, line 49: Where did you get "36" from?

7emansell: Randomish: needed to be more than 24, the number of divisions, but also needed to somewhat account for the space taken up by loading the hero and the footer

avertrees: yeah good question - looking at the sheet lana provided https://docs.google.com/spreadsheets/d/1oDILU2vJBKqTcv0nINuDkfwmp9hIyZc_0jzHcadSGrY/edit?gid=0#gid=0 there are less than 36 divisions to display

app/src/components/pages/divisionsPage/divisionsPage.tsx
```diff
@@ -12,7 +12,7 @@ import { DC_URL } from "@/src/config/constants";
 import CollectionLanesLoading from "../../collectionLanes/collectionLanesLoading";
 
 export default function DivisionsPage({ data }) {
-  console.log("clientside divisions", data);
+  console.log("clientside divisions", data.divisions.length);
```
EdwinGuzman, line 15: I can't run this locally right now but I'm getting this run time error:
Screenshot 2024-09-23 at 10 34 10 AM

This will be good to catch and return a default `{ divisions: [] }` from the server component.

app/src/components/pages/divisionsPage/divisionsPage.tsx
EdwinGuzman: I think you can get away with a much smaller number to display the mocked skeleton loaders. Perhaps just 6 or so? Not sure the user will scroll down quickly enough to see the rest of the loaders.

app/src/components/collectionLanes/collectionLanes.tsx
```diff
@@ -6,38 +6,67 @@ import {
...
+          }}
+          level="h2"
+          size="heading3"
+        >
           {lane.name}
         </Heading>
         <Spacer />
         <Link
           id={`row-see-more-${lane.slug}`}
-          type="standalone"
```
EdwinGuzman, line 28: Why doesn't this work anymore?

app/src/components/collectionLanes/collectionLanes.tsx
```diff
@@ -6,38 +6,67 @@ import {
...
+            " > svg": {
+              fill: "ui.link.primary",
+            },
+            [`@media screen and (min-width: ${headerBreakpoints.smTablet}px)`]:
+              {
+                display: "inline",
+              },
+            color: "ui.link.primary",
+            fontWeight: "medium",
+            alignItems: "end",
```
EdwinGuzman, line 59: `align-self` is the right rule. This one works if the container has `display: "flex"` for its children, so you can remove this.

EdwinGuzman: I see the other "see more" link has "flex" so maybe this carried over from that one?

EdwinGuzman (changes requested):
Can you include that the DS was updated in the changelog in this ticket?

app/src/components/pages/divisionsPage/divisionsPage.tsx
7emansell: If the user is anywhere near the footer and reloads, I assumed we want to return them to where they were

app/src/components/collectionLanes/collectionLanes.tsx
7emansell: I wanted to explicitly prevent the arrow wrapping onto another line and the necessary CSS made it easier to just recreate from scratch

EdwinGuzman (approved):
Can you create a ticket for the `standalone` type issue? It should be fixed in the DS so you _do_ use the `standalone` type instead of custom css.

7emansell:
Yes but I don't think what I/Lily wanted here is expected behavior from the DS component? @EdwinGuzman

PR #201, opened 2024-09-20
DR-3192: Card refactor

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3192](https://newyorkpubliclibrary.atlassian.net/browse/DR-3192)

- Condenses `ItemCard` and `CollectionCard` into `DCCard`
   - Adds `ItemCardDataType` for the typeguard
- Pulls out `useTooltipOffset` hook and sets `cardRef` and `cardOffset` on the lane/grid level, allowing the hook to run a lot less (this hook could serve a dual purpose with the Next Image implementation down the road)
- Adds more tests for the `DCCard` with both items and collections

Everything should look the same!

### Checklist:


app/src/hooks/useTooltipOffset.ts
```diff
@@ -0,0 +1,24 @@
+import { useState, useEffect } from "react";
+
+export function useTooltipOffset(cardRef) {
+  const [offset, setOffset] = useState<[number, number]>([0, -130]);
+  const getOffset = () => {
```
EdwinGuzman, line 5: It's not exactly clear from this function and `useTooltipOffset` names, but the offset is a coordinate and that might be useful to know for whoever calls this function.

app/src/hooks/useTooltipOffset.ts
```diff
@@ -0,0 +1,24 @@
+import { useState, useEffect } from "react";
+
+export function useTooltipOffset(cardRef) {
```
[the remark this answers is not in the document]
EdwinGuzman: I know this is only going to be used within this app and you both know about the code, but my comment is more for maintainability purposes.

app/src/components/dcCard/DCCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+              alt: "",
+              id: `no-image-${id}`,
+              isLazy: true,
+              aspectRatio: "twoByOne",
+              src: "/noImage.png",
+            }
+      }
+    >
+      <CardContent>
+        {isCollection && record.containsOnSiteMaterials && (
```
EdwinGuzman, line 70: Based on the interfaces declared, the `CollectionCardDataType` interface is the only type that has `containsOnSiteMaterials`. Can you include this check in the `isCollectionCardDataType` type guard? In other words, is it true that a collection contains both `containsOnSiteMaterials` and `numberOfDigitizedItems`? Or is there another possible type that contains `numberOfDigitizedItems` but not `containsOnSiteMaterials` and is not an `ItemCardDataType`?

app/src/components/dcCard/DCCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+            {`${Math.floor(record.numberOfDigitizedItems)} item${
+              Math.floor(record.numberOfDigitizedItems) !== 1 ? "s" : ""
+            }`}
+          </Text>
+        )}
+      </CardContent>
+    </Card>
+  );
+  return isLargerThanLargeTablet && truncatedTitle ? (
+    <Tooltip offset={cardOffset as [number, number]} content={record.title}>
```
EdwinGuzman, line 109: The `[number, number]` type occurs in the hook as well. I think you can declare this as "Coordinates" or "Offset" type.

EdwinGuzman: If that type is imported and used in the component prop type declaration`cardOffset?: number[];`, then the "as" should not be needed anymore (which is generally preferable ts).

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
```diff
@@ -10,17 +10,17 @@ import PageLayout from "../../pageLayout/pageLayout";
 import { useParams } from "next/navigation";
 import { headerBreakpoints } from "../../../utils/breakpoints";
 import { slugToString } from "../../../utils/utils";
-import CollectionCard from "../../../components/cards/collectionCard";
```
EdwinGuzman, line 13: There are other imports you can remove since they are not being used:
`SimpleGrid`, `PageLayout`, `DCCard`, `CollectionCardModel`, `useBreakpoints`, `CollectionDataType`, and `DCSimpleGrid`. Perhaps slightly out of scope of this PR but pointing it out since `DCCard` and `CollectionsGrid` are imported.

app/src/components/dcCard/DCCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+function isCollectionCardDataType(
+  record: CollectionCardDataType | ItemCardDataType
+): record is CollectionCardDataType {
+  return "numberOfDigitizedItems" in record;
+}
+
+const DCCard = ({
+  cardOffset,
+  cardRef,
```
EdwinGuzman, line 33: Can you rename this as `ref` or use React's `forwardRef` function https://react.dev/reference/react/forwardRef?

app/src/types/ItemCardDataType.ts
```diff
@@ -0,0 +1,9 @@
+export interface ItemCardDataType {
```
EdwinGuzman, line 1: This is not entirely relevant to this PR but I found `CollectionCardDataType` and `CollectionDataType` confusing. They're very similar in name and properties. Likewise, `ItemCardDataType` and `CollectionCardDataType` share almost all properties. I think this one, `ItemCardDataType`, can be used as a base that `CollectionCardDataType` extends from with its own two additional properties.

app/src/hooks/useTooltipOffset.ts
```diff
@@ -0,0 +1,24 @@
...
+  const [offset, setOffset] = useState<[number, number]>([0, -130]);
+  const getOffset = () => {
+    if (cardRef.current) {
+      const image = cardRef.current.children[0] as HTMLElement;
```
EdwinGuzman, line 7: I'm glad you were able to extract the hook and got it to work as we discussed. I did not consider having to pass a ref to every card component but then only reading the first one in the children array. If this is the case, I think you can get away with passing the ref only to the first `DCCard` component in the map function and not pass it to the rest. I don't think it's a big issue rendering- or computation-wise but minor optimization.

EdwinGuzman (changes requested):
Nice refactor! Just some updates with types and the tooltip offset hook.

app/src/components/dcCard/DCCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function isCollectionCardDataType(
+  record: CollectionCardDataType | ItemCardDataType
+): record is CollectionCardDataType {
+  return "numberOfDigitizedItems" in record;
+}
+
+const DCCard = ({
+  cardOffset,
```
avertrees, line 32: change to toolTipOffset for the time being plz

app/src/components/dcCard/DCCard.tsx
7emansell: Even with a declared type, it's the DS Tooltip that doesn't like it (that's on me) so I'll still have to use `as`

app/src/components/dcCard/DCCard.tsx
7emansell: My thinking here was that ultimately items, as they're represented in DCF, will have some kind of `isOnsite` field. I doubt it would be the same name but the idea is similar enough that I didn't want to include it in the type guard. `numberOfDigitizedItems` is clearly never going to be a field on an item, so I think that's the only safe bet so far– not to say this type guard couldn't be built up once we have a definitive `Item` type (as you mention below)

app/src/hooks/useTooltipOffset.ts
7emansell: I agree and also I think that change should happen in this ticket, not here https://newyorkpubliclibrary.atlassian.net/browse/DR-3191

EdwinGuzman (approved):

avertrees (approved):

PR #205, opened 2024-10-01
DR-3191: Refactor Lane component

7emansell:
## Ticket:

Resolve JIRA ticket [DR-3191](https://newyorkpubliclibrary.atlassian.net/browse/DR-3191)

- Removes `CollectionLanes` and `ItemLane`... one `Lane` to rule them all
- Passes responsibility of mapping the data over the lanes to parent component (instead of passing all the lanes to the `Lane` component)

## Open Questions

Should this component be called `DCLane` for consistency? Even though there is no DS `Lane` component to distinguish it from?
Currently, called `Lane` and _exported as_ `DCLane`, which I also did to the `Card` and `SimpleGrid`

Everything should look the same!

### Checklist:

[DR-3191]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman:
Very nice refactor of the components into a single one. Added comments, but none are really blockers.

app/src/components/lane/lane.tsx
```diff
@@ -0,0 +1,180 @@
...
+              },
+            color: "ui.link.primary",
+            fontWeight: "medium",
+            alignSelf: "end",
+            _hover: {
+              textDecoration: "underline 1px dotted !important",
+              textUnderlineOffset: "2px",
+              "> svg": { fill: "ui.link.secondary" },
+            },
+            _visited: { "> svg": { fill: "ui.link.tertiary" } },
```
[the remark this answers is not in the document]
7emansell: I like to have them all there, it makes it more readable. A solution here is simply deleting the CSS module and having all the styles inline, which I just did, but I know that is controversial...

app/src/types/CollectionDataType.ts
```diff
@@ -5,7 +5,7 @@ export interface CollectionDataType {
   imageID: string | null;
   numberOfDigitizedItems: number;
   containsOnSiteMaterials: boolean;
-  containsAVMaterial: boolean;
+  containsAVMaterial?: boolean;
```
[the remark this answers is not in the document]
7emansell: This was a mistake ! It always should

avertrees (approved):

app/src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -1,32 +1,53 @@
...
   return isLoaded ? (
     <>
-      <CollectionLanes
-        lanesWithNumItems={[data.lanesWithNumItems[0]]}
-        seeMoreLink={`${DC_URL}/collections/lane`}
-      />
-      <FeaturedContentComponent randomNumber={data.randomNumber} />
-      <CollectionLanes
-        lanesWithNumItems={data.lanesWithNumItems.slice(1)}
+      <div style={{ marginTop: "-3rem" }} />
```
EdwinGuzman, line 27: Why did you have to add this?

app/src/components/homePageMainContent/homePageMainContent.tsx
```diff
@@ -1,32 +1,53 @@
...
+      <DCLane
+        key={0}
```
EdwinGuzman, line 29: I think this will conflict with the `0` below in the map. Do you see anything in the console?

app/src/components/homePageMainContent/homePageMainContent.tsx
7emansell: The `Lane` component has a top margin which is needed in all cases except the first lane on the homepage, which needs to be closer to the hero. Since now the lane component is represented as a singular lane, which the parent component replicates as needed, I think the parent component should also be responsible for removing the margin on the first child (instead of the `Lane`s, plural, innately knowing the first one needs to not have a margin). I could do this by declaring a separate CSS module with a first child rule, or just do this.

app/src/components/homePageMainContent/homePageMainContent.tsx
7emansell: No console issues (at least not more than we usually have)

EdwinGuzman (approved):
Really good refactor!

PR #208, opened 2024-10-07
DR-3193/refactor grid component

avertrees:
## Ticket:

- JIRA ticket [Refactor Collection/Item Grid Component](https://newyorkpubliclibrary.atlassian.net/browse/DR-3193)

- creates one `CardsGrid` component to handle both item/collection card pages

## Open Questions

-

-

### Checklist:


app/src/components/grids/cardsGrid.tsx
```diff
@@ -0,0 +1,55 @@
+"use client";
+
+import React from "react";
+import useBreakpoints from "../../hooks/useBreakpoints";
+import CollectionDataType from "../../types/CollectionDataType";
+import { CollectionCardModel } from "../../models/collectionCard";
+import ItemDataType from "@/src/types/ItemDataType";
+import ItemCardDataType from "@/src/types/ItemCardDataType";
```
7emansell, line 8: nit unused

app/src/components/grids/cardsGrid.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ItemCardModel } from "@/src/models/itemCard";
+import { SimpleGrid as DCSimpleGrid } from "../simpleGrid/simpleGrid";
+import { Card as DCCard } from "../card/card";
+
+interface CardsGridProps {
+  records: CollectionDataType[] | ItemDataType[];
+}
+
+function isCollectionType(
```
7emansell, line 17: maybe make this an export here, or import it from lanes

7emansell (commented):
Looks good, it might also be a good time to write unit tests for this

app/src/components/grids/cardsGrid.tsx
```diff
@@ -0,0 +1,55 @@
...
+      {records?.map((record, index) => {
+        if (isCollections) {
+          const collectionCardModel = new CollectionCardModel(record);
+          return (
+            <DCCard
+              key={index}
+              id={index}
+              slug={collectionCardModel.title}
+              record={collectionCardModel}
+              isLargerThanLargeTablet={isLargerThanLargeTablet}
```
EdwinGuzman, line 38: Do these `DCCard`s need the `ref` and `tooltipOffset` props?

app/src/components/grids/cardsGrid.tsx
```diff
@@ -0,0 +1,55 @@
...
+  records: CollectionDataType[] | ItemDataType[];
+}
+
+function isCollectionType(
+  records: CollectionDataType[] | ItemDataType[]
+): records is CollectionDataType[] {
+  return "numberOfDigitizedItems" in records[0];
+}
+
+export const CardsGrid = ({ records }: CardsGridProps) => {
```
EdwinGuzman, line 23: This looks good. It's already being used in this PR but will it also be used in the `Lane` component? This similar DOM exists there so you could use this component there.

If you do use it, however, then you might not need the type guard or `useBreakpoints` since that already happens in the `Lane`. I imagine that will break functionality in `collectionLanePage.tsx` or `divisionPage.tsx`? It's fine if you can't reuse the component in `Lane` but just wanted to point it out.

EdwinGuzman (commented):
Looks good but just a few questions.

app/src/components/grids/cardsGrid.tsx
avertrees: ah, good catch. I think so.

app/src/components/grids/cardsGrid.tsx
avertrees: Open to having a conversation about this but atm I would like to separate the `Lane` from this component. Conceptually they are different. Lane as a see more link and always show 4 cards,`CardsGrid` can show any amount of cards & doesn't expect to have a see more link.

app/src/components/grids/cardsGrid.tsx
avertrees: I only think I need to add this to the collection cards for now, not the item cards?

7emansell: Both need it

app/src/components/grids/cardsGrid.tsx
avertrees: moved it to the utils file

app/src/components/grids/cardsGrid.tsx
EdwinGuzman: The `CardsGrid` could also be flexible enough to support four cards or x amount and conditionally show the "see more" link. But not needed right now.

7emansell (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Updated
 
 - Refactored collection/item lanes into one `Lane` component (DR-3191)
+- Refactored collections/items grid into one `CardsGrid` component (DR-3193)
```
EdwinGuzman, line 13: After you merge from `main`, move this up so it's under the new "Unreleased" version

EdwinGuzman (approved):

PR #214, opened 2024-10-21
DR-3056: Next Image in Card component

7emansell:
## Ticket:

- JIRA ticket [DR-3056](https://newyorkpubliclibrary.atlassian.net/browse/DR-3056)

- Creates `cardImage` component that uses Next's Image

### Checklist:


app/src/components/exploreFurther/exploreFurther.tsx
```diff
@@ -49,13 +50,20 @@ const ExploreFurther = () => {
...
             imageProps={{
-              alt: "",
-              aspectRatio: "sixteenByNine",
-              component: undefined,
+              component: (
+                <Image
+                  src={`/${item.image}`}
+                  alt={""}
+                  width={225}
+                  height={126}
```
EdwinGuzman, line 58: For local images, the next `Image` should be able to determine the width and height automatically, so you shouldn't need these. I'm not sure about "sizes" or "style" though. I'm assuming it's for the aspect ratio?

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
+import Image from "next/image";
+import React from "react";
+import { useEffect, useRef, useState } from "react";
+import { isCollectionCardDataType } from "./card";
+
+const CardImage = ({ record }) => {
+  const [imageHeight, setImageHeight] = useState(144);
```
EdwinGuzman, line 7: Set `144` in a constant variable to reduce magic numbers.

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
...
+        sizes="(max-width: 480px) 100vw, (max-width: 1024px) 50vw, 25vw"
+        style={{
+          width: "100%",
+          minHeight: "100%",
+        }}
+        width={288}
+        height={144}
+        onError={(_event) => {
+          console.warn(
+            `Card image failed to load, fallback image loaded instead. ImageURL: ${record.imageURL}`
```
EdwinGuzman, line 55: Prefix this with `CardImage: Card image failed ...` so it's easier to find where the warning is coming from.

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <div
+      ref={imageRef}
+      style={{
+        overflow: "hidden",
+        height: imageHeight,
+      }}
+    >
+      <Image
+        src={imageSrc}
+        alt=""
```
EdwinGuzman, line 38: There might be some occasions where alt text is needed and images are not decorative, so include this as a prop, perhaps from `record`'s title?

EdwinGuzman: Actually, you might need a new prop that is something along the lines of `isDecorative`. If it's false, then include the alt text, otherwise it should be empty. So far, all images we've encountered that this component can render has been decorative, right? If so, this is not needed right now.

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
...
+        id={
+          record.imageID
+            ? isCollection
+              ? `image-${record.title}-${record.imageID}`
```
EdwinGuzman, line 42: The `.title` may not be a slug so it will not work for the `id` attribute (`id`s cannot have white space).

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
+import Image from "next/image";
+import React from "react";
+import { useEffect, useRef, useState } from "react";
+import { isCollectionCardDataType } from "./card";
+
+const CardImage = ({ record }) => {
```
EdwinGuzman, line 6: Give `record` a type/interface.

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
...
+        setImageHeight(0.5 * imageWidth);
+      }
+    };
+    updateImageHeight();
+    window.addEventListener("resize", updateImageHeight);
+
+    return () => window.removeEventListener("resize", updateImageHeight);
+  }, []);
+
+  const [imageSrc, setImageSrc] = useState(
```
avertrees, line 24: nice

app/src/components/card/cardImage.tsx
```diff
@@ -0,0 +1,64 @@
...
+        const imageWidth = image.offsetWidth;
+
+        setImageHeight(0.5 * imageWidth);
+      }
+    };
+    updateImageHeight();
+    window.addEventListener("resize", updateImageHeight);
+
+    return () => window.removeEventListener("resize", updateImageHeight);
+  }, []);
```
EdwinGuzman, line 22: This means this runs for every image. Sometimes, the page feels laggy but I'm not sure if it's because of this. Let's revisit this for a later PR

EdwinGuzman (changes requested):
Some comments. The image rendering seems fast but you can also tell that the lazy loading is not there anymore (but not a big deal).

EdwinGuzman:
This warning comes up on the homepage and it's for every instance of the next image. It logs on load but also when resizing. Try to fix this.

Screenshot 2024-10-21 at 3 55 00 PM

app/src/components/card/cardImage.tsx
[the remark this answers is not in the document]
7emansell: I don't know... I think i made this up, i'm removing it

app/src/components/card/cardImage.tsx
7emansell: Yes this was the approved alt text for all card images so far. Will change once Clare gives the go ahead

app/src/components/exploreFurther/exploreFurther.tsx
7emansell: I'm not sure why, but it doesn't recognize these images as public, so there's an error if I remove these props

app/src/components/card/cardImage.tsx
7emansell: Pulled this out to the lane/grid level (same as the tooltip offset hook) which I had totally forgotten but that had been my plan since I did the tooltip. Actually i think i could go a step further and combine those two hooks since they're using the same thing (the image height) to do some calculations then return a number. What do you guys think, is that confusing to have one hook return two things ? @avertrees @EdwinGuzman

EdwinGuzman: It's a bit confusing and ideally a hook would return one value. I'd say it's okay to put them together since they are doing pretty much the same logic, but keep them separate. It'll be two hooks but once you move it up in the DOM tree, it'll be two hooks running instead of the 24+ for each image on the page.

app/src/hooks/useCardImageHeight.ts
```diff
@@ -0,0 +1,20 @@
+import { useState, useEffect } from "react";
+
+export function useCardImageHeight(cardRef) {
```
EdwinGuzman, line 3: Eventually, this hook needs a set of tests.

EdwinGuzman: Actually, none of the hooks in DC have tests.

For now, can you add jsdoc style comments
```
/**
 * ....
 */
```

app/src/components/exploreFurther/exploreFurther.tsx
```diff
@@ -49,13 +50,20 @@ const ExploreFurther = () => {
...
+                  sizes="(max-width: 600px) 100vw, 25vw"
```
EdwinGuzman, line 59: Add `deviceSizes` in the config file to make use of this prop https://nextjs.org/docs/app/api-reference/components/image#devicesizes

app/src/components/exploreFurther/exploreFurther.tsx
EdwinGuzman: Ah, it's because the image file need to be imported: https://nextjs.org/docs/app/building-your-application/optimizing/images#local-images

I think it's worth trying this to make use of the optimization but I think it'll make the loop trickier to work with.

EdwinGuzman (commented):
Good update, just a few minor comments.

app/src/components/exploreFurther/exploreFurther.tsx
7emansell: Lmk what you think on this approach with the image added to the `exploreFurtherData`

app/src/components/card/cardImage.tsx
7emansell: I think that's how it is rn unless you mean combining them in one file. But currently 2 hooks on the lane/grid

EdwinGuzman (approved):
Yes, I think the latest approach of importing the images in the data file is better than the component. Nice!

app/src/components/card/cardImage.tsx
EdwinGuzman: Right, I wasn't too clear but keep them as 2 hooks as it exists now.

app/src/types/ExploreFurtherDataType.ts
```diff
@@ -1,7 +1,8 @@
+import { StaticImageData } from "next/image";
```
avertrees, line 1: also nice

avertrees (approved):
looks great!

PR #216, opened 2024-10-29
DR-3107: Connect swim lane landing page to API

7emansell:
## Ticket:

Resolves JIRA ticket [DR-3107](https://newyorkpubliclibrary.atlassian.net/browse/DR-3107)

- Connects swim lane landing page to API
- Handles pagination ref

## Open Questions

- The loading is horrible. Don't know what to do about this (I don't *think* it's the fault of the Next Image...) but either way would appreciate thoughts

`collections/lane/recently-digitized-collections?page=2#recently-digitized-collections` 

- Pagination ref (see question)

### Checklist:

[DR-3107]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman:
> Should the url have the slug hash at the end? This is copied from the division landing page

This makes sense for consistency but the page doesn't go to the heading on load. It should work like the `#public_domain` hash but I can't tell what's different about this hash or heading. (BTW, the suggested hash here and in divisions uses hyphens but public_domain has an underscore).

app/src/components/grids/cardsGrid.tsx
```diff
@@ -16,44 +16,49 @@ interface CardsGridProps {
...
-              record={itemCardModel}
-              isLargerThanLargeTablet={isLargerThanLargeTablet}
-            />
-          );
-        }
-      })}
-    </DCSimpleGrid>
-  );
-};
+    //console.log("first record", records[0]);
```
avertrees, line 27: nit: remove comment

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
```diff
@@ -3,27 +3,50 @@ import {
...
+
+  const { replace } = useRouter();
+
+  const totalPages = totalNumPages(data.numResults, data.perPage);
+
+  const updatePageURL = async (pageNumber: number) => {
+    const params = new URLSearchParams();
+    params.set("page", pageNumber.toString());
+    setCurrentPage(pageNumber);
+    const url = `${pathname}?${params.toString()}#${slug}`;
```
avertrees, line 45: tl;dr we don't need the "#{slug}" here, that was added to the divisions landing page because I added an anchor link to the Collections header since the paginated component (grid of Collections cards) on the divisions landing page starts in the middle of the page and you have to scroll down to view the pagination menu.

avertrees:
> The loading is horrible. Don't know what to do about this (I don't think it's the fault of the Next Image...) but either way would appreciate thoughts

The loading on the /divisions/slug pages are also horribly slow and that was pre next image... it feels Next.js specific and IDK what to do about it either. We can investigate this after I get back.

app/src/components/grids/cardsGrid.tsx
```diff
@@ -16,44 +16,49 @@ interface CardsGridProps {
...
 }
 
-export const CardsGrid = ({ records }: CardsGridProps) => {
-  const { isLargerThanLargeTablet } = useBreakpoints();
-  const isCollections = isCollectionType(records);
-  const cardRef = useRef<HTMLDivElement>(null);
-  const tooltipOffset = useTooltipOffset(cardRef);
-  const imageHeight = useCardImageHeight(cardRef);
+export const CardsGrid = forwardRef<HTMLDivElement, CardsGridProps>(
+  ({ records }, ref?) => {
```
avertrees, line 20: > How do we feel about the ref going to the first card when the page changes?

Think it makes sense. See what Claire says but I don't see a problem with this? Maybe also ask Lana as she can fill in for Lily atm. @EdwinGuzman what do you think?

7emansell:
Claire also recommended the pagination behavior we discussed (hidden "Page x of y" that appears when you increment the page). She pointed out that there's no other affordance for the fact that the page has changed, other than scrolling all the way down to see the bolded page number, so this seems absolutely necessary. Bringing this to Lana as a pattern that should be consistent throughout DCF @EdwinGuzman

app/src/components/grids/cardsGrid.tsx
7emansell: see [comment](https://github.com/NYPL/digital-collections/pull/216#issuecomment-2450181288)

7emansell:
Pagination focus behavior approved by Lana

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
```diff
@@ -49,19 +72,43 @@ export default function CollectionLanePage() {
...
+      <Heading
+        size="heading6"
+        sx={{
+          height: "1px",
+          left: "-10000px",
+          overflow: "hidden",
+          width: "1px",
+          _focus: {
+            fontWeight: "400 !important",
+            height: "auto",
```
EdwinGuzman, line 87: The page will not jump if you can give it a defined height all the time. Otherwise, once you focus away from the heading, the grid jumps up a bit.

EdwinGuzman (commented):
Some comments for now. It'd be nice to have the "Page x of y" visible all the time (maybe not for the first page).

I want to investigate the loading state more before approving. Sometimes, the focus goes to the string and a few seconds later the updated page data renders.

As a side note, if the data is fetched and rendered quickly, the change is almost instance, but you can tell that the images are still taking some time to load. I don't think it's a problem but some IIIF images do render very slowly.

app/src/components/card/cardImage.tsx
```diff
@@ -35,12 +37,14 @@ export const CardImage = ({ record, imageHeight }: CardImageProps) => {
...
         width={initialImageHeight * 2}
         height={initialImageHeight}
+        decoding="sync"
         onError={(_event) => {
           console.warn(
             `CardImage: Card image failed to load, fallback image loaded instead. ImageURL: ${record.imageURL}`
           );
           setImageSrc("/noImage.png");
         }}
+        placeholder="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mN8WQ8AAlcBas53/MIAAAAASUVORK5CYII="
```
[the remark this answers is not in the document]
7emansell: It's meant to be just a gray rectangle (this format is what Next wants for some reason). I don't want to use the `noImage` as the placeholder because it'll be confusing when the image does load

7emansell:
Approaches tried (but not necessarily implemented correctly?):
1. Creating a `loading.tsx` file at the level of the `CardsGrid` component
2. Creating a `loading.tsx` file at the level of the page component
3. Using `useEffect` and an `isLoading` state to switch between skeleton loader and the actual cards on the `CardsGrid` component
4. Same thing on the `Card`
5. Same thing at the page component level
6. Setting `isLoading` true (then false) in `updatePageUrl`
7. Setting a 2 second timeout that sets `isLoading` true (then false) in `updatePageUrl` 
8. Using the `placeholder` prop on the card image
9. Using the `onLoad` prop on the card image to set `isLoading` true/false and having that state go up to the Card/CardsGrid/page component 
@avertrees

app/src/components/card/cardImage.tsx
```diff
@@ -35,12 +37,14 @@ export const CardImage = ({ record, imageHeight }: CardImageProps) => {
         }}
         width={initialImageHeight * 2}
         height={initialImageHeight}
+        decoding="sync"
```
EdwinGuzman, line 40: Have you noticed if this prop has made any impact?

7emansell: It has not that I've noticed (was switching it between "sync" and "async" for a while during experimentation)

7emansell: But i think theoretically we do want it to be "sync"

7emansell:
VQA approved i just never got final dev approval on this

package.json
```diff
@@ -46,6 +46,7 @@
     "next": "14.2.3",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "sharp": "^0.33.5",
```
EdwinGuzman, line 49: nit: remove the caret.

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
```diff
@@ -3,29 +3,57 @@ import {
...
+
+  const headingRef = useRef<HTMLHeadingElement>(null);
+
+  const updatePageURL = async (pageNumber: number) => {
+    const params = new URLSearchParams();
+    params.set("page", pageNumber.toString());
+    setCurrentPage(pageNumber);
+    const url = `${pathname}?${params.toString()}`;
+    setIsLoaded(false);
+    push(url);
```
EdwinGuzman, line 51: What I'm seeing locally is:

- I click on "next"
- the skeleton loaders display
- the focus goes to the text "displaying ..."
- then the page URL updates and the new data comes in

Let's keep investigating this but the order might need to move around or the timeout should be longer.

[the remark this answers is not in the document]
7emansell: Do you feel like what you're seeing is the worst case loading time? I thought we had already cleared this loading pattern a week ago and it's going to prod tomorrow on the division landing page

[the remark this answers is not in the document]
7emansell: @avertrees looping you in here as well– would we be comfortable setting a longer loading time, maybe specifically on the division landing page? @EdwinGuzman I feel 90% confident that the order of operations is the best we can do / the most logical, and has been approved by Clare, so the place for refinement is in the loading timeout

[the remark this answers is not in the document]
EdwinGuzman: Longer than 2seconds?

EdwinGuzman (approved):

app/src/components/pages/collectionLanePage/collectionLanePage.tsx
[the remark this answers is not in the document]
avertrees: would updating the divisions landing page load timeout be set to a time consistent with all the pages?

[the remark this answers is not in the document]
7emansell: Wait not sure what you're asking but the timeout can be set specific to endpoint

avertrees (approved):

PR #226, opened 2024-11-12
updated NR transactions

EdwinGuzman:
## Ticket:

- JIRA ticket [DR-3261](https://newyorkpubliclibrary.atlassian.net/browse/DR-3261)

- This is an update from the previous PR #225 .
- This one actually captures and names transactions and sends it to NR. This is, perhaps because of the new setup and configuration, the new way to do this. It's manual and not great but it works. 
- This does not work for API routes but I'm not sure if those were previously captured. 
- What I can't get to work is the 404 page. If I add a 404 page, _all_ pages get grouped there. 
- This gets us 90% of the way there and it's better than before. We can iterate on this but we do need deeper NR knowledge.

Resources
- https://docs.newrelic.com/docs/apm/agents/nodejs-agent/api-guides/nodejs-agent-api/#rules This was the most helpful one and it's hidden under the "URL pattern matching" accordion (ugh)
- https://github.com/newrelic/node-newrelic/blob/904f41b26637394a24aa13f31ff94b100ae6d090/documentation/nextjs/transactions.md Ideally, this should be automatic but it's not...

## Open Questions

How do you feel about this approach?  Another working approach is to import `newrelic` into _server_ components and then add `newrelic.setTransactionName('MyCustomTransaction');` inside of the component function. Unfortunately, this does not work for API routes.

Go to https://one.newrelic.com/nr1-core/apm-features/transactions/MTIxMzM0fEFQTXxBUFBMSUNBVElPTnwxNTg1OTM0Njk3?account=121334&duration=86400000&state=9ad27e64-da16-c3bc-d04c-1dce045b8aa0
Update the time range in the upper right corner to ~24hrs.

Screenshot 2024-11-12 at 9 49 43 AM

- N/A

### Checklist:

[DR-3261]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newrelic.js
+++ b/newrelic.js
@@ -43,4 +43,18 @@ exports.config = {
       "response.headers.x*",
     ],
   },
+  rules: {
+    name: [
+      { pattern: "/api/lanes/*", name: "Nextjs/GET//api/lanes/[slug]/" },
+      { pattern: "/api/lanes", name: "Nextjs/GET//api/lanes" },
+      { pattern: "/api/homepage", name: "Nextjs/GET//api/homepage" },
+      { pattern: "/api/feedback", name: "Nextjs/POST//api/feedback" },
+      { pattern: "/api/featuredItem", name: "Nextjs/GET//api/featuredItem" },
+      { pattern: "/divisions/*", name: "/Nextjs/GET/divisions/[slug]/" },
+      { pattern: "/divisions", name: "Nextjs/GET//divisions" },
+      { pattern: "/about", name: "Nextjs/GET//about" },
+      { pattern: "/404", name: "Nextjs/GET//404" },
+      { pattern: "/", name: "Nextjs/GET//" },
+    ],
+  },
 };
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,4 @@
 /** @type {import('next').NextConfig} */
-
-// const nrExternals = require("@newrelic/next/load-externals");
 // This is found in the codebase example but both seem to work and send data
 // correctly. Keep the above but can use the bottom for debugging.
 const nrExternals = require("newrelic/load-externals");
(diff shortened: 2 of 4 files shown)


newrelic.js
```diff
@@ -43,4 +43,18 @@ exports.config = {
...
+      { pattern: "/api/homepage", name: "Nextjs/GET//api/homepage" },
+      { pattern: "/api/feedback", name: "Nextjs/POST//api/feedback" },
+      { pattern: "/api/featuredItem", name: "Nextjs/GET//api/featuredItem" },
+      { pattern: "/divisions/*", name: "/Nextjs/GET/divisions/[slug]/" },
+      { pattern: "/divisions", name: "Nextjs/GET//divisions" },
+      { pattern: "/about", name: "Nextjs/GET//about" },
+      { pattern: "/404", name: "Nextjs/GET//404" },
+      { pattern: "/", name: "Nextjs/GET//" },
+    ],
+  },
```
avertrees, line 59: I think this approach is cleaner than adding `newrelic.setTransactionName('Nextjs/GET//about');` to the server component but maybe that's just me. 

Regardless, we'll need to remember to add a line of code somewhere for New Relic to pick up any new pages. It might be worth adding the phase 3 and phase 4 routes to this file now. 

Regardless, this approach is acceptable to me.

EdwinGuzman: I don't want to add any routes which are not live in prod just yet. There's little risk of doing so but don't want anyone to get the wrong idea of seeing those events captures and thinking that it's somehow live to patrons.

7emansell (approved):

avertrees (approved):

PR #244, opened 2024-11-20
DR-3297: Rewrite getHomePageData to no longer use `items/counts`

7emansell:
## Ticket:

- JIRA ticket [DR-3297](https://newyorkpubliclibrary.atlassian.net/browse/DR-3297)

- Adds new `getItemsCountFromCollections()` which hits `/collections/{uuid}`  to get `numItems` for each passed `uuid` (really only a helper for `getHomePageData`)
- NOTE: this is declared and tested in a separate file because of Jest mocking conventions. When I get to this https://newyorkpubliclibrary.atlassian.net/browse/DR-3271 I'll refactor all of the API helper tests and it'll be less weird

Go to the homepage and think to yourself, is this slower than usual 

### Checklist:

[DR-3297]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/pages/divisionsPage/divisionsPage.tsx
+++ b/app/src/components/pages/divisionsPage/divisionsPage.tsx
@@ -33,7 +33,7 @@ export default function DivisionsPage({ summary, divisions }: DivisionsProps) {
       ]}
       adobeAnalyticsPageName={createAdobeAnalyticsPageName("divisions")}
     >
-      {divisions && divisions.length > 0 ? (
+      {divisions && divisions?.length > 0 ? (
         <>
           <Box
             sx={{
--- a/app/src/utils/itemCount.ts
+++ b/app/src/utils/itemCount.ts
@@ -7,17 +7,17 @@ import { apiResponse } from "./api";
 export const getItemsCountFromCollections = async (uuids: string[]) => {
   const results = {};
   const apiUrl = `${process.env.API_URL}/api/v2/collections/`;
-  await Promise.all(
-    uuids.map(async (uuid) => {
-      try {
-        const response = await apiResponse(`${apiUrl}${uuid}`);
-        const numItems = response.numItems || 0;
-        results[uuid] = numItems;
-      } catch (error) {
-        console.error(`Error fetching collection ${uuid}:`, error.message);
-        results[uuid] = 0;
-      }
-    })
-  );
+  const fetchPromises = uuids.map(async (uuid) => {
+    try {
+      const response = await apiResponse(`${apiUrl}${uuid}`);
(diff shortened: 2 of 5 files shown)


avertrees (approved):

app/src/components/pages/divisionsPage/divisionsPage.tsx
```diff
@@ -33,7 +33,7 @@ export default function DivisionsPage({ summary, divisions }: DivisionsProps) {
       ]}
       adobeAnalyticsPageName={createAdobeAnalyticsPageName("divisions")}
     >
-      {divisions && divisions.length > 0 ? (
+      {divisions && divisions?.length > 0 ? (
```
EdwinGuzman, line 36: wow sneaky?

app/src/utils/itemCount.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Returns a map of collection uuids with corresponding item counts.
+ * @param {string[]} uuids - collection uuids
+ */
+export const getItemsCountFromCollections = async (uuids: string[]) => {
+  const results = {};
+  const apiUrl = `${process.env.API_URL}/api/v2/collections/`;
+  await Promise.all(
+    uuids.map(async (uuid) => {
```
EdwinGuzman, line 11: This is me being nitpicky but the syntax is interesting. This works so no need to change but I would generally prefer to create the list of unresolved promises in a variable and then pass it to the `Promise.all` call. Just minor readability comment.

EdwinGuzman (approved):
Is the backend updating their cached time concurrently with this update? It'd be good to just test this to keep all the different variable separate so we know what works.

7emansell:
Backend is making no changes until we see the outcome of this update as far as I know @EdwinGuzman

app/src/utils/itemCount.ts
7emansell: Is this what you're thinking ?

app/src/components/pages/divisionsPage/divisionsPage.tsx
avertrees: yes

avertrees: as per my request @EdwinGuzman

EdwinGuzman (approved):
Yep, looks good

PR #252, opened 2024-11-26
DR-3293 fix restricted thumbnails

chrisb87:
## Ticket: 

- JIRA ticket [DR-3293](https://newyorkpubliclibrary.atlassian.net/browse/DR-3293)

- Updates thumbnail logic to prevent tall images from having restricted thumbnails

Previously thumbnails were generated by requesting an image 288px wide, and whatever height would keep the aspect ratio intact. For tall images this would sometimes result in a thumbnail taller than 300px, which surpasses the limit for public derivatives.

With this change, thumbnails are always 288x288px and cropped if the source image is not square. I met with Lana and Lilly and this was the solution we landed on as the best option for thumbnails of images of varying shapes.


avertrees (approved):
wow okay, never thought I'd be happy about the "square" parameter

7emansell (approved):
Looks great

PR #253, opened 2024-11-26
DR-3271: API helper test coverage

7emansell:
## Ticket:

- JIRA ticket [DR-3271](https://newyorkpubliclibrary.atlassian.net/browse/DR-3271)

- Moves `apiResponse` into its own file, and all helpers into their own file, `apiHelpers`
- Adds tests for `apiResponse` and `apiHelpers`

## Open Questions

- A lot of the `apiHelpers` tests are very redundant, since they're pretty much just getters, and if there was an issue, the `apiResponse` tests are really where it'd show up. Not sure if we want to just delete those redundant getter tests and only test the helpers with more involved fallback behavior? Or is it worth it to write out mock responses for all of them ? 
     - `mockApiResponses` now has full mocks
- Is `apiResponse` the right name for that function ?

### Checklist:

[DR-3271]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/divisions/page.tsx
```diff
@@ -13,6 +13,7 @@ export const metadata: Metadata = {
 
 export default async function Divisions() {
   const data = await getDivisionData();
+  console.log(data);
```
EdwinGuzman, line 16: Remove the log.

__tests__/__mocks__/data/mockApiResponses.tsx
```diff
@@ -0,0 +1,86 @@
+export const mockFeaturedItemResponse = {
+  headers: { status: "success", code: "200", message: "ok" },
+  numResults: "1",
+  capture: {
+    uuid: "510d47d9-7c7c-a3d9-e040-e00a18064a99",
+    imageLinks: { imageLink: [Array] },
```
EdwinGuzman, line 6: I'm surprised this doesn't throw an error when it's run. Replace `Array` with mocked data.

__tests__/__mocks__/data/mockApiResponses.tsx
```diff
@@ -0,0 +1,86 @@
...
+    apiUri:
+      "http://api.repo.nypl.org/api/v2/items/mods/510d47d9-7c7c-a3d9-e040-e00a18064a99",
+    typeOfResource: "still image",
+    imageID: "54795",
+    sortString: "0000000001|0000000011|0000000430|0000000001",
```
EdwinGuzman, line 11: Out of curiosity, do either of you know what this means?

__tests__/__mocks__/data/mockApiResponses.tsx
```diff
@@ -0,0 +1,86 @@
...
+    version: "3.4",
+    schemaLocation:
+      "http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-4.xsd",
+    titleInfo: {
+      type: "",
+      authority: "",
+      usage: "primary",
+      lang: "",
+      script: "",
+      title: [Object],
```
EdwinGuzman, line 39: Same here and throughout the file. This may not throw an error when used in tests, but update the fixture to be closer to real data by updating `[Object]` and `[Array]`

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+} from "__tests__/__mocks__/data/mockApiResponses";
+
+jest.mock("./apiResponse");
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("getHomePageData", () => {
+  it("creates response containing random number and all 7 lanes", async () => {
```
EdwinGuzman, line 24: 👍

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+});
+
+describe("getFeaturedItemData", () => {
+  it("creates response containing featuredItem and numDigitizedItems", async () => {
+    const result = await getFeaturedItemData();
+
+    // Fallback values:
+    expect(result.numberOfDigitizedItems).toEqual("875,861");
+    expect(result.featuredItem.imageID).toEqual(
+      defaultFeaturedItem.production.featuredItem.imageID
```
EdwinGuzman, line 59: note this will need an update because of #254

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+    expect(result.lanesWithNumItems.length).toEqual(7);
+    // Test uuid response doesn't match the collection uuid, so all of these should be 0.
+    expect(
+      result.lanesWithNumItems[0].collections[3].numberOfDigitizedItems
+    ).toEqual("0");
+  });
+});
+
+describe("getFeaturedItemData", () => {
+  it("creates response containing featuredItem and numDigitizedItems", async () => {
```
EdwinGuzman, line 53: This is testing the default values when `getFeaturedImage` and `getNumDigitizedItems` don't return data from their own API call. I think it's fair to only test this function through the default response and test the other two functions separately (so you don't need an integration test). Just make a note in the description that it's returning default fallback data.

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+    expect(result.numberOfDigitizedItems).toEqual("875,861");
+    expect(result.featuredItem.imageID).toEqual(
+      defaultFeaturedItem.production.featuredItem.imageID
+    );
+  });
+});
+
+describe("getDivisionData", () => {
+  it("forms the correct request from params", async () => {
+    const result = await getDivisionData({
```
EdwinGuzman, line 66: Nit: you're not using `result` so you can remove it.

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+      slug: "testSlug",
```
EdwinGuzman, line 67: Add a test where `slug` is _not_ passed. In the function, the `if (slug)` is a branch that should be tested (which you did here) but add a test for when the `if` statement is false.

This can conceptually also be, one test for the main divisions page and one test for the division/slug page.

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+      pageNum: 1,
+      perPage: 3,
+    });
+
+    expect(apiResponse as jest.Mock).toHaveBeenCalledWith(
+      `${process.env.API_URL}/api/v2/divisions/testSlug?page=1&per_page=3`
+    );
+  });
+  it("returns successful response", async () => {
```
EdwinGuzman, line 76: What about when the `apiResponse` throws an error? How does `getDivisionData` handle it? Add a test.

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+    (apiResponse as jest.Mock).mockResolvedValueOnce(
+      Promise.resolve({
+        count: {
+          $: 78,
+        },
+      })
+    );
+    const result = await getNumDigitizedItems();
+
+    expect(result).toEqual("78");
```
EdwinGuzman, line 115: I like the expectation you have on line 72. Can you add a similar one here to test/verify the URL that is called?

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+});
+
+describe("getRandomFeaturedItem", () => {
+  it("returns expected item", async () => {
+    (apiResponse as jest.Mock).mockResolvedValueOnce(
+      Promise.resolve(mockFeaturedItemResponse)
+    );
+    const item = await getRandomFeaturedItem();
+    expect(item).toEqual(mockFeaturedItemResponse);
+    expect(item).toHaveProperty("capture");
```
EdwinGuzman, line 187: Similar comment as above, add an expectation for the URL that is called. I think that's a good test.

app/src/utils/apiHelpers.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+    expect(item.imageID).toEqual("482815");
+  });
+});
+
+describe("getItemData", () => {
+  it("returns expected item", async () => {
+    (apiResponse as jest.Mock).mockResolvedValueOnce(
+      Promise.resolve(mockItemResponse)
+    );
+    const item = await getRandomFeaturedItem();
```
EdwinGuzman, line 207: This should be `getItemData`. This brings up a good point in these tests because this test still passes: many of these functions are wrappers around `apiResponse`. Essentially, all these tests are mocking and testing that internal API call and its response. Like other tests above, add an expectation that a specific URL was called. This doesn't entirely address my concern but gets closer to really testing the functions.

app/src/utils/apiHelpers.ts
```diff
@@ -132,18 +131,17 @@ export const getItemsCountFromUUIDs = async (uuids: string[]) => {
...
 
-export const getItemByIdentifier = async (
-  identifierType: string,
-  identifier: string,
-  urlParam?: { [key: string]: any }
-) => {
-  const apiUrl = `${process.env.API_URL}/api/v2/items/${identifierType}/${identifier}`;
-  return apiResponse(apiUrl, { params: urlParam });
+export const getRandomFeaturedItem = async () => {
+  const apiUrl = `${process.env.API_URL}/api/v2/items/featured`;
```
EdwinGuzman, line 138: Nice refactor.

EdwinGuzman (commented):
This is a really good update. Thanks for writing tests.

> A lot of the apiHelpers tests are very redundant, since they're pretty much just getters, and if there was an issue, the apiResponse tests are really where it'd show up. Not sure if we want to just delete those redundant getter tests and only test the helpers with more involved fallback behavior? Or is it worth it to write out mock responses for all of them ?

I brought this up in a test where the description for the function and the function that was being called did not match. So yes, perhaps many tests are redundant. I think it's worth it to mock out all the responses and/or verify the URL that is being called. Each function has a specific need an in many cases are called in multiple locations (API route and RSC file) so, to me, it's worth having a wrapper function for these calls.

> Is apiResponse the right name for that function ?

Did I make that? I think it made sense early one but it could be updated. I don't have a preference.

-----

We can discuss any of my feedback next week.

app/src/utils/apiHelpers.test.tsx
7emansell: I'm not sure how to go about doing this since `getDivisionData` doesn't catch or handle at all– the error handling is all in `apiResponse`.  Is that a flaw in these getters ?

7emansell:
> Is apiResponse the right name for that function ?

> Did I make that? I think it made sense early one but it could be updated. I don't have a preference.

No, I named it, just wondering if that still makes sense to everyone

app/src/utils/apiHelpers.test.tsx
EdwinGuzman: Yes, it's a bit of a flaw. We decided to let Next.js catch the thrown error that bubbles up in its own error page, so the app itself is handling this appropriately (e.g. the app does not crash or throw the unstyled 500 error page). If we catch the error and don't do anything, the app won't display the error page.

Coding-wise, these functions cannot really be used outside of this app, not that we would though. But if you're using a function like this outside of this nextjs app and it throws an error and you don't catch it, it is perhaps unexpected. You can (and should) type the functions or add more jsdoc/esdoc style comments to say explicitly it returns a data object or throws an error, `const getDivisionData : [typeof apiResponse] | Error = async ({ ...`

Again, we made this tradeoff and went with Next handling the thrown errors so overall, we're ok. For function completeness though, you can add a test for the thrown error that is similar to the `apiResponse` timeout  error
```
await expect(apiCall).rejects.toEqual(
      new Error("apiResponse: Request timed out")
    );
```

This is a very interesting topic so we can keep talking about this. This is an ecmascript proposal but I really like this idea: https://github.com/arthurfiorette/proposal-safe-assignment-operator
You'd end up with `const [error, response] ?= await fetch("https://api.example.com/data")` and similar for these getters.

Instead of throwing an error, the function returns the error and the response. Then at a higher level you can throw the error to let Nextjs handed it. It's lots of layers but this would be a good repo-wide project to work on. 

To be clear, for now, an error test is enough.

EdwinGuzman (commented):
The latest round looks good. From the thread in `apiHelpers.tests.tsx`, add a few more tests for the errors thrown. A good one would be for the new update in `getLaneData` where an error is thrown if the `slug` arg is not passed.

I think `apiResponse` is okay, it could be `fetchURL`, `fetchRepo`, or `fetchAPI`? It is a wrapper around fetch with the added timeout and repo API response parsing, so that might be helpful to know in the function name.

PR #255, opened 2024-12-02
DR-3100 connect all collections page to api

avertrees:
## Ticket:

- [Connect collections page to Repo API endpoint](https://newyorkpubliclibrary.atlassian.net/browse/DR-3100)
- [Create no results view for all collections search](https://newyorkpubliclibrary.atlassian.net/browse/DR-3324)

- connects /collections page to api endpoint
- search/sort/filter pagination
- uses loading and ref patterns implemented in collections/lane/:slug pages

## Open Questions

- debouncing? https://nextjs.org/learn/dashboard-app/adding-search-and-pagination#best-practice-debouncing. It could be nice and Next.js docs say it's a best practice, however I'm not sure how this fits with accessibility guidelines/UX. Could be implemented later but wanted to flag it now.
- should sort be reset to the default on search submit? 

- locally or in the vercel link
- search bar, sort menu, and pagination should all work
- if the fields are default, they should not be included in the url
- "cat" and "dog" searches return one result
- any string of random characters, such as "askdjakd", will return the no results page

-

### Checklist:


app/collections/page.tsx
```diff
@@ -10,6 +16,22 @@ export const metadata: Metadata = {
   },
 };
 
-export default async function Collections() {
-  return <CollectionsPage data={mockCollectionCards} />;
+export default async function Collections({ searchParams }: CollectionsProps) {
+  const data = await getCollectionsData({
+    keyword: searchParams.collection_keyword,
+    sortID: searchParams.sort,
+    pageNum: searchParams.page,
```
EdwinGuzman, line 23: I'm assuming that `collection_keyword`, `sort`, and `page` are all param names needed for Repo API?

This isn't wrong but try this syntax:
```
const { collection_keyword: keyword, sort: sortID, page: pageNum } = searchParams;
  const data = await getCollectionsData({
    keyword,
    sortID,
    pageNum,
  });
```

This is to separate the naming convention from the URL query params and the naming convention you need in the app.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
 } from "@nypl/design-system-react-components";
+import { useSearchParams, usePathname, useRouter } from "next/navigation";
 import { headerBreakpoints } from "../../../utils/breakpoints";
+import { CardsGrid } from "../../grids/cardsGrid";
+import LaneLoading from "../../lane/laneLoading";
+import { displayResults, totalNumPages } from "../../../utils/utils";
+import type { SyntheticEvent } from "react";
 import { createAdobeAnalyticsPageName } from "@/src/utils/utils";
 
 export const CollectionsPage = ({ data }) => {
```
EdwinGuzman, line 21: From a previous convo, pass down `searchParams` from the page component down to this client component. Then you don't need to use `useSearchParams`.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+  const { push } = useRouter();
+  const searchParams = useSearchParams();
+  const pathname = usePathname();
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  const collections = data.collection; // typeof data.collection === "object" ? [data.collection] : data.collection // this breaks empty search
```
EdwinGuzman, line 27: We talked about this earlier. Either check if `data.collections` is an array, or perhaps check if `data.collections.apiUri` exists. If it's true, then it's a single object which needs to be an array.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+  const headingRef = useRef<HTMLHeadingElement>(null);
+
+  // pagination
+  const numFound = data.numFound ? data.numFound : data.numResults;
```
EdwinGuzman, line 31: This can also be `data.numFound || data.numResults`.

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Then in lines 35-37, you can use the value directly such as `searchParams.page` rather than the `get` function.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+      setIsLoaded(true);
+      headingRef.current?.focus();
+    }, 2000);
+  };
+
+  // I'm not actually sure if this is necessary but it's in the research catalog
+  const handleSearchChange = (e: SyntheticEvent) => {
+    const target = e.target as HTMLInputElement;
+    currentCollectionKeyword = target.value;
+  };
```
EdwinGuzman, line 58: Yes. This works but change it to use state instead. This means you need to use `searchParams.collection_keyword` as the default state value such as `const [currentCollectionKeyword, setCurrentCollectionKeyword] = useState<string>(searchParams.collection_keyword || ");`

Then this function can use the `setCurrentCollectionKeyword` function to update the value.
You should not update a variable directly in this manner in order to reduce rerendering inconsistencies.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+
+  // pagination
+  // Question: Do we want to introduce debouncing?
```
EdwinGuzman, line 61: Yes.

EdwinGuzman: Actually, what would we need it for?

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+    setIsLoaded(false);
+    await push(`${pathname}?${queryString}`);
+    setTimeout(() => {
+      setIsLoaded(true);
+      headingRef.current?.focus();
+    }, 2000);
+  };
+
+  // TODO: right now, whenever a user hits earch, the url is created with all params including the defaults.
+  // Perhaps the query string should only include the values that are not defaults.
```
EdwinGuzman, line 78: Yes, update this function to remove the defaults.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+      params.set(
+        name.toString(),
+        name === "page" ? paramObj[name].toString() : paramObj[name]
+      );
+    });
+    return params.toString();
+  };
+
+  // sort
+  const onMenuClick = async (id) => {
```
EdwinGuzman, line 91: When I test this, "Z to A" returns results in A to Z and "A to Z" returns results in Z to A. So the ordering is off somewhere.

app/src/utils/utils.ts
```diff
@@ -106,6 +106,13 @@ export function isCollectionType(
   return "numberOfDigitizedItems" in records[0];
 }
 
+export const collectionsSortOptions = {
```
EdwinGuzman, line 109: 👍

app/src/utils/api.ts
```diff
@@ -165,6 +164,22 @@ export const getDivisionData = async ({
   return res;
 };
 
+export const getCollectionsData = async ({
+  keyword = "",
+  sortID = "date-desc", // TODO: rename sortID TODO: store defaults as constants
```
EdwinGuzman, line 169: Make this a type, such as `type sortIds = "date-desc" | "date-asc" | ...;` Then you can define the set of values needed as constants. You should also reuse this in `collectionsSortOptions`, probably something like `const collectionsSortOptions: { [key in sortIds]: string } = {`.

This defines the `collectionsSortOptions` as an object of specific key values and string values (although you _could_ also give the string values a type).

app/src/utils/api.ts
```diff
@@ -165,6 +164,22 @@ export const getDivisionData = async ({
...
+  pageNum = 1,
+  perPage = CARDS_PER_PAGE,
+}: {
+  keyword?: string;
+  sortID?: string;
+  pageNum?: number;
+  perPage?: number; // OPEN QUESTION: perhaps remove the "perPage" field because the API default is 48 and that is what we want.
```
EdwinGuzman, line 176: Yes.... might be a YAGNI but I would personally err on the side of implementing this in case 48 changes to another number.

EdwinGuzman (commented):
I did not review the styling to focus on the functionality.

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: Ah this must have something to do with my misunderstanding of the concept of title "ascending" or "descending."

avertrees: try again?

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: should be updated now

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: do you think we only need to use state for the input field/currentCollectionKeyword? I feel like the other two (sort,pagination) can be managed the way that they are right now.

app/src/utils/api.ts
avertrees: so keep it?

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,111 @@
...
+import NoResultsFound from "../../results/noResultsFound";
 
 export const CollectionsPage = ({ data }) => {
+  const { push } = useRouter();
+  const searchParams = useSearchParams();
+  const pathname = usePathname();
+  const [isLoaded, setIsLoaded] = useState(false);
+  let collections = [];
+
+  if (data.headers.message === "Collections retrieved successfully") {
```
EdwinGuzman, line 33: This can be brittle if the text changes; the frontend doesn't own it and if the backend changes this string, it will cause errors. Is it not possible to check if `data.collections` is null, undefined, or empty?

When there's a Search model/class, there should be a flag for this such as `data.hasCollections`. For now, make this a variable so you're not testing the string in the if statement.

Make sure to write a test for `getCollectionsData` to reflect what the API returns and this message there.

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: I'm going to leave the approach up to you, but I hope this helps:

- Search for "cat"
- click on the "clear" X button in the input
- perform an empty search

The single "cat" result still shows. When you manually remove the text from the input and perform an empty search, it works. You have to somehow keep track of the keyword for this use case.

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Works fine.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -37,20 +141,106 @@ export const CollectionsPage = ({ data }) => {
...
+                    onClick: onMenuClick,
+                    type: "action",
+                  },
+                ]}
+              />
+            </Box>
+            <Heading
+              size="heading5"
+              sx={{ marginBottom: "l" }}
+              ref={headingRef}
```
EdwinGuzman, line 206: By putting the `Menu` and this header in the conditional, when you perform a search or sort, focus gets lost. Focus should go to the heading after performing a search or selecting a sorting option.

app/src/utils/utils.ts
```diff
@@ -118,3 +125,18 @@ export function displayResults(
...
+  Object.keys(paramObj).map((name, value) => {
+    params.set(
+      name.toString(),
+      name === "page" ? paramObj[name].toString() : paramObj[name]
+    );
+  });
+  return params.toString();
+};
+
+// paramObj options:
```
EdwinGuzman, line 142: Instead of listing it down in a comment, add types to define the properties in the `paramObj`.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -37,20 +141,106 @@ export const CollectionsPage = ({ data }) => {
...
+              )}
+            results`}
+            </Heading>
+            <CardsGrid records={collections} />
+          </>
+        ) : (
+          <NoResultsFound searchTerm={currentCollectionKeyword} />
+        )
+      ) : (
+        Array(Math.ceil(collections.length / 4)).fill(
```
EdwinGuzman, line 224: Why do you need this here? Note that `CardsGrid` has it's own set of `LaneLoading` so this here bounces around the number of skeleton loaders on the page.

EdwinGuzman (commented):
Added more comments with the latest updates. 

It's working but some parts are not needed (like the skeleton loader at the top level) and the focus after searching/sorting is now lost.

Consider disabling the input elements after performing any user interaction or some other way to stop a current action.

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: We don't _need_ it. It would introduce a new feature of retrieving new data when a user stops typing. https://nextjs.org/learn/dashboard-app/adding-search-and-pagination#best-practice-debouncing

Just flagging it now but perhaps this is a product decision. We can build it out later.

EdwinGuzman: A few points:

- We should definitely use debouncing where necessary. My question was mostly why do we need/want debouncing with pagination? It's typically used for text inputs so I was curious about that. Perhaps a typo but if not, how does debouncing help with pagination?
- The page does not perform a search as the user types, they must hit "search". For the current set of features, the app doesn't need debouncing. If it's needed in the future, yes, you can build it, and yes, it's a product decision.
- I'd like to get away from having to import various packages for features but `use-debounce` looks small and very well maintained.

avertrees:
making some minor updates to reflect the v7 designs https://www.figma.com/design/LRA06Nqw3z0JEvBbl87DWC/DC-Facelift%3A-Top-Level-Category-Pages?node-id=11735-79867&node-type=canvas&t=Wb6zYPyA28hz3a98-0

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,116 @@
...
+  );
+
+  const [currentSort, setCurrentSort] = useState<string>(
+    params.sort || DEFAULT_COLLECTION_SORT
+  );
+
+  const [currentCollectionKeyword, setCurrentCollectionKeyword] =
+    useState<string>(params.collection_keyword || DEFAULT_SEARCH_TERM);
+
+  const handleLoadingState = async (queryString) => {
```
EdwinGuzman, line 61: It's great you were able to turn this into a reusable function and use it various places. I'm going to be nitpicky about the name and suggest it is updated to something that explains that it's "performing" the search by routing to the new search URL. It is handling the loading state but in the grand scheme of things, it's minor compared to the search that is going on here.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -23,12 +122,22 @@ export const CollectionsPage = ({ data }) => {
     >
       <Box
         sx={{
+          maxWidth: "730px",
```
EdwinGuzman, line 125: Where did you get this number?

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -37,20 +146,111 @@ export const CollectionsPage = ({ data }) => {
           subtitle="Explore the New York Public Library's diverse collections, including digitized photographs, manuscripts, maps, and more. Start exploring by using the search bar below or browse through the collections."
         />
         <SearchBar
-          sx={{ maxWidth: "462px", marginTop: "l" }}
+          sx={{ maxWidth: "462px" }}
```
EdwinGuzman, line 149: Where did you get this number from?

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -37,20 +146,111 @@ export const CollectionsPage = ({ data }) => {
...
+      <Heading
+        size="heading5"
+        sx={
+          collections?.length > 0
+            ? { marginBottom: "l" }
+            : { display: "none", marginBottom: "l" }
+        }
+        ref={headingRef}
+        tabIndex={-1}
+        id={"all-collections-page"}
```
EdwinGuzman, line 220: nit: you dont need braces around these types of strings

app/src/components/results/noResultsFound.tsx
```diff
@@ -0,0 +1,34 @@
...
+  Text,
+  Link,
+  List,
+} from "@nypl/design-system-react-components";
+import Image from "next/image";
+import useBreakpoints from "@/src/hooks/useBreakpoints";
+import { useFeedbackContext } from "@/src/context/FeedbackProvider";
+
+export default function NoResultsFound({ searchTerm }) {
+  console.log("searchTerm is: ", searchTerm);
```
EdwinGuzman, line 13: Remove the console eventually and after you can better log this error. This is a server component, so you should be able to use the logger (for cloudwatch), right?

app/collections/page.tsx
7emansell: Also just want to note here that in the OG DC url query, it's `collection_keywords`, not `collection_keyword`, if we're thinking about URL parity

app/src/config/constants.ts
```diff
@@ -13,3 +13,7 @@ export const ADOBE_EMBED_URL =
 export const TRUNCATED_LENGTH = 80;
 
 export const CARDS_PER_PAGE = 48;
+
+export const DEFAULT_PAGE_NUM = 1;
```
EdwinGuzman, line 17: Not necessary now, but as this app adds more pages and it will have multiple searches, this might be tricky to maintain. For example, you might have `DEFAULT_COLLECTION_SORT`, `DEFAULT_MAIN_SEARCH_SORT`, etc for any page also need as sort. 

Perhaps `DEFAULT_COLLECTION_SORT` can have a more generic name, or there's a file for `/collection` page constant and utils. Again, not necessary now but keep that in mind.

app/src/components/pages/collectionsPage/collectionsPage.tsx
7emansell: I agree, and I also think this should be moved to the page level anyways (where the 404 is checked).

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,116 @@
...
+
+  if (data.headers.message === "Collections retrieved successfully") {
+    collections = Array.isArray(data.collection)
+      ? data.collection
+      : [data.collection];
+  }
+
+  const headingRef = useRef<HTMLHeadingElement>(null);
+
+  // pagination
```
7emansell, line 45: Nitpicky but I actually don't think this comment, or L49, is needed

app/src/components/pages/collectionsPage/collectionsPage.tsx
7emansell: In general, and I'm sure I introduced most of these inconsistencies, we should have the loading state at the same level on every page that's displaying cards. I'm not sure which makes more sense (within the CardsGrid/Lane components or the page component) but we should pick one

app/src/components/results/noResultsFound.tsx
7emansell, line 13: nitpick but just remember to remove this and the unused imports ^^ in this file

7emansell (changes requested):
Looks great so far!!!!!

7emansell:
Also, some searches (like "a") return no results when I really wouldn't expect them to– not sure if that's to do with QA data or what

app/src/utils/utils.ts
```diff
@@ -118,3 +128,31 @@ export function displayResults(
   const end = Math.min(page * perPage, numFound);
   return `${start}-${end} of ${numFound}`;
 }
+
+export const createQueryStringFromObject = (paramObj) => {
+  const params = new URLSearchParams();
+  Object.keys(paramObj).map((name, value) => {
```
EdwinGuzman, line 134: Two points here:

- `.map` functions return a new array. Use `forEach` instead.
- `value` is not being used so it can be removed. You're iterating over the keys and that's a single value, not a pair.

app/src/utils/utils.ts
```diff
@@ -118,3 +128,31 @@ export function displayResults(
...
+      name.toString(),
+      name === "page" ? paramObj[name].toString() : paramObj[name]
+    );
+  });
+  return params.toString();
+};
+
+export const createCollectionsQueryStringFromObject = (paramObj) => {
+  const newParams = {};
+  Object.keys(paramObj).map((key) => {
```
EdwinGuzman, line 145: Same idea here with using `forEach` instead of `map`.

app/src/utils/utils.ts
```diff
@@ -118,3 +128,31 @@ export function displayResults(
...
+    if (
+      !(
+        paramObj[key] === DEFAULT_SEARCH_TERM ||
+        paramObj[key] === DEFAULT_PAGE_NUM ||
+        paramObj[key] === DEFAULT_PAGE_NUM.toString() ||
+        paramObj[key] === DEFAULT_COLLECTION_SORT
+      )
```
EdwinGuzman, line 152: You can simplify and make this more reusable if you change it to 
```
const defaultValues = [
  DEFAULT_SEARCH_TERM,
  DEFAULT_PAGE_NUM,
  DEFAULT_PAGE_NUM.toString(),
  DEFAULT_COLLECTION_SORT
];
if (!defaultValues.includes(paramObj[key])) {
  newParams[key] = paramObj[key];
}
```

Then you can reuse the function if `defaultValues` is passed as an argument for other cases.

app/src/utils/utils.ts
```diff
@@ -118,3 +128,31 @@ export function displayResults(
   const end = Math.min(page * perPage, numFound);
   return `${start}-${end} of ${numFound}`;
 }
+
+export const createQueryStringFromObject = (paramObj) => {
+  const params = new URLSearchParams();
+  Object.keys(paramObj).map((name, value) => {
+    params.set(
+      name.toString(),
+      name === "page" ? paramObj[name].toString() : paramObj[name]
```
EdwinGuzman, line 137: This specific check, along with `DEFAULT_PAGE_NUM.toString()`, are minor code flags. I know you're making `DEFAULT_PAGE_NUM` a number since that's what the DS uses, but for consistency and to make these functions easier, a string is a better choice. Then in `collectionPage.tsx`, you can use it as `Number(DEFAULT_PAGE_NUM)`. Then it's only changed once and these functions will be easier to maintain.

EdwinGuzman (commented):
More comments but good progress again.

The biggest thing I'd like to see after the comments are addressed, are tests for `colllectionsPage.tsx`. There should be a baseline of tests for user interactions and verifying you get back results or the `noResultsFound` component.

EdwinGuzman:
Two more things I forgot to bring up:
- The API endpoint is pretty fast and I think you can bring down the timeout to less than 2 seconds. There's a bit of jumpy-ness after performing a search, where the focus goes, and when the results render.
- The "No Results" text takes a bit longer than expected to show up, and it could use a loader before it appears. However, this was essentially scope creep into the first ticket you were working on. So this can be an improvement that can be done afterward.

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: not a nit, this is evidence of bad practice

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: https://github.com/NYPL/digital-collections/pull/180/files#diff-db9636018921e1dbb9ce8c307175c049f72c22d0add5d992ef77ef35a2ca68c1R38 

Screenshot 2024-12-10 at 1 51 22 PM (2)

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: I used the width from the description on the division landing page: 
https://github.com/NYPL/digital-collections/pull/193/files#diff-78d2323f3e142cea5e900f6080991f164b2ecea631e8d0b5e9b64c1490f9f7dfR41

avertrees:
> Also, some searches (like "a") return no results when I really wouldn't expect them to– not sure if that's to do with QA data or what

@7emansell really wouldn't or would expect them to?

Lana and Kris have been battling out whether or not keyword search on title should have more results or not for a few weeks.  There's not much we can do here, it's a solr query design decision. https://api.repo.nypl.org/api/v2/collections?q=a

7emansell:
> > Also, some searches (like "a") return no results when I really wouldn't expect them to– not sure if that's to do with QA data or what
> 
> @7emansell really wouldn't or would expect them to?
> 
> Lana and Kris have been battling out whether or not keyword search on title should have more results or not for a few weeks. There's not much we can do here, it's a solr query design decision. https://api.repo.nypl.org/api/v2/collections?q=a

I would expect them to since https://digitalcollections.nypl.org/collections?utf8=%E2%9C%93&collection_keywords=a returns so many results... But I see that that's not something that'll be fixed here

avertrees:
@7emansell yeah I noticed that earlier and Lana and Kris are aware. We have the thumbs up for now. If we want to change it, it has to be updated in the new repo.

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Yes, that's fair. You may need to do 
```
sx={{
  "> input": { maxWidth: "462px" }
}}
```
This is setting the max width to the entire component: 
Screenshot 2024-12-11 at 9 13 19 AM

app/collections/page.tsx
EdwinGuzman: If we can build out a redirect service, I'd like `collection_keywords` to be `q`. That's the conventional query key search (for example https://www.google.com/search?q=cats).

avertrees: good find @7emansell thank you.

@EdwinGuzman I would like that as well. However, since we do not have a redirect service atm, `collection_keywords` will be the expected param.

app/src/utils/api.ts
avertrees: keeping it for now

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: fine

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: actually, to get the `CardsGrid` loading state to work the way we want it to work, we'll have to pass down the value of "isLoading" as a prop. I think this will require us to update several other paginated pages. 

Due to the existing size of this PR, I don't really want to touch more pages and add MORE features on additional pages to test in the mix. I think we should have a larger conversation about our loading logic and clean this up later.

__tests__/pages/collectionspage.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { mockCollectionsResponse } from "__tests__/__mocks__/data/mockApiResponses";
+import { CollectionsPage } from "@/src/components/pages/collectionsPage/collectionsPage";
+import CollectionSearchParams from "@/src/types/CollectionSearchParams";
+
+jest.mock("next/navigation", () => ({
+  useRouter: jest.fn(),
+  usePathname: jest.fn(),
+}));
+
+describe.skip("Homepage Accessibility", () => {
```
EdwinGuzman, line 13: Nit: update the title to CollectionsPage and remove the skip. You may need to mock the `push` function from `useRouter` for the test to run.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+    setTimeout(() => {
+      setIsLoaded(true);
+      headingRef.current?.focus();
+    }, 1000);
+  };
+
+  const handleSearchSubmit = async (e: SyntheticEvent) => {
+    e.preventDefault();
+    // something is so weird about state here.
+    // change the values in the object passed down to createCollectionsQueryStringFromObject to be currentPage and currentSort and tell me if it works for you....
```
EdwinGuzman, line 71: Ah yeah, that's because the `set...` functions are asynchronous while the `createCollectionsQueryStringFromObject` is not. So this is updating the current values for page and sort, not waiting until that's done, and calling the function with the page and sort values. You might be seeing stale values _before_ they get updated from the async calls.

One option is to put these type of side effects inside the set function as
```
let queryString = "";
setCurrentPage((_prevValue) => {
  queryString = createCollectionsQueryStringFromHash({
      collection_keywords: currentCollectionKeywords,
      sort: DEFAULT_COLLECTION_SORT,
      page: DEFAULT_PAGE_NUM,
    });
  return Number(DEFAULT_PAGE_NUM);
}
```

This should work but it's only referencing the page value. With TWO update functions, this gets trickier. You can try doing a double nested callback function but, as you can see from the example above, it'll get messy...

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,17 +1,113 @@
...
+    const target = e.target as HTMLInputElement;
+    setcurrentCollectionKeywords(target.value);
+  };
+
+  const onPageChange = async (pageNumber: number) => {
+    setCurrentPage(pageNumber);
+    const queryString = createCollectionsQueryStringFromHash({
+      collection_keywords: currentCollectionKeywords,
+      sort: currentSort,
+      page: pageNumber.toString(),
```
EdwinGuzman, line 92: Similar comment above. I'm a bit surprised you're not encountering async issues here where it's possible that `pageNumber` is the previous value before `setCurrentPage(pageNumber);` finishes running.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -37,20 +143,111 @@ export const CollectionsPage = ({ data }) => {
...
         />
       </Box>
       <HorizontalRule sx={{ marginTop: "xxl", marginBottom: "xxl" }} />
-      <SearchResults showFilter isSearchPage={false} data={data} />
+      <Box
+        sx={
+          collections?.length > 0
+            ? { display: "flex", gap: "xs", marginBottom: "l" }
+            : { display: "none", gap: "xs", marginBottom: "l" }
+        }
```
EdwinGuzman, line 169: Rewrite this so it ends up with `display: collections?.length > 0 ?  "flex" : "none",` and the other properties are always there (since they're the same in both conditions).

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Sounds good.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -37,20 +143,111 @@ export const CollectionsPage = ({ data }) => {
...
+      <Heading
+        size="heading5"
+        sx={
+          collections?.length > 0
+            ? { marginBottom: "l" }
+            : { display: "none", marginBottom: "l" }
+        }
+        ref={headingRef}
+        tabIndex={-1}
+        id="collections"
```
EdwinGuzman, line 217: I'm seeing for searches or sorting updates with many results that the page jumps up and the header blocks the content:

Screenshot 2024-12-13 at 12 35 15 PM

Try replicating with a blank search and then updating the sort value. This is a problem but one you can either tackle later or remove the `#collections` from the URL. I think you can remove it -- it might be redundant since you are programmatically focusing on the text after user actions.

EdwinGuzman (commented):
Good fixes and I left minor comments.

I think you are seeing async issues for the page, sort, and keyword search. I'm inclined to say we should tackle them separately, but let me know what you think.

Here's one example:

- Search for "photo", you get four results,
- Update the sort to "Title Z to A", you get 7 results,
- Update the sort to "Title A to Z", you get 7 results,

This first part might be a data issue. Can you confirm is that's true? Is the API returning a different set of results when the sort changes? That seems incorrect.

Continue with the above example:

- Search for "art", you get 117 results,
- The sort is still set to "Title A to Z" but the results are not sorted,
- If you click on the same sort value, they do get sorted.

For this second issue, I think you should reset the sort (both the value and the display) to the default after _every_ new search. What gets tricky about this is that it's a Product decision. You can make this a separate ticket to fix after getting Product input.

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: lana and lily asked for the #collections to be added to the url so the pagination behavior is consistent with the divisions landing page.

EdwinGuzman: Good point. In this case, it's not great user experience for the header to block the page in this manner. You can't see the searchbar so I think the sort menu should not be visible at all rather than partially covered. I think you can manage to push the page around when the element is focused like this and it might already be in the header for this app. We can ask Emma next week.

avertrees: It's not great user experience for now but I don't believe the header is going to block the menu after we move the menu to the right side https://newyorkpubliclibrary.atlassian.net/browse/DR-3323.

avertrees: I think we should wait until DR-3323 is done first.

7emansell: this is an interesting question because to me, intuitively, on clicking "Search" the page shouldn't pull you to the "Displaying results", it should just keep you where you are scroll-wise, but on pagination, it should. Just me ?
The scroll padding on the header is very statically set right now (https://github.com/NYPL/digital-collections/blob/c34939a29a15c1e88f11d8e799f5d37a5afde2e4/app/globals.css) and doesn't have anything to do with where focus is, only with how tall the header is, so a more dynamic approach would be needed if we want to conditionally add padding to where the header ends up when using anchor links

__tests__/pages/collectionspage.test.tsx
avertrees: will deal with this in another PR

avertrees:
found that the bug @EdwinGuzman mentioned above is a result of stale cached api data. will create a ticket to address this since the implementation will touch every page.

will also create another ticket for changing "collection_keywords" to "q" in the url, since that will break old urls.

EdwinGuzman (approved):

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: I don't think we'll have to conditionally add padding after the ticket mentioned above is done, which is a feature that will block the release of this page but can be worked on separately

7emansell: Regardless of whether the header blocks the menu, the jump down the page on Search is a weird experience– is there a way to focus on the "Displaying results" without scrolling down the page, unless "Displaying results" isn't already visible?

__tests__/pages/collectionspage.test.tsx
7emansell: ** i think this is in scope of this PR but if you want to move it out can you link a ticket

7emansell (changes requested):
as mentioned elsewhere (and an issue that started long before this ticket so totally not on you to resolve)– the loading is really jumpy. bc of that, I don't think this page is ready for prod, so the other tickets you mentioned creating (dealing with stale data) should be blockers to releasing this . Hopefully those improve the loading experience. Changes requested are just fixing the a11y test, checking on the jump to "Displaying results" after search (like can this be prevented), and linking the blocker tickets here

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: The behavior on Search is a good point. I'm going to check in with Lana and Lily about this because it is technically possible to _not_ scroll down to the "Displaying results" section on Search by not add `#collections` to the url after search is hit. Not the most fun thing to implement but it is feasible.

7emansell: Okay yeah if I'm the only person that feels this way then dw about it but it does just seem unintuitive to me that the screen scrolls even though the element is already visible

avertrees: Lily and Lana are on board with removing it from Search - I also think it's weird adding it Sort so I removed it from both lol

__tests__/pages/collectionspage.test.tsx
avertrees: That code ^ is not working for me for some reason. I was going to spend more time on this when working on https://newyorkpubliclibrary.atlassian.net/browse/DR-3323

7emansell: this does not run ? 

```
import { render } from "@testing-library/react";
import { axe } from "jest-axe";
import React from "react";
import { mockCollectionsResponse } from "__tests__/__mocks__/data/mockApiResponses";
import { CollectionsPage } from "@/src/components/pages/collectionsPage/collectionsPage";
import { useRouter } from "next/navigation";

jest.mock("next/navigation", () => ({
  useRouter: jest.fn(),
  usePathname: jest.fn(),
}));

beforeEach(() => {
  (useRouter as jest.Mock).mockImplementation(() => ({
    pathname: "/collections",
  }))
});

describe("Collections page Accessibility", () => {
  const searchParams = {
    collection_keywords: "flower",
    sort: "title-asc",
    page: "2",
  };
  it("passes axe accessibility test", async () => {
    const { container } = render(
      <CollectionsPage
        data={mockCollectionsResponse}
        params={searchParams}
        renderCollections={true}
      />
    );
    expect(await axe(container)).toHaveNoViolations();
  }, 60000);
});

```
Nbd but because this test exposes a couple other issues that you should resolve in this PR, i think it's worth getting it going

app/src/components/pages/collectionsPage/collectionsPage.tsx
avertrees: kk added it back to sort as per lanas request and moved sort to the right so the menu isn't blocked off by the header.

__tests__/pages/collectionspage.test.tsx
avertrees: This runs, I was putting the `beforeEach(()` on the wrong line.

PR #258, opened 2024-12-02
No ref/Search endpoint

avertrees:
Reflecting what I think makes sense given the product requirements listed in: 

- the designs https://www.figma.com/design/s78ZbQTw6kG5fqUKOVV2zy/DC-Facelift%3A-Search-Results?node-id=12934-146779&p=f&t=cigeAmTsWaiV5dKC-0
- the BRD https://docs.google.com/document/d/1MHVFDTAOmDFo9S53SH5IM4zAhVNu9RcUSeGWvS5FZuQ/edit?tab=t.0
- this thread: https://nypl.slack.com/archives/C05LCMUT1EV/p1733431434371739 
- and the initial schema Kris made for search, documented here: https://newyorkpubliclibrary.atlassian.net/browse/DR-3321

Search Results Card object:

```
      {
        uuid: "string",
        recordType: "sub-collection | item",
        title: "string",
        imageID: "string | null",
        numberOfDigitizedItems: "number",
        containsOnSiteMaterial: "boolean",
        containsAVMaterial: "boolean", 
        containsMultipleCaptures: "boolean", 
        contentType: "image | audio | video | pdf | null",
        highlights: "{ highlighted_field_name:[ string ] }",
        firstIndexed_dt: "date",
      },
```

field and query parameter docs:
https://docs.google.com/document/d/1_HDSzW5ToWf6ariAm7G0VWoUTvPBCqbq7q-9gInqQiM/edit?tab=t.0#heading=h.11uw6iuzw3ql


EdwinGuzman (changes requested):
I think the endpoints should be grouped as `app/api/schema/...`. Makes it easier to differentiate whether it's an endpoint for the app to use or for schema collaboration purposes. It might also make it easier to write the regex in the reverse proxy repo as a catch-all for all these schema endpoints.

app/api/search/route.tsx
```diff
@@ -0,0 +1,8 @@
+import { NextResponse } from "next/server";
+import searchSchema from "../../src/data/schemas/searchSchema";
+
+export const GET = async () => {
+  return NextResponse.json(searchSchema, { status: 200 });
```
7emansell, line 5: I'm not sure I understand how `status` is working in this setup. Does this mean it would live on the response (like we've been talking about) instead of within the `headers` field

avertrees: that was the idea, yes

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,77 @@
...
+      status: "200",
+      code: "200",
+      message: "some copy",
+    },
+    // Pagination fields
+    numResults: "6",
+    page: "1",
+    perPage: "40",
+    sort: "title DESC",
+    collections: [
```
avertrees, line 16: questions I have following the slack thread linked in the description from 12/5/2024...

if we're expecting collections, containers, and items to be returned in the paginated list of cards, should the array of objects be named "results" and include all three?

in other words, do we need to separate the objects into separate arrays by type?

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,77 @@
...
+      },
+    ],
+    items: [
+      {
+        uuid: "string",
+        title: "string",
+        url: "string",
+        imageID: "string | null",
+        imageURL: "string",
+        containsMultipleImages: "boolean",
```
avertrees, line 55: sounds like we want a field for "on site only" for items

avertrees: as well as a label for the type of media presented in the item

app/src/data/schemas/searchSchema.ts
avertrees: ie. results: [{}]
where each object has a field like "typeOfRecord" that is either "collection" "sub-collection" or "item"
copy is not final here but you get the idea

app/src/data/schemas/collectionSchema.ts
```diff
@@ -0,0 +1,14 @@
+const collectionObject = {
```
EdwinGuzman, line 1: Where is this being used? Or is it for reference?

app/src/data/schemas/collectionsLandingPageSchema.ts
```diff
@@ -0,0 +1,41 @@
+const collectionsLandingPageSchema = {
+  response: {
+    numResults: "6",
+    page: "1",
+    perPage: "40",
```
EdwinGuzman, line 5: For consistency, use "string" instead of specific values.

app/src/data/schemas/collectionsSchema.ts
```diff
@@ -0,0 +1,48 @@
...
+    // move from "collection" to "collections"
+    collections: [
+      {
+        title: "string",
+        uuid: "string",
+        url: "string",
+        imageID: "string",
+        numberOfDigitizedItems: "string", // remove numItems
+        containsAVMaterial: "boolean",
+        constainsOnSiteMaterials: "boolean",
```
EdwinGuzman, line 15: These will always be either `true` or `false` right? So a `null` or empty value won't be the same as `false`?

app/src/data/schemas/itemObject.ts
```diff
@@ -0,0 +1,14 @@
+const itemObject = {
```
EdwinGuzman, line 1: This is also not being used yet.

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,64 @@
...
+        recordType: "collection | sub-collection | item",
+        title: "string",
+        url: "string",
+        imageID: "string | null",
+        imageURL: "string",
+        numberOfDigitizedItems: "number",
+        containsOnSiteMaterials: "boolean | null if type is `item`",
+        contentType: "image | multiple images | audio | video | pdf | null",
+        onSiteOnly:
+          "boolean | null if type is `collection` or `sub-collection`",
```
EdwinGuzman, line 26: We briefly talked about this but it might be easier to make it more typescript-y and give each object "collection", "sub-collection", and "item" a type. Then it's could be 
```
results: [
  {
    uuid: ...
    recordType: "collection"
    ....
  },
  {
    uuid: ...
    recordType: "sub-collection"
    ....
  },
  {
    uuid: ...
    recordType: "item"
    ....
  }
]
```

app/src/data/schemas/collectionsLandingPageSchema.ts
7emansell: ^^

app/src/data/schemas/collectionsLandingPageSchema.ts
```diff
@@ -0,0 +1,41 @@
...
+    items: [
+      {
+        uuid: "string",
+        recordType: "item",
+        title: "string",
+        url: "string",
+        imageID: "string | null",
+        imageURL: "string",
+        contentType: "image | multiple images | audio | video | pdf",
```
7emansell, line 14: Is there a way for this to express that these are the 5 strings we're interested in, but that we expect (I think) that there are actually more content types returned on the item. So we just want a string here, then on the model we would transform it as needed?

app/src/data/schemas/collectionsLandingPageSchema.ts
```diff
@@ -0,0 +1,41 @@
...
+        onSiteOnly: "boolean",
```
7emansell, line 15: would an item have an 'isAVMaterial' field

7emansell: Or would that be covered by `contentType`

app/src/data/schemas/collectionsSchema.ts
```diff
@@ -0,0 +1,48 @@
...
+      },
+      {
+        title: "string",
+        uuid: "string",
+        url: "string",
+        imageID: "string",
+        numberOfDigitizedItems: "string",
+        containsAVMaterial: "boolean",
+        constainsOnSiteMaterials: "boolean",
```
7emansell, line 24: nit but contains, not constains. Just in case this is what gets copy pasted into new api

app/src/data/schemas/itemObject.ts
7emansell: I don't think either of these will be "used" until they exist in new repo api but the idea was have an expected shape to base models off of

app/src/data/schemas/itemObject.ts
```diff
@@ -0,0 +1,14 @@
...
+  uuid: "string",
+  recordType: "item",
+  title: "string",
+  url: "string",
+  imageID: "string | null",
+  imageURL: "string",
+  contentType: "image | multiple images | audio | video | pdf",
+  onSiteOnly: "boolean",
+  highlights: "{ highlighted_field_name:[ string ] }",
```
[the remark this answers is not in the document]
7emansell: I'm also a little confused by this one. Can you give an example of what this might return as

app/src/data/schemas/collectionsLandingPageSchema.ts
7emansell: But `page`/`perPage` should either always be a number or a string across all of these

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,64 @@
+const searchSchema = {
+  response: {
+    // Keyword included in search
+    keyword: "string",
+    // Pagination fields
+    numResults: "integer",
+    page: "integer",
+    perPage: "integer",
```
7emansell, line 8: reiterating comment from collectionsSchema: `page`/`perPage` should either always be a number or a string across all of these

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,64 @@
...
+    // Sort fields
+    sort: "string ie: title DESC",
+    // Filter fields
+    filterField: "string",
```
7emansell, line 12: Is this an array of strings? Can you give an example of what this might return as

app/src/data/schemas/searchSchema.ts
7emansell: I like this idea. In general, this schema makes it really clear what the search results row card will have available to it

7emansell (changes requested):
Looks great so far and this is so helpful for hashing out exactly what we could expect/ what we want to expect on the front end

app/src/data/schemas/collectionSchema.ts
avertrees: just for reference

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,64 @@
...
+    // Results
+    results: [
+      {
```
avertrees, line 15: after reading the requirements for the modal in the brd, I think we can include the list of filter categories and the first 10 filters per category in this endpoint. each category would have a field for the total count of filters so the app knows when to display the "View all" button that loads the modal.

app/api/schema/search/facets/route.tsx
```diff
@@ -0,0 +1,8 @@
+import { NextResponse } from "next/server";
+import searchFacetsSchema from "../../../../src/data/schemas/searchFacetsSchema";
+
+export const GET = async () => {
```
avertrees, line 4: this is the intended api to use for the modal. 

idk if `facets` is the right term to use here? `filters` is kind of more implicit of what is actually being searched. 

regardless, interacting with the api to get the list of the filters in the modal would require a keyword search ie. `q=` and a way to identify the _category_ we are searching, ie. `category=`.

app/src/data/schemas/collectionsLandingPageSchema.ts
avertrees: that would be covered by `contentType`

app/src/data/schemas/collectionsLandingPageSchema.ts
avertrees: in the schema? I'm not sure. might have to be something we explain when we pass it off.

I keep asking myself "why can't we use the `type_of_resource` field?" but I think the answer is that we expect one of these 5 to be returned for every item and if we return the resources included in the `type_of_resource` and filter out these 5 on the front end, we might not have 1:1 parity

app/src/data/schemas/itemObject.ts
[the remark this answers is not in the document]
avertrees: I took this from the ticket Kris worked on for the initial search functionality in the new api. I'll see what info I can get from him because this field also confuses me.

app/src/data/schemas/itemObject.ts
avertrees: I can remove these but yeah it's just to differentiate what fields are specific to an item vs. a collection/sub-collection.

app/src/data/schemas/collectionsSchema.ts
avertrees: good call, "false" is sufficient

app/src/data/schemas/collectionsLandingPageSchema.ts
7emansell: Ah okay. this might be a product question but I was thinking there's a case where none of those 5 strings are returned, so on the frontend we simply don't display a pill tag for the content type

app/src/data/schemas/collectionsLandingPageSchema.ts
avertrees: @7emansell @EdwinGuzman do either of you have a preference? right now all api endpoints return a stringified integer for their paginated values. meaning, if we ask for all new endpoints to return an integer, we'd need to check the type of these values on the frontend while we migrate the pre-existing endpoints over to the new api.

app/src/data/schemas/searchSchema.ts
EdwinGuzman: I'm not sure about this and I want you to talk to the backend team about this. This is verging on writing and getting data from the API _just_ for the frontend. What if in the future, the modal approach does not work and the frontend needs to display 5 or 20 filters? I'm not disagreeing with this approach but just want acknowledgment from everyone that this is the right way to do this.

What I find helpful from APIs is a "self" link to more resources. So, besides the total count of filters property, there can be a property for "self" that is another endpoint to get the rest of the filters. Otherwise, how would you know where to fetch the filters needed? Just an idea.

app/api/schema/search/facets/route.tsx
EdwinGuzman: If `facets` and `filters` are interchangeable, then the exact name doesn't matter too much, as long as we all know what it _should_ be called when talking about the options in the modal.

app/src/data/schemas/searchSchema.ts
avertrees: this would be on par with what the research catalog does https://www.nypl.org/research/research-catalog/api/search?q=cats

app/src/data/schemas/itemObject.ts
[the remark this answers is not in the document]
avertrees: Some examples from Kris:

```
"highlights": {
        "mainTitle_st": [
          "Christopher <em>Walken</em> in the stage production Kid Champion"
        ]
      },
```

```
"highlights": {
        "mainTitle_st": [
          "Beatrice Lillie, Ronnie <em>Walken</em> aka Christopher <em>Walken</em> (far right), and company in the stage"
        ]
      },
```

```
      "highlights": {
        "mainTitle_st": [
          "Beatrice Lillie, Ronnie <em>Walken</em> aka Christopher <em>Walken</em> (far right), and company in the stage"
        ]
      },
```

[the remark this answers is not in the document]
avertrees: That's what I understand from how Kris wrote the initial endpoint. 

I can talk to Lana and Kris about whether we ever expect there to ever be more than one highlight field.

app/src/data/schemas/collectionsLandingPageSchema.ts
avertrees: Oh, technically none of those 5 options would ever be displayed for a `collection` or `sub-collection`. But Lana confirmed that we do expect one of these to be displayed for every `item`.

app/src/data/schemas/collectionsLandingPageSchema.ts
avertrees: Update: in the new Collections api endpoint, Kris is returning an integer for the paginated fields.

Question I have: if we decide we want integers returned here, what's the best way to indicate that that we expect an integer when the word "integer" itself is a string?

app/src/data/schemas/searchSchema.ts
avertrees: sorry, want to make sure I understand correctly:

are you saying that 

`recordType: "collection | sub-collection | item",`

is _not_ clear and that it would be more clear if each object in the array had a different type?

ie. 
```
results: [
  {
    uuid: ...
    recordType: "collection"
    ....
  },
  {
    uuid: ...
    recordType: "sub-collection"
    ....
  },
  {
    uuid: ...
    recordType: "item"
    ....
  }
]
```

app/src/data/schemas/collectionsLandingPageSchema.ts
EdwinGuzman: This might be excessive, but with the schemas there could also be endpoints that return examples.

EdwinGuzman: To answer your question, I don't have any preference.

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,64 @@
+const searchSchema = {
+  response: {
```
avertrees, line 2: @emu47

avertrees: @chrisb87

app/src/data/schemas/collectionsUUIDSchema.ts
```diff
@@ -0,0 +1,7 @@
+const collectionsUUIDSchema = {
+  response: {
+    // TODO: Awaiting product requirements
```
EdwinGuzman, line 3: Can you return the comment in the response? That way, when you visit `/api/schema/collections/1234`, it's clear that it's not waiting reqs and not broken. 

`"TODO:": "Awaiting...."` is enough

EdwinGuzman:
Approving the schemas now and acknowledging this will go through many more iterations and updates.

app/src/data/schemas/searchSchema.ts
```diff
@@ -0,0 +1,169 @@
...
+    keyword: "string",
+    // Pagination fields
+    numResults: "integer",
+    page: "integer",
+    perPage: "integer",
+    // Sort fields
+    sort: "string ie: title DESC",
+    // Filter fields
+    // facets are an array of objects past in the request body
+    filterByPublicDomain: "boolean",
```
avertrees, line 13: booleanFilter: "publicDomain" or "availableOnline" or "onSiteMaterial" or ""

avertrees:
merging this in

PR #263, opened 2024-12-18
update label text on sort menu for all collections page

avertrees:
## Ticket:

- JIRA ticket [include "sort by" label inside the drop down](https://newyorkpubliclibrary.atlassian.net/browse/DR-3323)

-

## Open Questions

-

-

### Checklist:

### DIFF
--- a/app/src/components/pages/collectionsPage/collectionsPage.tsx
+++ b/app/src/components/pages/collectionsPage/collectionsPage.tsx
@@ -161,7 +161,6 @@ export function CollectionsPage({ data, params, renderCollections }) {
           }}
           onSubmit={handleSearchSubmit}
           labelText="Search Collections by title"
-          aria-label="Search Collections by title"
         />
       </Box>
       <HorizontalRule sx={{ marginTop: "xxl", marginBottom: "xxl" }} />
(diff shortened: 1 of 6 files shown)


app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -159,11 +160,12 @@ export function CollectionsPage({ data, params, renderCollections }) {
             onChange: (e) => handleSearchChange(e),
           }}
           onSubmit={handleSearchSubmit}
-          labelText={""}
+          labelText="Search Collections by title"
+          aria-label="Search Collections by title"
```
7emansell, line 164: major nit but should collections be capitalized

7emansell (approved):
only the 1 nit

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Hmm good point. I don't think you need the aria-label, the `labelText` should cover that.

EdwinGuzman (commented):
Looks good but can you verify that the aria-label is not needed?  The labelText should add the value in the form element which is enough.

avertrees:
@EdwinGuzman the aria-label was added for the a11y test and it did not fix the issue. I can remove it for now and the DS docs do say that the labelText should be enough to populate the aria label. I made another ticket for the a11y tests: [DR-3342 FE | Write a11y tests for all phase 2 pages](https://newyorkpubliclibrary.atlassian.net/browse/DR-3342?atlOrigin=eyJpIjoiMjA0ZTM3Nzk4MTc5NDE3YjkwYWQxYzYzY2ZjNDQ1ODQiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

EdwinGuzman (approved):

PR #267, opened 2025-01-07
vqa fixes for no results page

avertrees:
following lily's VQA feedback on the "no results" view of the All Collections page. she approved and this just needs to go off to QA.

### DIFF
--- a/app/src/components/results/noResultsFound.tsx
+++ b/app/src/components/results/noResultsFound.tsx
@@ -4,13 +4,13 @@ import {
   Text,
   List,
 } from "@nypl/design-system-react-components";
+import styles from "../../styles/Collections.module.css";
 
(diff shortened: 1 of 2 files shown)


7emansell (commented):
looks good, just wondering, should we have a loading state? even maybe just one skeleton loading lane

EdwinGuzman (approved):
pending Emma's question

avertrees:
@7emansell @EdwinGuzman I think yes, we should have a loading state. After looking into how to implement the loading state for the no results view, it requires some reconfiguration of the logic. I think it's best if we make the loading state for this component part of the A/C for https://newyorkpubliclibrary.atlassian.net/browse/DR-3330

7emansell:
sounds good

7emansell (approved):

PR #270, opened 2025-01-09
DR-3331: Search spike (clientside hook to get URL)

7emansell:
POC so definitely needs some refinement (see comments)
[TAD (sort of) to describe this search approach](https://docs.google.com/document/d/1BEIhffdr8_iMK2-hdwd6PFX4_SjR6fGyoYNXhPmyFXw/edit?usp=sharing)

The general structure I'm going for:  
- the `SearchProvider` initializes the `SearchManager` with either default values or values from the current URL
- components can use `useSearchContext` to access the `SearchManager` and call its methods to get a new query string (which updates search state), but they're responsible for then actually updating the URL
- when the URL changes, the `useEffect` in `SearchProvider` updates the `SearchManager` state (commented out)
- OR just `useSearchParams` triggers a rerender and creates a new instance of the `SearchManager` with the new search params
- and, vice versa, `SearchManager` methods generate query strings for updating the URL

## Open Questions
 See comments, and additionally:
- Is it worth adding (or kind of replacing parts of `SearchManager` with) `useReducer`? It would make it more React-y but, imo, not necessarily clearer what's going on


app/search/index/page.tsx
```diff
@@ -1,28 +1,49 @@
...
-  searchParams: { page: string };
+  searchParams: SearchParams;
 };
 
-export default async function Search() {
+export default async function Search({ searchParams }: SearchProps) {
+  const pageName = !searchParams.keywords ? "all-items" : "search-results";
+
+  const data = await getSearchData({
+    keywords: searchParams.keywords || "type_s:Item",
```
7emansell, line 31: this "type_s:Item" is for Repo API to just return all items. When building this for real, we need to return all items and collections

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,112 +1,65 @@
...
-    Number(params.page) || Number(DEFAULT_PAGE_NUM)
-  );
-
-  const [currentSort, setCurrentSort] = useState<string>(
-    params.sort || DEFAULT_COLLECTION_SORT
-  );
-
-  const [currentCollectionKeywords, setcurrentCollectionKeywords] =
-    useState<string>(params.collection_keywords || DEFAULT_SEARCH_TERM);
+  const collectionSearchManager = new SearchManager({
```
7emansell, line 47: This search has its own instance of SearchManager, as will the search within a single collection

app/src/utils/searchManager.tsx
```diff
@@ -0,0 +1,230 @@
...
+  filter: string;
+  value: string;
+};
+
+export class SearchManager {
+  currentPage: number;
+  currentSort: string;
+  currentKeywords: string;
+  currentFilters: Filter[];
+  isCollectionSearch: boolean;
```
7emansell, line 26: this flag for collections search and generally how this class is split between two sets of parameters  is not ideal... But works for now

app/src/context/SearchContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return filters.map(({ filter, value }) => `[${filter}=${value}]`).join(",");
+};
+
+const SearchContext = createContext<SearchContextType | undefined>(undefined);
+
+export const SearchProvider = ({ children }: { children: React.ReactNode }) => {
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+
+  // const searchManager = useMemo(() => {
```
7emansell, line 51: this works but is wasteful re. creating a bunch of new SearchManagers

app/search/index/page.tsx
7emansell: Which prompts the question of what will be the "All items" page? Since we are planning to have search return collections and items all together

app/src/components/search/search.tsx
```diff
@@ -48,16 +38,34 @@ const Search = () => {
           },
         }}
       >
+        {/*  <form
```
7emansell, line 41: when this wraps search, it allows it to work with no JS, but since we have skeleton loaders based on `useEffect` all the way down to the Card level, hard to tell

app/src/utils/searchManager.tsx
charmingduchess: I think this might use some Replace Conditional with Polymorphism a la Refactoring.

charmingduchess: You could hard code two separate classes, CollectionSearchManager and GeneralSearchManager. However, it seems like the difference between the collections queries and other queries are the params being returned. Another approach is to have a GenericSearchManager that could accept parameters like `handlePageChangeParams`, and factory methods for Collection and General SearchManagers would provide those values for the different classes.

app/src/utils/searchManager.tsx
```diff
@@ -0,0 +1,220 @@
...
+          keywords: this.currentKeywords,
+          sort: this.currentSort,
+          page: pageNumber,
+          filters: filterToString(this.currentFilters),
+        });
+
+    return queryString;
+  }
+
+  handleSortChange(id: string) {
```
charmingduchess, line 116: if all of these `handle*` methods are returning query strings, maybe the name can reflect that?

app/src/components/pages/collectionsPage/collectionsPage.tsx
charmingduchess: might this be wanting to be a useRef? I think otherwise it will generate a new object on every rerender!

app/src/utils/apiHelpers.ts
```diff
@@ -8,8 +8,11 @@ import {
   DEFAULT_PAGE_NUM,
   DEFAULT_SEARCH_TERM,
   COLLECTION_SORT_OPTIONS,
+  DEFAULT_SORT,
+  DEFAULT_FILTERS,
 } from "../config/constants";
 import { fetchApi } from "./fetchApi";
+import { Filter } from "./searchManager";
```
EdwinGuzman, line 15: import this as a type by `import type { Filter } ...`. This makes it easier to read and know that it's a type and also helps the compiler a bit by giving it a hint of what's imported.

app/src/utils/searchManager.tsx
7emansell: open to ideas here bc I want to express 1. it's doing something to state 2. AND returning a query string

app/search/index/page.tsx
avertrees: Great question. I don't think product has a plan for an "All items" page right now given that in OG DC, the /search page stood in for the "All Items" page.

app/src/context/SearchContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+    new SearchManager({
+      initialPage: DEFAULT_PAGE_NUM,
+      initialSort: DEFAULT_SORT,
+      initialFilters: [],
+      initialKeywords: DEFAULT_SEARCH_TERM,
+      isCollectionSearch: false,
+    })
+  );
+
+  useEffect(() => {
```
EdwinGuzman, line 73: I don't necessarily agree with this. I removed `useEffect`s in this PR in Turbine https://github.com/NYPL/nypl-ds-test-app/pull/122/files#diff-3af8a06a249633ad6569f63b96449a1b73fe6e3e58dca0396e748ae36234443e

What you have going on this is workflow is great and try to make it work without this `useEffect`.

EdwinGuzman (commented):
We should go over some comments together. Overall, the biggest things are
- break up the class into multiple classes rather than using the `isCollectionSearch` property
- try to remove the `useEffect` -> The underlying question is, is this approach meant to use the URL query params as the source of truth or state for searching? If that's so, the data can flow top down so it should be possible to remove the `useEffect`. You may (or will) need to set local state for _some_ ds components like checkboxes or the multiselect.

app/src/context/SearchContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+  DEFAULT_SORT,
+  DEFAULT_SEARCH_TERM,
+} from "../config/constants";
+import { Filter, SearchManager } from "../utils/searchManager";
+
+interface SearchContextType {
+  searchManager: SearchManager;
+}
+
+export const stringToFilter = (filtersString: string | null): Filter[] => {
```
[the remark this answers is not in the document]
dgcohen: Correct, you can use the query attribute in getServerSideProps which separates the query string values. If you'll need to do this parsing on the client side I'd recommend using a library like qs.

app/src/utils/searchManager.tsx
dgcohen: Do these functions have to return the query strings implicitly? Or can the updated query string be accessed via a separate getter function?

app/search/index/page.tsx
```diff
@@ -1,28 +1,49 @@
...
 
 export type SearchProps = {
-  params: { slug: string };
-  searchParams: { page: string };
+  searchParams: SearchParams;
 };
 
-export default async function Search() {
+export default async function Search({ searchParams }: SearchProps) {
+  const pageName = !searchParams.keywords ? "all-items" : "search-results";
```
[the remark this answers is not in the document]
7emansell: On old DC, the empty keyword search doubles as the "Items" page, but for some reason it's important to differentiate between those for analytics. 
However, we know that the empty search on new DC will actually have items, collections, subcollections– so I don't think the "all items" name would even apply. This is a big open question for product rn.

Constants for the Adobe Analytics page names is out of the scope of this work i think, but sure.

app/search/index/page.tsx
7emansell: Ditto here– this is POC so the eventual query for all items, collections, and subcollections will definitely be different and not worth a constant rn

app/src/utils/searchManager.tsx
7emansell: I think they should return the query string but it's true that these all do two "things" which is wrong. I'll return to the naming once I've locked in the data flow for sure (I think these should be named different things if they no longer need to update state).

PR #274, opened 2025-01-15
DR-3336: Update `fetchApi` to accept cache options

7emansell:
## Ticket:

- JIRA ticket [DR-3336](https://newyorkpubliclibrary.atlassian.net/browse/DR-3336)

- In anticipation of changing the DCF caching strategy, updates `fetchApi` to accept time-based revalidation and a browser cache setting

## Open Questions

- Do we want to prevent our future selves from passing confusing/conflicting cache options (for example, as in the test, `cache: 'no-store'` with `next: {revalidate: 3600}`– setting time revalidation on the serverside request means nothing if the browser won't use cached responses at all) or add a `console.warn` or something?

### Checklist:

[DR-3336]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Updated
+
+- Updated `fetchApi` to accept cache strategy/revalidation parameters (DR-3336)
+
 ## [0.2.6] 2025-1-13
 
 ### Updated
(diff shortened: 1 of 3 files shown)


EdwinGuzman:
> Do we want to prevent our future selves from passing confusing/conflicting cache options....

I think so, but where are you using this function on the client side? I thought all fetch calls were in server components right now.

7emansell:
> > Do we want to prevent our future selves from passing confusing/conflicting cache options....
> 
> I think so, but where are you using this function on the client side? I thought all fetch calls were in server components right now.

Good point– perhaps this is why I shouldn't code unneeded functionality in advance ... potentially I don't think this PR should get merged in at all

PR #275, opened 2025-01-16
DR-3342: Page a11y tests and corresponding updates

7emansell:
## Ticket:

- JIRA ticket [DR-3342](https://newyorkpubliclibrary.atlassian.net/browse/DR-3342)

- Adds (or unskips) axe page tests
- Adds mock response data
- Adds navigation mocking for pages with search or slugs
- Updates image ID when there's no image

### Checklist:

[DR-3342]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


7emansell:
Blocked by (only tests failing now) https://newyorkpubliclibrary.atlassian.net/browse/DSD-1990

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -65,7 +65,7 @@ export function CollectionsPage({ data, params, renderCollections }) {
     setTimeout(() => {
       setIsLoaded(true);
       headingRef.current?.focus();
-    }, 1000);
+    }, 2000);
```
7emansell, line 68: I'm kinda sneaking this in bc it looks so much better.. lmk though

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -161,7 +192,7 @@ export function CollectionsPage({ data, params, renderCollections }) {
               onChange: (e) => handleSearchChange(e),
             }}
             onSubmit={handleSearchSubmit}
-            labelText="Search Collections by title"
+            labelText="Search collections by title"
```
EdwinGuzman, line 195: Just noticing now that this and the labelText on line 188 both are aria labels. They're set for different elements but I wonder if it's redundant (but guess not since it hasn't already been flagged).

EdwinGuzman (approved):

EdwinGuzman:
Thanks for all the new tests!

app/src/components/pages/collectionsPage/collectionsPage.tsx
7emansell: I think it's better because (returning to the ongoing conversation about perceived speed) the loading state is more "accurate" . 
The no results is a gnarly one because we use the length of the returned results to determine how many rows of the skeleton loader we should display, to minimize cumulative layout shift. If the length is 0 (bc there's no results), there's no loader. I'll try `Math.max(Math.ceil(collections.length / 4), 1)` or something to see if that looks weird.
I know we are on the same page but as always the real takeaway here is that even with these bandaids we need a legit solution to our loading state

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -237,7 +268,7 @@ export function CollectionsPage({ data, params, renderCollections }) {
             />
           )
         ) : (
-          Array(Math.ceil(collections.length / 4)).fill(
+          Array(Math.max(Math.ceil(collections.length / 4), 1)).fill(
```
7emansell, line 271: Yeah i think this does look better for the no results loading @EdwinGuzman

PR #280, opened 2025-01-24
DR-3364: Skeleton of `search/index` page

7emansell:
## Ticket:

- JIRA ticket [DR-3364](https://newyorkpubliclibrary.atlassian.net/browse/DR-3364)

- Builds the skeleton of the search page and the filters (which will be reused on the `collections/:uuid` page) with mock data
- Added baseline a11y test for the page
- Removes `mainContent` box from `PageLayout` to allow for the full-width search header on `search/index` and `collections/:uuid` pages
- Moves search page component into its own `/pages/searchPage` folder to parallel the rest of the page components

** I didn't leave comments throughout this code indicating what data needs to be replaced bc it's pretty much everything

### Checklist:

[DR-3364]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman:
> Removes mainContent box from PageLayout to allow for the full-width search header on search/index and collections/:uuid pages

Can you expand on this? I don't completely understand why the margin, maxWidth, and padding has to move from PageLayout to every page component.

7emansell:
> Can you expand on this? I don't completely understand why the margin, maxWidth, and padding has to move from PageLayout to every page component.

It's necessary (or at least necessary to do something) because those properties are not applied on _every_ page component. The search and collection page both have a full width search section at the top that would override this container, the about page has a different maxWidth, and both error pages don't use it at all. For now, I think it makes more sense to only put it on the pages that use it

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+            {`Displaying ${displayResults(data.numResults, CARDS_PER_PAGE, 1)}
+                    results for "keyword"`}
+          </Heading>
+          <Filters />
+        </Box>
+      </Box>
+      <Box
+        maxWidth="1280px"
+        mx="auto"
+        sx={{ paddingLeft: { base: 0, xl: "s" } }}
```
EdwinGuzman, line 51: A concern here is that the styles repeat above.

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+            onClick={() => {}}
+            tagSetData={[
+              { id: "audio", label: "Audio" },
+              { id: "video", label: "Video" },
+            ]}
+            type="filter"
+          />
+        </Flex>
+        <HorizontalRule />
+        <Heading marginTop="xl" size="heading5">{`Displaying ${displayResults(
```
EdwinGuzman, line 70: Wrap this in a div with `aria-live` or add the attr to Heading. Or add `tabIndex={-1}` to programmatically focus on the heading after it gets updated.

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+            isUnderlined={false}
+            hasVisitedState={false}
+            gap="xxs"
+            type="action"
+            href="#"
+          >
+            Back to top{"  "}
+            <Icon
+              name="arrow"
+              sx={{ "> svg": { fill: "ui.link" } }}
```
EdwinGuzman, line 90: Is this needed?

app/src/components/search/filters.tsx
```diff
@@ -0,0 +1,115 @@
...
+  "Publishers",
+  "Division",
+  "Type",
+  "Date Range",
+];
+
+const exampleMultiSelect = (name: string) => {
+  return (
+    <MultiSelect
+      sx={{ width: "-webkit-fill-available" }}
```
EdwinGuzman, line 28: Interesting. Why?

app/src/components/search/filters.tsx
```diff
@@ -0,0 +1,115 @@
...
+                </Tooltip>
+              }
+            />
+          </CheckboxGroup>
+        </>
+      )}
+      <Button
+        id={"see-more-filters"}
+        buttonType="secondary"
+        onClick={() => setIsExpanded(!isExpanded)}
```
EdwinGuzman, line 107: Use the prev value in the callback as `setIsExpanded((prevValue) => !prevValue)`.

EdwinGuzman (commented):
Minor comments but overall the skeleton is great. I'm mostly concern about layout inconsistencies and having to repeat the base layout in every page component.

EdwinGuzman:
> It's necessary (or at least necessary to do something) because those properties are not applied on every page component. The search and collection page both have a full width search section at the top that would override this container, the about page has a different maxWidth, and both error pages don't use it at all. For now, I think it makes more sense to only put it on the pages that use it


It should be possible to set the width/maxwidth on the page and have the header section "breakout" from the layout.

app/src/components/search/filters.tsx
```diff
@@ -0,0 +1,115 @@
...
+        templateColumns="repeat(4, 1fr)"
+        marginBottom="m"
+        width="fill"
+        sx={{ gap: "m" }}
+      >
+        {isExpanded
+          ? exampleFacetCategories.map((filter) => exampleMultiSelect(filter))
+          : exampleFacetCategories
+              .slice(0, 4)
+              .map((filter) => exampleMultiSelect(filter))}
```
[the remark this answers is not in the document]
7emansell: Not sure I can use the `MultiselectGroup`– it overrides the grid they are meant to display in and has breakpoints built in that are different from DC. It's also not design-confirmed that only one can open at a time

7emansell:
> It should be possible to set the width/maxwidth on the page and have the header section "breakout" from the layout.

For sure, is this the pattern you'd like to see?

app/src/components/search/filters.tsx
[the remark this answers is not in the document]
EdwinGuzman: Actually, this was unnecessary on my part. In thhe end, this won't be a `MultiSelect` but the custom "select" component. So it's ok to ignore.

(Do confirm if it's one opened at a time)

EdwinGuzman:
Yes, that will help make the layout more consistent and in one place.

7emansell:
We could ? cc @EdwinGuzman @avertrees

EdwinGuzman:
The pages are behind the RP and the middleware, so the only place you can really see it is on vercel. That'll work for vqa but not QA (they should ideally test the qa env). If it's a feature branch, merging it in can be the final step after a complete a11y, vqa, and qa review and approval of the pages. Otherwise, removing the middleware and RP can be the final step after the review and approval.

Also, consider that there will be changes on other pages while the two new ones are created. Those should be tested separately as new changes are implemented. So perhaps just merging directly into `main` is the way to go. This will work very well if there is a good regression testing plan from QA.

7emansell:
If we're okay with just merging into `main` as we have been let's do that– I think the only page that should be regression tested (more than QA would just check normally) is `/collections` and that's already exposed and will be first to be tested, so that'll work out well

app/src/components/search/filters.tsx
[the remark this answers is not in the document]
EdwinGuzman: Yeah, you'd have to keep track of all the new custom select components when you get to that, but it's a future problem. If there already is a ticket to create the custom select component, add this as AC.

EdwinGuzman:
Sounds good, asking for a regression test before deploying to production so nothing is broken would be a good process to have as well (or automated tests after a deployment as LSP tries to do).

PR #284, opened 2025-01-29
DR-3363: Search result card

7emansell:
## Ticket:

- JIRA ticket [DR-3363](https://newyorkpubliclibrary.atlassian.net/browse/DR-3363)

- Creates the search result model and type
- Creates mock data for search cards as well as a raw search response
- Creates the `searchCard` component
- Creates the `searchCardsGrid` component
- Adds some styling updates to the existing skeletons of `search` and `collections/:uuid` pages
    - Adds the sort menu to search page (for skeleton accuracy, but this will be VQAd in [a separate PR](https://newyorkpubliclibrary.atlassian.net/browse/DR-3366))

This PR is specifically about the search result model/card, but it's displayed on the full search page– so, to limit the scope, don't worry about anything on the page above the "Displaying x results" heading. 

Visible on `/search/index`, and since we're just using mock data for search results right now so you can't test an empty search, I added two `searchCardsGrid`s, one with keywords and one without. (The comments indicating which is which will be removed).

Most significant files in this PR are `components/card/searchCard.tsx`, `models/searchCard.ts`, and `types/searchCardType`.

### Checklist:

[DR-3363]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/types/SearchCardType.ts
```diff
@@ -0,0 +1,26 @@
+export type SearchResultRecordType = "Collection" | "Sub-collection" | "Item";
+export type SearchResultContentType =
+  | "Image"
+  | "Audio"
+  | "Video"
+  | "PDF"
+  | null;
+
+export type SearchResultHighlightType = { field: string; text: string }[];
```
7emansell, line 9: this may end up not being an array– the `highlights` field on the search response is still somewhat under construction

app/src/models/searchCard.ts
```diff
@@ -0,0 +1,43 @@
...
+  recordType: SearchResultRecordType;
+  numberOfDigitizedItems: number;
+  containsOnSiteMaterial: boolean;
+  containsAVMaterial: boolean;
+  containsMultipleCaptures?: boolean;
+  contentType: SearchResultContentType;
+  highlights: SearchResultHighlightType;
+  firstIndexed: string;
+
+  constructor(data: any) {
```
Toxiapo, line 23: Just thinking out loud here, but the data type is essentially the `SearchCardType`, though with some keys being slightly different because they come from the server response. Is that correct?

app/src/components/pages/collectionPage/collectionPage.tsx
```diff
@@ -12,13 +12,15 @@ import {
...
+  Menu,
 } from "@nypl/design-system-react-components";
 import React from "react";
 import Filters from "../../search/filters";
 import { headerBreakpoints } from "@/src/utils/breakpoints";
-import { CardsGrid } from "../../grids/cardsGrid";
 import { displayResults } from "@/src/utils/utils";
-import { CARDS_PER_PAGE } from "@/src/config/constants";
+import { CARDS_PER_PAGE, SEARCH_SORT_LABELS } from "@/src/config/constants";
+import { mockSearchResponse } from "__tests__/__mocks__/data/mockSearchResponse";
```
Toxiapo, line 22: nit: unused import

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+export const SearchCard = ({ result, keywords }: SearchCardProps) => {
+  return (
+    <Card
+      id={result.uuid}
+      imageProps={{
+        alt: "",
+        aspectRatio: "sixteenByNine",
+        id: result.imageID
+          ? `image-${result.imageID}`
+          : `no-image-${result.imageID}`,
```
Toxiapo, line 110: I dont think `imageID` can be undefined or null 🤔

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+    </Box>
+  );
+};
+
+export const SearchCard = ({ result, keywords }: SearchCardProps) => {
+  return (
+    <Card
+      id={result.uuid}
+      imageProps={{
+        alt: "",
```
Toxiapo, line 106: Curious what's the reason to leave this prop empty.

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+          margin: 0,
+          display: "inline",
+        }}
+      >
+        {highlight.field}:{" "}
+      </Text>
+      {words.map((word: string, index: number) => {
+        const isKeyword = keywords.some(
+          (keyword: string) => keyword.toLowerCase() === word.toLowerCase()
+        );
```
Toxiapo, line 75: Should the highlight cover the entire word, or just any matching part? Right now, when the search term is "example in," it also highlights "in" separately.

Toxiapo (commented):
This looks great! I just have a few questions that are mainly to help me understand the codebase

app/src/components/card/searchCard.tsx
7emansell: Image is described by the title on the card, this is a pattern we've confirmed elsewhere with Clare 👍

app/src/components/card/searchCard.tsx
7emansell: It's meant to highlight any word that appears in the keyword string, so yes, for "example in" it's expected that anywhere "example" or "in" appears is highlighted. This behavior is also still being sort of figured out on the product side so this is just the initial approach

app/src/models/searchCard.ts
7emansell: Yes exactly

app/src/components/card/searchCard.tsx
Toxiapo: Great, thank you!

7emansell:
Waiting on VQA before merging this in

__tests__/__mocks__/data/mockSearchCards.tsx
```diff
@@ -0,0 +1,74 @@
+import SearchCardType from "@/src/types/SearchCardType";
```
EdwinGuzman, line 1: nit: `import type ...`

__tests__/__mocks__/data/mockSearchResponse.tsx
```diff
@@ -0,0 +1,69 @@
...
+      title: "Farm Security Administration Photographs",
+      imageID: "1952272",
+      numberOfDigitizedItems: 36,
+      containsOnSiteMaterial: false,
+      containsAVMaterial: false,
+      contentType: null,
+      highlights: {
+        Description: ["Farm in example photographs"],
+      },
+      firstIndexed_dt: "1907-02-01T00:00:00Z",
```
EdwinGuzman, line 50: Is the property name final or until the schemas are final and the API done?

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+import SearchCardType, {
+  SearchResultRecordType,
+} from "@/src/types/SearchCardType";
+
+export interface SearchCardProps {
+  result: SearchCardType;
+  keywords: string;
+}
+
+const onSiteMaterialBadge = (recordType: SearchResultRecordType) => {
```
EdwinGuzman, line 22: As you keep working on this, would it help to have small classes for each of the record types? I'm mostly wondering if it will be helpful to have a class for an item, for example, so that you don't have to check `recordType === "Item"` or similar for the `contentType === "Image"` below. 

Or, how about a new class property for the badge display text in order to move the logic to the class. It would be very similar to `this.url` where it's also checking `data.recordType === "item"`. As much as possible to move logic to the class and "simply" display content in the component can be helpful for readability.

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+  );
+};
+
+const highlightedText = ({ highlight, keyword }) => {
+  const words = highlight.text.split(" ");
+  const keywords = keyword.split(" ");
+  return (
+    <Box
+      sx={{
+        [`@media screen and (max-width: 600px)`]: {
```
EdwinGuzman, line 58: Can you use the new breakpoint? Is it 600px in the designs?

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+      </Text>
+      {words.map((word: string, index: number) => {
+        const isKeyword = keywords.some(
+          (keyword: string) => keyword.toLowerCase() === word.toLowerCase()
+        );
+
+        return (
+          <span key={index}>
+            {isKeyword ? (
+              <Text
```
EdwinGuzman, line 80: This is rendering a `p` inside of a `span`. Not technically wrong but better to not have block elements inside inlined ones. You can add `as="span"` and then remove the "display" rule below.

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+        id: result.imageID
+          ? `image-${result.imageID}`
+          : `no-image-${result.uuid}`,
+        isAtEnd: false,
+        isLazy: true,
+        size: "default",
+        src: result.imageURL,
+      }}
+      layout="row"
+      mainActionLink={result.url}
```
EdwinGuzman, line 117: A minor concern: on the search index page, for the first three card examples there is a lot of white space on the right side. It's still clickable and I wonder if it's confusing. Something to ask Lily/Lana.

app/src/components/card/searchCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const displayLabel =
+    result.recordType === "Item"
+      ? result.contentType === "Image" && result.containsMultipleCaptures
+        ? "Multiple images"
+        : result.contentType
+      : result.recordType;
+
+  return (
+    <TagSet
+      onClick={() => {}}
```
EdwinGuzman, line 42: Are these meant to be clickable? If so, the entire card is clickable so it won't work. Try the following:
```
<TagSet
      onClick={() => {
        console.log("ok")
        alert("test")
      }}
```

`onClick` is optional so it should be ok to remove if it won't be used.

EdwinGuzman (commented):
This is great! I tried to mostly leave comments on things that you may not be waiting on data or vqa for.

__tests__/__mocks__/data/mockSearchResponse.tsx
7emansell: The schemas for this endpoint are considered done, but I'm sure I can ask Kris to revisit since the endpoint is still being developed

app/src/components/card/searchCard.tsx
7emansell: Designs still pending and this can be worked out in VQA

app/src/components/card/searchCard.tsx
EdwinGuzman: Are you trying to find out from Product?

app/src/components/card/searchCard.tsx
EdwinGuzman: 768px. Leave this as is, it's used elsewhere.

app/src/components/card/searchCard.tsx
7emansell: No just seems like they shouldn't be since all you can do is read them. Not sure why `TagSet` renders these as buttons even when `onClick` and `isDismissible` aren't passed

EdwinGuzman: Yes... good point. Can you create a ticket, please? They shouldn't be buttons if they're only meant to display a value and not be clickable.

app/src/components/card/searchCard.tsx
EdwinGuzman: The tags still give the impression they can be clicked and when I click on them, it goes to the collection. Maybe it's just me but nbd right now.

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -81,18 +93,67 @@ const SearchPage = ({ data }) => {
...
+          <Heading
+            size="heading5"
+            tabIndex={-1}
+            margin="0"
+          >{`Displaying ${displayResults(
+            data.numResults,
+            CARDS_PER_PAGE,
+            searchManager.page
+          )}
+                                        results`}</Heading>
```
EdwinGuzman, line 121: weird spacing? 
Screenshot 2025-02-14 at 5 10 31 PM

app/src/components/card/searchCard.tsx
7emansell: https://newyorkpubliclibrary.atlassian.net/browse/DSD-2012

app/src/components/card/searchCard.tsx
EdwinGuzman: This is still a concern.

app/src/utils/utils.ts
```diff
@@ -121,3 +118,19 @@ export function displayResults(
   const end = Math.min(page * perPage, numFound);
   return `${start}-${end} of ${numFound}`;
 }
+
+export function formatHighlightText(highlights) {
```
EdwinGuzman, line 122: There are really good tests to test this but they're indirect. For a utility function like this, add more direct function tests.

EdwinGuzman (commented):
Looks good but just add tests for `formatHighlightText`

PR #286, opened 2025-01-31
DR-3365: Implement search approach for `/collections`

7emansell:
## Ticket:

- JIRA ticket [DR-3365](https://newyorkpubliclibrary.atlassian.net/browse/DR-3365)

- Creates the `SearchProvider` context and adds it to `PageLayout`
- Creates the `SearchManager` class, composed of `CollectionSearchManager` and `GeneralSearchManager` classes, and adds a factory for creating instances
- Rewrites `/collections` page to use an instance of `CollectionSearchManager`
    - Also updates loading behavior given new data flow
- Adds tests for all above

NOTE: Filtering functionality is out of the scope of the PR, since `/collections` doesn't filter results. That said, I included a basic implementation of filtering in `SearchManager` and tests for this functionality. This work will definitely evolve in another PR so doesn't need feedback rn. 
I also updated `/search` to use the new search context introduced in this PR, to confirm that accessing the `GeneralSearchManager` instance from context works as expected

For a written description of what's going on here: [the search TAD ](https://docs.google.com/document/d/1BEIhffdr8_iMK2-hdwd6PFX4_SjR6fGyoYNXhPmyFXw/edit?usp=sharing)

### Checklist:

[DR-3365]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,125 +1,68 @@
...
-  const onMenuClick = async (id) => {
-    setCurrentSort(id);
-    const queryString = createCollectionsQueryStringFromObject({
-      collection_keywords: currentCollectionKeywords,
-      sort: id,
-      page: String(currentPage),
-    });
-    updateURL(`${queryString}#collections`);
-  };
+  const isPopulatedSearch =
```
7emansell, line 52: should this be on the `SearchManager` class instead maybe?

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,125 +1,68 @@
...
+    collectionSearchManager.keywords !== DEFAULT_SEARCH_TERM ||
+    collectionSearchManager.sort !== DEFAULT_COLLECTION_SORT ||
+    collectionSearchManager.page !== DEFAULT_PAGE_NUM;
 
   useEffect(() => {
     setIsLoaded(true);
-  }, []);
+    if (isPopulatedSearch) {
```
7emansell, line 59: Current behavior is that the page doesn't focus on the "Displaying results" unless you've populated the search, so matching this here

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,125 +1,69 @@
...
+      headingRef.current?.focus();
+    }
+  }, [data]);
```
7emansell, line 62: Making this `useEffect` dependent on the new data means the loaders actually correspond to when data arrives!! unfortunately it rerenders in a flickery way, most noticeable on keyword search (and not an issue on pagination/sort updates as far as i can tell). It seems like so close to the magic combination that will make it look right...

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Yes, makes sense to move it there.

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: Yes, but the focus does not go to the "Displaying results" text for the following cases which I think it should:
- search for "cat", clear the input and make an empty search for all items
- change the sort and then change it back to its default value
- go to any page, click on page 1 for the initial page.

app/src/components/pages/collectionsPage/collectionsPage.tsx
EdwinGuzman: I don't see a flicker when performing keyword searches.

I was to comment that eslint complains that the hook is missing dependencies (isPopulatedSearch and isLoaded), but when I try adding them there, I see the flicker. Add a eslint ignore flag to ignore this because it seems like you do only want this to update based on the data and not the update of the other variables. Still... eslint flags it because it is a best practice and recommendation.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -181,16 +126,21 @@ export function CollectionsPage({ data, params, renderCollections }) {
...
-            onChange: (e) => handleSearchChange(e),
+            defaultValue: collectionSearchManager.keywords,
+            onChange: (e) =>
+              collectionSearchManager.handleKeywordChange(
+                (e.target as HTMLInputElement).value
+              ),
+          }}
+          onSubmit={() => {
+            setIsLoaded(false);
+            updateURL(collectionSearchManager.handleSearchSubmit());
```
EdwinGuzman, line 142: Can you try adding `updateURL` inside of `setIsLoaded` as part of the callback function? I don't see a difference but just trying to reduce async issues.

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -225,60 +175,83 @@ export function CollectionsPage({ data, params, renderCollections }) {
...
+              COLLECTION_SORT_LABELS[collectionSearchManager.sort]
+            }`}
             listItemsData={[
               {
                 id: "date-desc",
                 label: "Newest to oldest",
-                onClick: onMenuClick,
+                onClick: () => {
+                  setIsLoaded(false);
+                  updateURL(
```
EdwinGuzman, line 188: Similar comment as above: try moving `updateURL` into `setIsLoaded`

app/src/utils/apiHelpers.ts
```diff
@@ -190,3 +193,22 @@ export const getLaneData = async ({
   const apiUrl = `${process.env.API_URL}/api/v2/collections?genre=${slug}&page=${pageNum}&per_page=${perPage}`;
   return await fetchApi(apiUrl);
 };
+
+export const getSearchData = async ({
```
EdwinGuzman, line 197: Can you add esdoc style comments and include that this is used on the `/search/index` page? Just to associate a page route with the relevant Repo API endpoint.

app/src/utils/apiHelpers.ts
```diff
@@ -190,3 +193,22 @@ export const getLaneData = async ({
...
+  keyword = DEFAULT_SEARCH_TERM,
+  sort = DEFAULT_SORT,
+  filters = DEFAULT_FILTERS,
+  page = DEFAULT_PAGE_NUM,
+  perPage = CARDS_PER_PAGE,
+}: {
+  keyword?: string;
+  sort?: string;
+  filters?: Filter[];
```
EdwinGuzman, line 206: I know you mentioned that filters is not part of this work, but where would the mapping of this array into string happen?

EdwinGuzman (commented):
Still mid-review but wanted to leave some comments now.

app/search/index/page.tsx
```diff
@@ -1,28 +1,50 @@
...
       breadcrumbs={[
         { text: "Home", url: "/" },
         { text: "Keyword Search", url: "/search/index" },
       ]}
-      adobeAnalyticsPageName={createAdobeAnalyticsPageName(
-        "search-results",
-        ""
-      )} //TODO: if there are no query params, page name should be createAdobeAnalyticsPageName("all-items", "")
+      adobeAnalyticsPageName={createAdobeAnalyticsPageName(pageName, "")}
+      searchParams={searchParams} //!
```
EdwinGuzman, line 45: `!` typo or something to pay attention to?

app/search/index/page.tsx
```diff
@@ -1,28 +1,50 @@
...
-  params: { slug: string };
-  searchParams: { page: string };
+  searchParams: SearchParams;
 };
 
-export default async function Search() {
+export default async function Search({ searchParams }: SearchProps) {
+  const pageName = searchParams.keywords ? "search-results" : "all-items";
+
+  const data = await getSearchData({
```
EdwinGuzman, line 30: The collection page redirects to the 404 with a bad API response. Should the same be done here?

app/src/utils/apiHelpers.ts
7emansell: https://github.com/NYPL/digital-collections/blob/77afc22d27e15b54ad2a18fa63b6555fae171d85/app/src/utils/searchManager.tsx#L77 rn but doesn't have to

app/src/utils/searchManager.tsx
```diff
@@ -0,0 +1,219 @@
...
+      page: this.currentPage,
+    })}`;
+  }
+
+  protected createQueryString(params: Record<string, any>) {
+    return filterQueryStringFromObject(params);
+  }
+}
+
+export class SearchManagerFactory {
```
EdwinGuzman, line 184: Great update since the POC.

EdwinGuzman (commented):
Great! Some comments, mostly nits. Will review the logic for searching to figure out the flickering issue.

app/src/components/pages/collectionsPage/collectionsPage.tsx
7emansell: offline– look at Vercel. Also for posterity, strict mode is NOT enabled in this repo

app/search/index/page.tsx
7emansell: something to pay attention to

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -39,8 +46,12 @@ const SearchPage = ({ data }) => {
...
             }}
           >
-            {`Displaying ${displayResults(data.numResults, CARDS_PER_PAGE, 1)}
-                    results for "keyword"`}
+            {`Displaying ${displayResults(
+              data.numResults,
+              CARDS_PER_PAGE,
+              searchManager.page
+            )}
+                    results for "${searchManager.keywords}"`}
```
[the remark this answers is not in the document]
7emansell: How so

app/src/components/pages/collectionsPage/collectionsPage.tsx
7emansell: Hmm. If an empty search for all items should trigger focus than how are we supposed to distinguish between that and the first time the user hits the page ? I think better to just always send focus, but I'm not sure how bad that is a11y-wise

charmingduchess (changes requested):

app/collections/page.tsx
```diff
@@ -30,16 +32,18 @@ export default async function Collections({ searchParams }: CollectionsProps) {
...
-        data={data}
-        params={searchParams}
-        renderCollections={renderCollections}
-      />
-    </Suspense>
+    <PageLayout
+      activePage="collections"
+      breadcrumbs={[
+        { text: "Home", url: "/" },
+        { text: "Collections", url: "/collections" },
```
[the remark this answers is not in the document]
7emansell: Yes, there's differences– in pages with slugs it'll be different, and some pages don't use the breadcrumbs at all

app/search/index/page.tsx
7emansell: Resolving since it's out of scope for this

app/src/components/pageLayout/pageLayout.tsx
```diff
@@ -47,37 +52,39 @@ const PageLayout = ({
...
-                {children as JSX.Element}
-              </Box>
-            </>
-          )}
-        </FeedbackProvider>
+        <SearchProvider searchParams={searchParams}>
+          <FeedbackProvider>
+            <SkipNavigation />
+            <Header />
+            {activePage === "home" ||
```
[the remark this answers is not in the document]
7emansell: Out of scope but noted and will be revisited when we adopt the new `TemplateAppContainer`

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,125 +1,65 @@
...
+  const [isLoaded, setIsLoaded] = useState(false);
+  const totalPages = totalNumPages(data.numResults, data.perPage);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const collections = data.collection
+    ? Array.isArray(data.collection)
+      ? data.collection
+      : [data.collection]
+    : [];
+
+  const collectionSearchManager = SearchManagerFactory.createSearchManager({
```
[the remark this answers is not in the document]
7emansell: Very true, updated!!

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -1,125 +1,65 @@
...
-  );
-
-  const [currentCollectionKeywords, setcurrentCollectionKeywords] =
-    useState<string>(params.collection_keywords || DEFAULT_SEARCH_TERM);
-
-  const updateURL = async (queryString) => {
+  const [isLoaded, setIsLoaded] = useState(false);
+  const totalPages = totalNumPages(data.numResults, data.perPage);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const collections = data.collection
```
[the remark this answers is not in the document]
EdwinGuzman: Hmm maybe a `Collections` class instantiated in `app/collections/page.tsx` before it gets passed to this component. There are other properties used from `data` but it would be a very small class that it may not be worth it.

app/src/utils/searchManager.test.tsx
EdwinGuzman, line 1: 👍

EdwinGuzman (approved):
Great work!

PR #287, opened 2025-02-05
No ref: DC custom search bar

7emansell:
## Ticket:

n/a

Should be merged in with [DR-3365](https://github.com/NYPL/digital-collections/pull/286) since it corrects the loading

- Replicates the DS `Searchbar` component but pared down to what DC uses, specifically getting rid of the `<form>` wrapper, which was triggering extra search submits
    - Added `maxWidth` as a prop since that's something we have been overriding but not everywhere

## Open Questions

- Generally, is it okay to have removed so much of the `Searchbar`? Wanted to keep what we actually use (and will use in the foreseeable future), so for example I removed the entire `select` element and all its corresponding props.
- Is it okay to add the "Enter" keyboard behavior back in (see comment)?

`search.test.tsx` tests should still cover this, and it should look and work exactly the same as before

- Want to check I haven't dropped any `aria-labels` or anything in the migration

### Checklist:

[DR-3365]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/pages/collectionsPage/collectionsPage.tsx
+++ b/app/src/components/pages/collectionsPage/collectionsPage.tsx
@@ -83,7 +83,8 @@ export function CollectionsPage({ data, collectionSearchParams }) {
           subtitle="Explore the New York Public Library's diverse collections, including digitized photographs, manuscripts, maps, and more. Start exploring by using the search bar below or browse through the collections."
         />
         <DCSearchBar
-          id={"search-collections"}
+          id="search-collections"
+          labelText="Search collections"
           maxWrapperWidth="462px"
           textInputProps={{
             id: "textinput",
--- a/app/src/components/pages/collectionsPage/collectionsPage.tsx
+++ b/app/src/components/pages/collectionsPage/collectionsPage.tsx
@@ -4,7 +4,6 @@ import {
   Box,
   Heading,
   HorizontalRule,
-  SearchBar,
   Menu,
   Pagination,
   Flex,
@@ -23,6 +22,7 @@ import {
 } from "@/src/config/constants";
 import { SearchManagerFactory } from "@/src/utils/searchManager";
 import { headerBreakpoints } from "@/src/utils/breakpoints";
+import DCSearchBar from "../../search/dcSearchBar";
 
 export function CollectionsPage({ data, collectionSearchParams }) {
   const { push } = useRouter();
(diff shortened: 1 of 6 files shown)


app/src/components/search/dcSearchBar.tsx
```diff
@@ -0,0 +1,204 @@
...
+        isClearable={textInputProps?.isClearable}
+        isClearableCallback={textInputProps?.isClearableCallback}
+        name={textInputProps?.name}
+        onChange={textInputProps?.onChange}
+        pattern={textInputProps?.pattern}
+        placeholder={textInputPlaceholder}
+        textInputType="searchBar"
+        type="text"
+        value={textInputProps?.value}
+        onKeyDown={(event) => {
```
7emansell, line 104: Added this to mimic the lost `<form>` behavior

samanthaandrews:
Took a few minutes today to orient myself around this code. Will submit review tomorrow morning.

app/src/components/search/dcSearchBar.tsx
samanthaandrews: i think this is nice!

samanthaandrews (changes requested):
just a few small changes! it looks great! please re-request review from me when you need it

app/src/components/search/dcSearchBar.tsx
```diff
@@ -0,0 +1,204 @@
...
+          },
+        }}
+      >
+        <Icon
+          align="left"
+          id={`searchbar-icon-${id}`}
+          name="search"
+          size="small"
+        />
+        <span>Search</span>
```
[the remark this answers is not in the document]
7emansell: This makes sense!! added

samanthaandrews (approved):
🚀

app/src/components/pages/collectionsPage/collectionsPage.tsx
```diff
@@ -23,11 +22,13 @@ import {
 } from "@/src/config/constants";
 import { SearchManagerFactory } from "@/src/utils/searchManager";
 import { headerBreakpoints } from "@/src/utils/breakpoints";
+import DCSearchBar from "../../search/dcSearchBar";
 
 export function CollectionsPage({ data, collectionSearchParams }) {
   const { push } = useRouter();
   const pathname = usePathname();
   const headingRef = useRef<HTMLHeadingElement>(null);
+  const isFirstLoad = useRef<boolean>(false);
```
EdwinGuzman, line 31: I gave you this variable name and now that I'm reading it, it could probably be better named. Maybe `isAfterFirstLoad`? Mostly because it's true every time except on the initial load.

app/src/components/search/dcSearchBar.tsx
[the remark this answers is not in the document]
EdwinGuzman: I do wonder if it's a bit self-referential since the span is contained within the button, but this seems ok.

EdwinGuzman:
> Generally, is it okay to have removed so much of the Searchbar? Wanted to keep what we actually use (and will use in the foreseeable future), so for example I removed the entire select element and all its corresponding props.

Yes, no worries about this. Can you create a ticket on the DS board and link it to this ticket https://newyorkpubliclibrary.atlassian.net/browse/DSD-1889? We'll keep this issue and wrapping elements in a `<form>` in mind for ds v4.

> Is it okay to add the "Enter" keyboard behavior back in (see comment)?

Yes, thanks for adding it. Typically, that's why native HTML elements are recommended so we don't have to reinvent them and add them manually.

7emansell:
What would the DS ticket include? The changes here are not things I would recommend for the broader DS usage of `Searchbar`, it's just that we don't use them in DC @EdwinGuzman

EdwinGuzman:
Add that the `form` wrapper should be removed so that nextjs apps that use the App router and/or this type of search won't have rendering issues. It's also mostly to keep track on the DS side that there's a custom version in DCF.

7emansell:
https://newyorkpubliclibrary.atlassian.net/browse/DSD-2001

PR #288, opened 2025-02-06
DR-3394: Custom filter select component

7emansell:
## Ticket:

- JIRA ticket [DR-3394](https://newyorkpubliclibrary.atlassian.net/browse/DR-3394)

- Creates `selectFilter` component using `Chakra`'s `Accordion` and the DS `RadioGroup` components
- Creates `selectFilterGrid` 
- Both of the above using mock data for now
- Also adds tests for both

-

### Checklist:


app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+    const newController = new AbortController();
+    setController(newController);
+
+    setTimeout(() => {
+      if (!newController.signal.aborted) {
+        setSelected(newSelection);
+        console.log(`selected: ${newSelection}`);
+      }
+    }, 400);
```
7emansell, line 73: You can check logs in the console to see when an option is "selected"– does this feel too laggy or too fast? Want to find a good compromise time

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <Icon
+        name="arrow"
+        size="small"
+        color="ui.black"
+        iconRotation={iconRotation}
+      />
+    );
+  };
+
+  const modalLink = (facet: string) => {
```
EdwinGuzman, line 49: Eventually this should be a button that opens the modal rather than a link.

app/src/components/search/filters/selectFilter.tsx
EdwinGuzman: I'd like this slightly longer. This is hard to tell until real data changes the page but I imagine users will take some time to switch between values in case they selected the wrong one.

Another thing to consider, when a selection is clicked or moved to (using arrows), the drop down should close, the page updates, and focus goes to the display text, right? If that's so, then the timing might not matter too much.

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+        facet === "Publishers" ? `` : `s`
+      }`}</Link>
+    );
+  };
+
+  const onChange = (newSelection: string) => {
+    if (controller) {
+      controller.abort();
+    }
+    const newController = new AbortController();
```
EdwinGuzman, line 65: I will have to double check but I'm pretty sure AbortController should _only_ be used with `fetch`. I imagined this being used on the server component to cancel requests when multiple ones are made, not necessarily for controlling user events. This might be doable with a boolean value instead of creating an abort controller but I'll verify this.

EdwinGuzman (changes requested):
Awesome! 

I'm requesting changes because it doesn't meet all the functionality in the ticket. I mentioned I'll double check the `AbortController` usage and I'm not too concerned about the link vs button opening the modal for this PR. However, the filters should close when you click outside or focus elsewhere on the page. Try using the `useCloseDropDown` hook https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/hooks-useclosedropdown--docs

app/src/components/search/filters/selectFilter.tsx
7emansell: https://newyorkpubliclibrary.atlassian.net/browse/DR-3396

app/src/components/search/filters/selectFilter.tsx
7emansell: I'll check in with Lily/Lana on this (it's not mentioned in requirements or designs)– that makes sense but I think it actually makes the timing more important, otherwise there will be a really jumpy user experience

app/src/components/search/filters/selectFilter.tsx
EdwinGuzman: I am finding more info on using the `AbortController` with fetch rather than without. This was probably one of the most helpful https://blog.comendrun.com/enhancing-reactnextjs-app-with-native-abortcontroller-interface-in-javascripttypescript

You're right that you should handle the behavior here rather than in the API handlers, but you can do the same without the `AbortController`. It adds a bit more complexity and is not associated with any API requests.

As an aside, this was also helpful but it's more related to timeouts. This might be helpful, at a later time, for the fetch with timeout function https://dtang.dev/timing-out-fetch-requests/

EdwinGuzman (approved):
Great!

7emansell:
Sorry, Lily just got back to me on the selection behavior. Had to update the DS's `useCloseDropDown` to be triggered on focus changing as well

EdwinGuzman (approved):
Get an update regarding how the focus out behavior should work but not a blocker.

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,145 @@
...
+};
+
+const SelectFilter = ({
+  filter,
+  isOpen,
+  onToggle,
+  headingRef,
+}: SelectFilterProps) => {
+  const [selected, setSelected] = useState<string | null>(null);
+  const [timeoutId, setTimeoutId] = useState<NodeJS.Timeout | null>(null);
```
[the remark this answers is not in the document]
7emansell: Is that a bad thing

app/src/components/search/filters/selectFilterGrid.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  useCloseDropDown(() => setOpenFilter(null), gridRef);
+
+  const handleToggle = (filterName: string) => {
+    setOpenFilter((prev) => (prev === filterName ? null : filterName));
+  };
+
+  return (
+    <Grid
+      ref={gridRef}
```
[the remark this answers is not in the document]
7emansell: It's not, is resolved now

app/src/hooks/useCloseDropDown.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  // Listen for click events on the document outside of the
+  // element that the `ref` prop points to.
+  useOutsideClick({
+    ref,
+    handler: () => actionCb(false),
+  });
+
+  // Listen for focus leaving the referenced element
+  useEffect(() => {
```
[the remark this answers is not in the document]
7emansell: Didn't end up being necessary 👍

Toxiapo (commented):
Nice! Just a couple questions.

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const onChange = (newSelection: string) => {
+    if (timeoutId) {
+      clearTimeout(timeoutId);
+    }
+    const newTimeoutId = setTimeout(() => {
+      console.log(`selected: ${newSelection}`);
+      setUserClickedOutside(true);
+      (
+        headingRef as MutableRefObject<HTMLHeadingElement | null>
```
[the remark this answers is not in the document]
7emansell: Yes this is expected for now

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,185 @@
...
+        setUserClickedOutside(false);
+      }
+    }
+  };
+
+  // Tells the accordion to close if open when user tabs outside of the container
+  const handleTabOutside = (e) => {
+    if (e.key === "Tab") {
+      const selectComponent = containerRef.current;
+      setTimeout(() => {
```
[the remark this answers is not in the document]
7emansell: No it is redundant!! Did not clean up, now removed

app/src/components/search/filters/dateFilter.tsx
```diff
@@ -0,0 +1,35 @@
...
+          width="100px"
+          id="dateStart"
+          labelText="Start year"
+          placeholder="Start year"
+          showLabel={false}
+        />
+        <Text sx={{ marginBottom: "0" }}> to </Text>
+        <TextInput
+          width="100px"
+          id="dateStart"
```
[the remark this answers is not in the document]
7emansell: Updated

app/src/components/search/filters/selectFilterGrid.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  { name: "Publishers", options: [{ name: "Publisher 1", count: 20 }] },
+  { name: "Type", options: [{ name: "Type 1", count: 15 }] },
+  { name: "Genre", options: [{ name: "Genre 1", count: 30 }] },
+];
+
+describe("SelectFilterGrid", () => {
+  it("renders the correct number of filters when collapsed", () => {
+    render(<SelectFilterGrid filters={mockFilters} isExpanded={false} />);
+
+    expect(screen.getByText("Collection")).toBeInTheDocument();
```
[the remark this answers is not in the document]
7emansell: I think the pattern I've been trying to follow is based on how the user could interact with the component, so framing `expect`s not by selecting by role or `data-testid` etc., but using text they could see on the screen or keyboard commands. But good point and definitely up for discussion cc @EdwinGuzman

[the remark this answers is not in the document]
7emansell: Sometimes it's necessary to use the role which I literally do in this test file. But I have been trying to start with visual elements

Toxiapo (approved):
🚀

app/src/components/search/filters/filterAccordion.tsx
```diff
@@ -0,0 +1,209 @@
...
+  Box,
+  chakra,
+  ChakraComponent,
+} from "@chakra-ui/react";
+import { Icon } from "@nypl/design-system-react-components";
+import React, { forwardRef, useEffect, useState } from "react";
+
+export type AccordionTypes = "default" | "warning" | "error";
+
+export interface AccordionDataProps {
```
EdwinGuzman, line 15: `AccordionTypes` and `AccordionDataProps` can be imported from the DS to reduce copy/paste code.

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const accordionButtonRef: React.RefObject<HTMLDivElement> =
+    useRef<HTMLDivElement>(null);
+  const containerRef: React.RefObject<HTMLDivElement> =
+    useRef<HTMLDivElement>(null);
+
+  // Tells the accordion to close if open when user clicks outside of the container
+  const handleClickOutside = (e) => {
+    if (e.type === "mousedown") {
+      const multiSelect = containerRef.current;
+      if (multiSelect && !multiSelect.contains(e.target)) {
```
EdwinGuzman, line 51: Nit: assign the conditional to a variable so you call `setUserClickedOutside(variableName)` once.

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+  };
+
+  // Tells the accordion to close if open when user tabs outside of the container
+  const handleTabOutside = (e) => {
+    if (e.key === "Tab") {
+      const selectComponent = containerRef.current;
+      if (
+        selectComponent &&
```
EdwinGuzman, line 64: similar nit as above

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </RadioGroup>
+      {modalButton(filter)}
+    </>
+  );
+
+  return (
+    <Box
+      {...rest}
+      ref={containerRef}
+      onClick={() => setUserClickedOutside(false)}
```
EdwinGuzman, line 154: The div shouldn't have an `onClick`. I kind of understand why it's added but can you describe the issue this is fixing?

app/src/components/search/filters/selectFilterGrid.test.tsx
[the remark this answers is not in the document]
EdwinGuzman: @dgcohen found last year that `getByRole` was a performance bottleneck compared to `getByLabelText`: https://github.com/NYPL/research-catalog/pull/64

I definitely try to stay away from `data-testid` unless I really, really have to. Otherwise, I switch between `getByLabelText` and `getByRole` but end up using `getByLabelText` more often.

EdwinGuzman (commented):
This looks good. Mostly nits but one question about an `onClick` on a div.

app/src/components/search/filters/filterAccordion.tsx
7emansell: realized I don't need those exact types so removed

app/src/components/search/filters/selectFilter.tsx
7emansell: it's unnecessary– I think when I added it I thought the filter by itself couldn't control when it goes outside, but it is because it's using the `containerRef` on the div

EdwinGuzman (approved):

7emansell:
merging this into main and making any VQA updates on https://github.com/NYPL/digital-collections/pull/290

PR #290, opened 2025-02-12
DR-3396: DC custom filter select modal

7emansell:
## Ticket:

- JIRA ticket [DR-3396](https://newyorkpubliclibrary.atlassian.net/browse/DR-3396)

- Creates the `selectFilterModal` based off the Chakra `Modal` 
- Connects it to the `selectFilter` component
    - Still using mock data, the implementation of the selected option will definitely change with real data

## Open Questions

- Do you agree that the search bar maybe doesn't need the submit button since it's literally live filtering (see comment)? This is not how the designs have it but as I was building it I was like hmmm this would be easier. Can always go back to standard search with pressing the button

- As with the `selectFilter`, keeping an eye on focus management

### Checklist:


app/src/components/search/filters/selectFilterModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+                  labelText: `Search ${filter.name}${
+                    filter.name === "Publishers" ? "" : "s"
+                  }`,
+                  name: "filter_keywords",
+                  placeholder: `Search ${filter.name.toLowerCase()}${
+                    filter.name === "Publishers" ? "" : "s"
+                  }`,
+                  onChange: (e) => setSearchText(e.target.value),
+                }}
+                onSubmit={() => {}}
```
7emansell, line 141: This `onSubmit` is doing nothing because the `onChange` already filters which feels smoother anyways. Lmk what you think

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -72,63 +91,42 @@ const SelectFilterComponent = forwardRef<
...
       (
         headingRef as MutableRefObject<HTMLHeadingElement | null>
-      )?.current?.focus();
+      ).current?.focus();
     }, 600);
 
     setTimeoutId(newTimeoutId);
   };
 
+  const sortedOptions = selected
```
7emansell, line 123: Sorting by selected first is also not specified yet in the designs, so looking for feedback– does it seem strange that it's first, should it also be first when you open the modal? (I thought no to the modal because it's paginated but that's vibes only not really a justification )

app/src/components/search/filters/selectFilterModal.tsx
Toxiapo: Yea, it feels smoother without the submit button, and it's pretty common

app/src/components/search/filters/selectFilterModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+                  id: "filter-search-text",
+                  isClearable: true,
+                  labelText: `Search ${filter.name}${
+                    filter.name === "Publishers" ? "" : "s"
+                  }`,
+                  name: "filter_keywords",
+                  placeholder: `Search ${filter.name.toLowerCase()}${
+                    filter.name === "Publishers" ? "" : "s"
+                  }`,
+                  onChange: (e) => setSearchText(e.target.value),
```
Toxiapo, line 139: I think this needs a `clear` callback? If you try to search something and presses the "X" button, it clears the field but doesn't reset the modal.

app/src/components/search/filters/selectFilterModal.tsx
Toxiapo: fwiw, NYPL Reader doesn't have a submit button on the search bar.

app/layout.tsx
```diff
@@ -77,7 +77,6 @@ export default async function RootLayout({
   return (
     <html lang="en">
       <head>
-        <script src="https://unpkg.com/react-scan/dist/auto.global.js"></script>
```
Toxiapo, line 80: 🤔 Was this the little tooltip toggle on the page? Curious what was it for.

app/src/components/search/filters/selectFilter.tsx
Toxiapo: I like this! But I think it's more of a product question

Toxiapo (commented):
Just the clear button button, otherwise the code looks good to me.

app/layout.tsx
7emansell: react-scan! To see when components were re-rendering

EdwinGuzman:
> Do you agree that the search bar maybe doesn't need the submit button since it's live filtering (see comment)? This is not how the designs have it but as I was building it I was like hmmm this would be easier. Can always go back to standard search with pressing the button

This works very well with immediate feedback but I wonder if this is true because all the data is mocked? Would it be slower if real data has to be fetched? (not on the search input when users type but to gather all the data) 
Either way, for accessibility, there most likely has to be a way to associate a character the user inputs with the updated list of results. Check with Clare. I think it's most helpful to not have the button, but it does go against our patterns and it might be difficult (but not impossible) to make the entire feature accessible.

Just a thought: there is no visual indication that this does "live filtering" of all the options in a category. So If I type "c" and then see the list decrease in size, I do have to think about what's going on. This is mostly because it's not _just_ that a result starts with "c" but the character "c" can be anywhere in the entire string. So, if I'm intentionally searching for "California", then "Wisconsin" doesn't make much sense (although it is at the end of the results so that's good -- lowers priority). This is more of a product question but I hoping to show how it's complicated as it is that it might be more so using a screenreader.

7emansell:
@EdwinGuzman a11y wise I added the `aria-autocomplete=list` to the `input` which I think is a good first step, I'll definitely flag it when the whole thing is reviewed. The sorting of the options and how the list is ranked is definitely a product question so I'll touch base w Lana today

app/src/components/search/filters/filters.tsx
```diff
@@ -72,6 +64,36 @@ const Filters = forwardRef<HTMLHeadingElement, FiltersProps>(
         </Button>
       </>
     );
+
+    return (
+      <>
+        <Box
+          sx={{
+            display: "block",
+            [`@media (max-width: ${headerBreakpoints.lgMobile}px)`]: {
```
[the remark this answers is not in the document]
EdwinGuzman: Ok, I see it now. The window doesn't get narrow enough to trigger this unless I use the responsive tool.

app/src/components/search/filters/filters.tsx
```diff
@@ -72,6 +64,36 @@ const Filters = forwardRef<HTMLHeadingElement, FiltersProps>(
...
+        <Box
+          sx={{
+            display: "block",
+            [`@media (min-width: ${headerBreakpoints.lgMobile}px)`]: {
+              display: "none",
+            },
+          }}
+        >
+          <Accordion
+            isDefaultOpen
```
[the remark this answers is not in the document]
EdwinGuzman: Now that I see this on mobile, it doesn't match the designs that have this initially collapsed and you have to click on "refine your search" to open it.

app/src/components/search/filters/selectFilterModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+                  border: "1px solid",
+                  borderColor: "ui.border.hover",
+                  padding: "s",
+                  height: "394px",
+                }}
+              >
+                <RadioGroup
+                  isFullWidth
+                  // Helper invalid text, even when false, creates a 8px margin.
+                  sx={{ marginBottom: "-8px" }}
```
EdwinGuzman, line 161: Interesting.

app/src/components/search/filters/selectFilterModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+                onClick={handleClose}
+                id="close-button"
+              >
+                Close
+              </Button>
+              <Button
+                id="confirm-button"
+                isDisabled={!selected}
+                onClick={() => {
+                  setSelected(
```
EdwinGuzman, line 220: This is not working as expected. The `onChange` in the RadioGroup already updates the selected value so this "Confirm" button doesn't really confirm the selected choice, it really just closes the modal. You can replicate this by selecting any unselected option and then clicking on "Close". I would expect to not have any value selected but when I open the same select, the value is actually selected.

Separately, but related, when I click "Close", the focus goes to the "Displaying 1-48 of .... results" text but it shouldn't. I did not make a selection that would update the result list, so the focus should go back to the "View all [name]" button.

app/src/components/search/filters/selectFilterGrid.tsx
```diff
@@ -11,7 +12,15 @@ const SelectFilterGrid = forwardRef<HTMLHeadingElement, SelectFilterGridProps>(
   ({ filters, isExpanded }, headingRef) => {
     return (
       <Grid
-        templateColumns="repeat(4, 1fr)"
+        sx={{
+          gridTemplateColumns: "repeat(1, 1fr)",
+          [`@media screen and (min-width: ${headerBreakpoints.lgMobile}px)`]: {
+            gridTemplateColumns: "repeat(2, 1fr)",
```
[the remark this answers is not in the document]
EdwinGuzman: Ignore this, it is working I just didn't see the responsive view well enough.

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -55,63 +74,45 @@ const SelectFilterComponent = forwardRef<
...
-        filter.name === "Publishers" ? `` : `s`
-      }`}</Button>
-    );
-  };
+  }, [isModalOpen]);
 
   const onChange = (newSelection: string) => {
-    if (timeoutId) {
-      clearTimeout(timeoutId);
+    if (!isModalOpen) {
```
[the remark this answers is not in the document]
7emansell: If the modal is open, the parent needs to prevent the dropdown from closing when the user is clicking/tabbing through the modal

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -55,63 +74,45 @@ const SelectFilterComponent = forwardRef<
...
-      )?.current?.focus();
-    }, 600);
-
-    setTimeoutId(newTimeoutId);
   };
 
+  const sortedOptions =
+    selected && !isModalOpen
+      ? [
+          selected,
```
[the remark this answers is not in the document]
7emansell: Just talked to Lana/Lily about this– it's first in the dropdown but in the modal it has to be in its natural order, otherwise you could lose the option once you select it (since it'll go to to the first page)

app/src/components/search/filters/selectFilterModal.tsx
```diff
@@ -0,0 +1,248 @@
...
+                  }}
+                />
+              </Flex>
+            </ModalBody>
+
+            <ButtonGroup padding="m" marginX="auto">
+              <Button
+                buttonType="secondary"
+                onClick={() => {
+                  setSelected(null);
```
EdwinGuzman, line 218: Making my comment here because this was updated in the last commit (I think). This is closer to the desired behavior but it's still off:
- If I open the model and don't make a selection, closing the modal by pressing esc and clicking on "Close" work as expected. The focus goes back to the "View all collections" button. ✅ 
- If I make a selection like "Pennsylvania" and then click on "Close" or press the esc key, the selection does not happen but now the focus is gone and the dropdown is closed. It should focus back on the "View all collections" button.
- I open a dropdown and select "Ohio", this works. I open the dropdown again and "Ohio" is listed first. When I open the modal, "Ohio" is now back to the location it was before it was selected in both the modal and the dropdown in the background. I'm not sure if this is really an issue or expected, but how it jumps around is odd. I decide I want to keep "Ohio" so I click on "Close". My selection is gone and when I open the dropdown, "Ohio" is not selected but it should still be selected. This is not a problem if you press esc instead of clicking on "Close".
- This is similar to the above but another test case: open the dropdown and select an option, let's say "California". Now open the dropdown again, click on "View all collections", select "Long Island", click on "Close". My original selection of "California" is gone.

app/src/components/search/filters/selectFilter.tsx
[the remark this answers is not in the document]
EdwinGuzman: Cool. I read this after I made my comment. Just note that when you open the modal, the dropdown in the background also updates the order of the options.

app/src/components/search/filters/selectFilterModal.tsx
7emansell: Okay introduced new modal selection state that only gets confirmed as the final selection once "Confirm" button is clicked. I think this fixes all the examples you gave here, but I guess I'm wondering if this is really an understandable user experience? Maybe the "Close" button should say "Clear" or something. Lmk what you think now that the behavior is more consistent @EdwinGuzman

EdwinGuzman (approved):
Nice update to fix the modal selection issue. I agree that this is not that understandable to the user. Now that this work is in a good state, it's worth walking through the user flow with Lana and Lily.

app/src/components/mobileSearchBanner/mobileSearchBanner.tsx
```diff
@@ -0,0 +1,19 @@
+import { headerBreakpoints } from "@/src/utils/breakpoints";
+import { Banner } from "@nypl/design-system-react-components";
+
+export const MobileSearchBanner = () => {
+  return (
+    <Banner
+      sx={{
+        marginBottom: "xs",
+        display: "flex",
+        [`@media (min-width: ${headerBreakpoints.lgMobile}px)`]: {
```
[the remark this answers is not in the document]
7emansell: I want to say that I did try this but I don't remember... I'll make a ticket

__tests__/pages/collectionpage.test.tsx
```diff
@@ -10,7 +10,7 @@ jest.mock("next/navigation", () => ({
   usePathname: jest.fn(),
 }));
 
-describe("Collection page accessibility", () => {
+describe.skip("Collection page accessibility", () => {
```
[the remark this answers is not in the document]
7emansell: Duplicate id issues (from mock data), just unskipped though

EdwinGuzman (approved):
Approving the latest round of updates though I know VQA feedback can keep this open.

7emansell:
cleared VQA awaiting a11y

app/src/components/mobileSearchBanner/mobileSearchBanner.tsx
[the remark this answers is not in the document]
7emansell: https://newyorkpubliclibrary.atlassian.net/browse/DR-3454

avertrees (approved):

avertrees (approved):

app/src/utils/searchManager.tsx
```diff
@@ -64,12 +70,27 @@ abstract class BaseSearchManager implements SearchManager {
...
+    return this.currentFacets;
+  }
+
   handleKeywordChange(value: string) {
     this.currentKeywords = value;
   }
 
   handleAddFilter(newFilter: Filter) {
-    this.currentFilters.add(JSON.stringify(newFilter));
+    const existingFilters = new Map(
```
EdwinGuzman, line 83: Interesting. Why did you have to use a `Map` here as well?

EdwinGuzman (approved):

app/src/components/search/filters/selectFilterModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+                  placeholder: `Search ${filter.name.toLowerCase()}${
+                    filter.name === "Publishers" ? "" : "s"
+                  }`,
+                  onChange: (e) => setSearchText(e.target.value),
+                }}
+                onSubmit={() => {}}
+              />
+              <div
+                id="search-live-region"
+                ref={liveRegionRef}
```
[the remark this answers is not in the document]
7emansell: This was before the tooltip, and recommended by Clare, specifically for handling autocomplete search

PR #296, opened 2025-02-24
DR-3414: Collections/Repo API wrappers

7emansell:
## Ticket:

- JIRA ticket [DR-3414](https://newyorkpubliclibrary.atlassian.net/browse/DR-3414)

- Defines two classes (`CollectionsApi` and `RepoApi`) in `apiClients` with their corresponding static methods
- Adds environment variable (`COLLECTIONS_API_URL`)– **needs to be updated in our task definitions with DevOps and you (the developer reading this...) need to update your `.env` according to `.env.example`**
- Moves tests, imports, mock data around accordingly

## Open Questions

- Is updating `fetchApi` (or, probably, just writing a new one for Collections API calls) in scope for this?

Everything should work the same– we're not currently hitting Collections API for anything, this just sets us up to do so. All Repo API calls should work as before

### Checklist:

[DR-3414]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/utils/apiClients.tsx
```diff
@@ -0,0 +1,254 @@
...
+    slug: string;
+    pageNum?: number;
+    perPage?: number;
+  }) {
+    const apiUrl = `${process.env.API_URL}/api/v2/collections?genre=${slug}&page=${pageNum}&per_page=${perPage}`;
+    return await fetchApi(apiUrl);
+  }
+}
+
+export class CollectionsApi {
```
7emansell, line 199: The functions below aren't used anywhere, these are just 2 endpoints that exist in some form on QA Collections API rn

app/src/utils/collectionsApiClient.test.tsx
EdwinGuzman, line 1: The tests were just moved but no new tests were added, correct?

EdwinGuzman:
> Is updating fetchApi (or, probably, just writing a new one for Collections API calls) in scope for this?

I think it's in scope. It looks like everything will generally be the same except for the response format. It will no longer be `data?.nyplAPI?.response`.

EdwinGuzman:
Or, make `fetchAPI` more generic by returning `data` and let the callee handle destructing the response.

app/src/utils/apiClients.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+    return {
+      uuid: apiResponse?.capture?.uuid || defaultResponse.uuid,
+      title: apiResponse?.capture?.title || defaultResponse.title,
+      imageID: apiResponse?.capture?.imageID || defaultResponse.imageID,
+    };
+  }
+
+  /**
+   *
```
EdwinGuzman, line 85: I think it's time to clean this up as well and remove unnecessary comments.

EdwinGuzman (approved):
This works well.

This isn't a problem but it's interesting to see a class that only has static functions. It's possible to create an object with function properties that essentially perform the same actions, such as
```
export const RepoApi = {
  getHomePageData: () => { ... },
  ...
}
```

```
import { RepoApi } from ...;

const data = await RepoApi.getHomePageData({...
```

No need to change, just pointing this out. I do prefer, style-wise, the class approach.

avertrees (approved):

app/src/utils/fetchApi.ts
```diff
@@ -46,18 +53,19 @@ export const fetchApi = async (
   try {
     const response = (await fetchWithTimeout(apiUrl, {
       method,
-      headers,
-      body: method === "POST" ? JSON.stringify(options?.body) : undefined,
+      ...(isRepoApi ? { headers } : {}),
```
7emansell, line 56: this will be refined once Collections API decides what kind of authorization method it'll use

app/src/utils/apiClients.tsx
```diff
@@ -0,0 +1,262 @@
...
+   */
+
+  static async getItemData(uuid: string) {
+    const apiUrl = `${process.env.API_URL}/api/v2/items/mods_captures/${uuid}`;
+    const res = await fetchApi({ apiUrl: apiUrl });
```
EdwinGuzman, line 90: nit: you can use the short hand `{ apiUrl }`

EdwinGuzman (approved):
Approving the last commit. The `fetchApi` update is more verbose, but I like that it's more explicit considering it's also a lower-level function that the two classes abstract away.

PR #298, opened 2025-03-03
DR-3367: Active filters 

7emansell:
## Ticket:

- JIRA ticket [DR-3367](https://newyorkpubliclibrary.atlassian.net/browse/DR-3367)

- Creates `activeFilters` component with `TagSet` and adds it to `/search` and `/collections/[slug]`
- Adds/updates filter management in `searchManager`

Add and remove filters etc. Complete filter modal is not in this branch but dropdowns should work enough to show how the `TagSet` / `SearchManager`/ URL communication work

### Checklist:

[DR-3367]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/search/filters/selectFilter.tsx
+++ b/app/src/components/search/filters/selectFilter.tsx
@@ -14,6 +14,8 @@ import React, {
   useState,
 } from "react";
 import FilterAccordion from "./filterAccordion";
+import { usePathname, useRouter } from "next/navigation";
+import { useSearchContext } from "@/src/context/SearchProvider";
 
 export type FilterOption = {
   name: string;
@@ -36,6 +38,12 @@ const SelectFilterComponent = forwardRef<
   const { filter, ...rest } = props;
   const [userClickedOutside, setUserClickedOutside] = useState<boolean>(false);
   const [timeoutId, setTimeoutId] = useState<NodeJS.Timeout | null>(null);
+  const { push } = useRouter();
+  const pathname = usePathname();
+  const updateURL = async (queryString) => {
+    push(`${pathname}?${queryString}`);
+  };
+  const { searchManager } = useSearchContext();
 
   // Create a ref to hold a reference to the accordion button, enabling us
   // to programmatically focus it.
@@ -104,9 +112,16 @@ const SelectFilterComponent = forwardRef<
     const newTimeoutId = setTimeout(() => {
       console.log(`selected: ${newSelection}`);
       setUserClickedOutside(true);
-      (
-        headingRef as MutableRefObject<HTMLHeadingElement | null>
-      )?.current?.focus();
+      updateURL(
+        searchManager.handleAddFilter({
+          filter: filter.name,
+          value: newSelection,
+        })
+      );
+      //setUserClickedOutside(true);
+      // (
+      //   headingRef as MutableRefObject<HTMLHeadingElement | null>
+      // )?.current?.focus();
     }, 600);
 
     setTimeoutId(newTimeoutId);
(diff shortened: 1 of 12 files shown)


app/src/components/search/filters/selectFilter.tsx
```diff
@@ -104,9 +112,16 @@ const SelectFilterComponent = forwardRef<
...
-      )?.current?.focus();
+      updateURL(
+        searchManager.handleAddFilter({
+          filter: filter.name,
+          value: newSelection,
+        })
+      );
+      //setUserClickedOutside(true);
+      // (
+      //   headingRef as MutableRefObject<HTMLHeadingElement | null>
```
EdwinGuzman, line 123: ?

app/src/utils/searchManager.tsx
```diff
@@ -64,13 +64,24 @@ abstract class BaseSearchManager implements SearchManager {
...
-    const updatedFilters = [...this.currentFilters, newFilter];
+    const updatedFilters = this.currentFilters.map((filter) =>
+      filter.filter === newFilter.filter ? newFilter : filter
+    );
+
+    const filterExists = this.currentFilters.some(
+      (filter) => filter.filter === newFilter.filter
+    );
+
+    if (!filterExists) {
```
EdwinGuzman, line 75: Consider making `this.currentFilters` or maybe just updatedFilters a `Set` to not have to validate the value already exists before adding it. You can then convert it back to an array as `this.currentFilters = [...updatedFiltersSet];`

app/src/utils/searchManager.tsx
```diff
@@ -207,3 +218,15 @@ const createQueryStringFromObject = (object: Record<string, string>) => {
   });
   return params.toString();
 };
+
+export const stringToFilter = (filtersString: string | null): Filter[] => {
```
EdwinGuzman, line 222: Add some tests, especially for any functions that use regex.

app/src/utils/searchManager.tsx
```diff
@@ -207,3 +218,15 @@ const createQueryStringFromObject = (object: Record<string, string>) => {
...
+  if (!filtersString) return [];
+  return Array.from(filtersString.matchAll(/\[([^\]=]+)=([^\]]+)\]/g)).map(
+    ([, filter, value]) => ({ filter, value })
+  );
+};
+
+export const filterToString = (filters: Filter[]): string => {
```
EdwinGuzman, line 229: Question regarding both functions: did you try the next.js way of parameterizing values or the qs function? Just so you can use existing functions.

It also may depend a bit on how the URL and query params should look like which might not be possible with qs.

EdwinGuzman (commented):
The active filters looks good but left some comments. Maybe out of scope for this but, when I select a filter, the display is updated correctly but the dropdown does not have my selection highlighted. Should that work now or not yet until real data is used?

7emansell:
Dropdown filters not working as they should particularly in regard to focus management, feedback for that will be resolved in [this ancient PR](https://github.com/NYPL/digital-collections/pull/290), so will note but only respond to active filter feedback in this one

app/src/utils/searchManager.tsx
7emansell: I did initially start with `URLSearchParams` but if the only value is that I could be using existing functions I think I'll just accept the sunk cost of these until they fail me (which I suspect will be a lot easier to find out with real data instead of easily handled strings)?

EdwinGuzman: Yes, that's fair.

EdwinGuzman (approved):
Good update with the `Set` data structure. Hopefully that's easier to work with.

PR #300, opened 2025-03-04
DR-3460: DC accessible toggletip

7emansell:
## Ticket:

- JIRA ticket [DR-3460](https://newyorkpubliclibrary.atlassian.net/browse/DR-3460)

- Creates `toggleTip` component and uses it on the rights filter

Test hovering, clicking, and tabbing. It is expected that once click opens the toggle tip, only click can close it. Every time it's opened (by whichever method) it should add content to the live region below it, and remove the content once the toggle tip is closed (see DOM structure). 

- Needs a11y review from Clare on just general functioning

### Checklist:

[DR-3460]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/toggleTip/toggleTip.tsx
```diff
@@ -0,0 +1,107 @@
...
+          </Box>
+        )}
+      </Box>
+      {/* Live region hidden */}
+      <Box
+        as="span"
+        ref={liveRegionRef}
+        role="status"
+        aria-live="polite"
+        display="none"
```
EdwinGuzman, line 103: Instead of display none, set the text off screen.

Toxiapo:
@7emansell This was my first time learning about toggleTips so I want to make sure I understand the interaction correctly. It seems like from the article on the Jira ticket that the toggle tips should be dismissed when the user clicks on anywhere of the page or pressing escape, but the implementation here shows that I can only close it by clicking the icon, or maybe I am missing something? 

>Instead of clicking the button showing or hiding the bubble, it only shows it. Hiding the bubble is achieved by unfocusing the button, mouse clicking away from the button or pressing escape.

app/src/components/search/filters/filters.tsx
```diff
@@ -11,6 +11,7 @@ import SelectFilterGrid from "./selectFilterGrid";
 import RightsFilter from "./rightsFilter";
 import DateFilter from "./dateFilter";
 import { headerBreakpoints } from "@/src/utils/breakpoints";
+import { ToggleTip } from "../../toggleTip/toggleTip";
```
Toxiapo, line 14: unused import

Toxiapo (commented):
I just had that one question regarding the interaction, otherwise looks great!

app/src/components/toggleTip/toggleTip.tsx
```diff
@@ -0,0 +1,142 @@
...
+          height="unset"
+          minWidth="unset"
+          _hover={{
+            background: "unset",
+            outline: "2px solid var(--nypl-colors-ui-link-primary)",
+          }}
+          onClick={() => {
+            setIsClickVisible((prev) => !prev);
+            setIsHoverFocusVisible(false);
+            if (!isClickVisible) {
```
EdwinGuzman, line 52: There's a risk of not using the right value for `isClickVisible` since it's updated two lines before and it might not be updated once this line is called. I don't see any issues while testing though but just pointing this out.

app/src/components/toggleTip/toggleTip.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const button = screen.getByRole("button");
+
+    fireEvent.focus(button);
+    expect(screen.getByText(toggleTipContent)).toBeInTheDocument();
+
+    fireEvent.blur(button);
+    expect(screen.queryByText(toggleTipContent)).not.toBeInTheDocument();
+  });
+
+  it("toggles tooltip on click and requires another click to close", async () => {
```
EdwinGuzman, line 37: If this is strictly following the link from Clare, it should only open when it's clicked and close when it's clicked outside.

EdwinGuzman (commented):
I'm interested to see what Clare says about this. Functionally, it's working well and I only have the concern about following the inclusive tooltip link regarding opening when click and closing only when clicking outside.

I tried this using mac vo and I don't see or hear the live region text. Generally, it's confusing because the label text itself isn't being read so it might be a setting I have? Only "More info" is read:

https://github.com/user-attachments/assets/d3a03ee7-fbb5-4164-91f5-9cb16a6b31e7

7emansell:
Thank you guys– so, based on your feedback and Clare's, here's what the toggle tip now is expected to do:
1. Click opens
2. Tab opens
3. Click elsewhere closes
4. Clicking again closes
5. Tabbing away closes
6. Esc closes

No more hover at all since then it needs to meet [these criteria](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html) which will change how it looks I think cc @EdwinGuzman @Toxiapo

app/src/components/toggleTip/toggleTip.tsx
```diff
@@ -0,0 +1,161 @@
...
+          onClick={() => {
+            setIsVisible((prev) => !prev);
+            if (!isVisible) {
+              addLiveRegion();
+            } else {
+              removeLiveRegion();
+            }
+          }}
+          onFocus={() => {
+            if (!isVisible) {
```
EdwinGuzman, line 95: I don't think you need this conditional here. It will always be false before the button is focused and you always want to set it visible on focus, right?

EdwinGuzman: However, from Clare's feedback, it shouldn't be visible on focus but rather when the user presses enter or space (which does work to close it).

EdwinGuzman (commented):
Just the minor update but otherwise, the component's code looks good!

7emansell:
Okay, so re. Clare, now it no longer opens on focus, so:

1. Click opens
2. Space/enter opens
5. Clicking again closes
6. Space/enter closes
7. Click elsewhere (loss of focus) closes
8. Tabbing away (loss of focus) closes
9. Esc closes
 
cc @EdwinGuzman

EdwinGuzman (approved):
Nice work!

EdwinGuzman:
Passed accessibility and vqa

PR #302, opened 2025-03-06
DR-3416: Connect search page to API

7emansell:

- Fleshes out mock search and collection response to match schemas as of 3/7
- Integrates search functionality from server component down to client filter components

## Open Questions

- See comments for areas that I think will need specific work (more than just plugging in data)

### Checklist:

[DR-3464]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DR-3416]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DR-3362]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/collectionMetadata/collectionMetadata.tsx
```diff
@@ -0,0 +1,199 @@
...
+  Heading,
+  ButtonGroup,
+  Button,
+  Link,
+  Text,
+  Box,
+} from "@nypl/design-system-react-components";
+import { useState } from "react";
+
+type CollectionMetadataProps = {
```
7emansell, line 12: This and `collectionsUUIDSchema` will need to be updated once we have the list of all metadata fields being returned. this file has some initial formatting but this should definitely be abstracted out instead of hard set for each field

app/src/components/search/filters/dateFilter.tsx
```diff
@@ -1,70 +1,113 @@
...
+
+const DateFilter = forwardRef<HTMLHeadingElement, DateFilterProps>(
+  ({ searchManager }, ref) => {
+    const { push } = useRouter();
+    const pathname = usePathname();
+    const updateURL = async (queryString) => {
+      push(`${pathname}?${queryString}`);
+    };
+
+    let startValue = searchManager.filters.find((f) => f.filter === "dateStart")
```
7emansell, line 26: this mix of setting the filter to the value that's currently in `searchManager` on mount and then managing it with state is a little confusing– there well might be another way to do this. This is the same pattern used in `selectFilter` and `rightsFilter`, so it'd need to be replaced in all 3 places

app/search/index/page.tsx
```diff
@@ -28,12 +31,22 @@ export type SearchProps = {
 export default async function Search({ searchParams }: SearchProps) {
   const pageName = searchParams.keywords ? "search-results" : "all-items";
 
-  const data = await CollectionsApi.getSearchData({
-    keyword: searchParams.keywords,
-    sort: searchParams.sort,
-    filters: searchParams.filters,
-    page: searchParams.page,
-  });
+  const searchResults =
```
7emansell, line 34: once the search endpoint is complete, my hope is that you can pretty much just uncomment this and then work out the bugs from there. Same on `/collections/[slug]`

app/src/components/collectionStructure/collectionStructure.tsx
7emansell, line 1: this whole prefetching situation can be ignored for now but the response from the `collections/:uuid/children` endpoint should align with what's happening here

PR #313, opened 2025-04-02
DR-3529: Search redirects

7emansell:
## Ticket:

- JIRA ticket [DR-3529](https://newyorkpubliclibrary.atlassian.net/browse/DR-3529)

- Sets up middleware for search query redirects from OG to Facelift
- Writes tests

## Open Questions

Should I specify this to the `search/index` path?

Try some searches and copy/paste them over to Facelift. Also, see [this mapping and list of examples](https://docs.google.com/document/d/1_HDSzW5ToWf6ariAm7G0VWoUTvPBCqbq7q-9gInqQiM/edit?tab=t.0#heading=h.1ku3d96grql6). If some redirects are unintuitive (specifically `date`) check this out first.

Also, this PR does NOT have the most recent search/filtering updates, so do not expect the filter panel to be correctly populated– check the URL only. 

### Checklist:

[DR-3529]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/middleware/searchRedirects.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+});
+
+it("drops ascending/descending relevance sort (default sort)", () => {
+  const request = {
+    nextUrl: new URL("http://localhost/search/index?sort=score+desc"),
+  } as NextRequest;
+  const response = middleware(request);
+
+  expect(NextResponse.redirect).toHaveBeenCalledWith(
+    "http://localhost/search/index?",
```
EdwinGuzman, line 36: You should drop the `?` as well

__tests__/middleware/searchRedirects.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+    301
+  );
+  expect(response).toBe("redirect response");
+});
+
+it("drops sequence sort", () => {
+  const request = {
+    nextUrl: new URL("http://localhost/search/index?sort=sortString+desc"),
```
EdwinGuzman, line 44: I see both `sortString_sort+desc` and `sortString+desc` in the doc. Can you also include a test for `sortString_sort` if it's needed?

__tests__/middleware/searchRedirects.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+it("transforms many params", () => {
+  const request = {
+    nextUrl: new URL(
+      "http://localhost/search/index?filters%5Bdate%5D%5B%5D=1900-9999&filters%5Bgenre%5D%5B%5D=Cards&filters%5Bname%5D%5B%5D=Ogden%27s+Cigarettes&filters%5Brights%5D=pd&keywords="
+    ),
+  } as NextRequest;
+  const response = middleware(request);
+
+  expect(NextResponse.redirect).toHaveBeenCalledWith(
+    "http://localhost/search/index?q=&filters=%5BdateStart%3D1900%5D%5Bgenre%3DCards%5D%5Bname%3DOgden%27s+Cigarettes%5D%5Brights%3DpublicDomain%5D",
```
EdwinGuzman, line 254: When I try the url above in the vercel preview, I get this string except without the `q=`. I've noticed this for other cases in the table in the doc as well; `q=` is not added when `keywords=` is in the original url

EdwinGuzman (commented):
Left some comments but a question before I continue reviewing. In the doc, the mapping for `date=x-y` is `filters=[dateStart=x&dateEnd=y]`. But locally and in vercel the URL updates to `/search/index?filters=[dateStart=1900][dateEnd=1905]`. This is also in a test and is true for all other filters. Should each key/value pair be wrapped in brackets or should they be separated by an ampersand?

7emansell:
> Left some comments but a question before I continue reviewing. In the doc, the mapping for `date=x-y` is `filters=[dateStart=x&dateEnd=y]`. But locally and in vercel the URL updates to `/search/index?filters=[dateStart=1900][dateEnd=1905]`. This is also in a test and is true for all other filters. Should each key/value pair be wrapped in brackets or should they be separated by an ampersand?

Sorry for the inconsistencies– updated now. it should be `[dateStart=1900][dateEnd=1905]`, as for all other filters. The ampersand is a little sketchy since so many filter names actually include an ampersand, I think brackets are the clearest/safest bet

__tests__/middleware/searchRedirects.test.tsx
7emansell: `sortString_sort` is a typo, now removed

__tests__/middleware/searchRedirects.test.tsx
7emansell: this is a little weird bc if `keywords` is empty, then `q` will be empty, and `searchManager` is set up to drop it from the URL, the same way we don't include `sort=relevance`. The unit tests just don't arrive at that point. 
But it's still inconsistent locally, so I updated it to absolutely remove `q=` if there are no keywords, and updated the test cases and doc to reflect that. 
Same thing with the `?` at the end when all params have been dropped, it does eventually get removed.

__tests__/middleware/searchRedirects.test.tsx
7emansell: https://github.com/NYPL/digital-collections/pull/313#discussion_r2029227023

__tests__/middleware/searchRedirects.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const request = {
+    nextUrl: new URL("http://localhost/search/index?keywords=test"),
+  } as NextRequest;
+  const response = middleware(request);
+
+  expect(NextResponse.redirect).toHaveBeenCalledWith(
+    "http://localhost/search/index?q=test",
+    301
+  );
+  expect(response).toBe("redirect response");
```
avertrees, line 26: do these tests just check whether or not the middleware triggers a redirect? should we have tests for the string manipulation to ensure the urls are as expected or is that overkill?

7emansell: I would say all of these tests do check that the urls are as expected, but there are cases where we continue to manipulate the URL to ensure a more readable search query (mostly by dropping defaults like `sort=relevance`). That's all in the `searchManager`. Not overkill to check those, but it should happen in `searchManager.test.tsx` so I'll add those in DR-3416

avertrees: sounds great

avertrees (approved):

PR #314, opened 2025-04-03
DR-3535: Add Collections API authentication

7emansell:
## Ticket:

- JIRA ticket [DR-3535](https://newyorkpubliclibrary.atlassian.net/browse/DR-3535)

- Updates `fetchApi` to pass corresponding token (previously only passed token on Repo API calls)
   - Adds a test for this ^
- Adds `COLLECTIONS_API_AUTH_TOKEN` and `QA_COLLECTIONS_API_AUTH_TOKEN` to environment and environment documentation
- Adds build variables to Github Actions scripts
  - Also adds as secrets to the repo
- Adds tokens as environment variables in Vercel

Add these to your own `.env`! There should be no apparent changes (especially bc authentication is not set up on Collections API yet)

### Checklist:

[DR-3535]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ENVIRONMENTVARS.md
```diff
@@ -19,16 +19,18 @@ If an environment variable is updated, make sure to restart the server for the a
...
+| `APP_ENV`                    | string | "development"                                                               | App environment key used to determine various environment-specific app settings. |
+| `API_URL`                    | string | "https://api.repo.nypl.org" or "https://qa.api.repo.nypl.org"               | Base url for Repo API.                                                           |
+| `AUTH_TOKEN`                 | string | ""                                                                          | Auth token for Repo API.                                                         |
+| `COLLECTIONS_API_URL`        | string | "https://qa-api-collections.nypl.org" or "https://api-collections.nypl.org" | Base url for Collections API.                                                    |
+| `COLLECTIONS_API_AUTH_TOKEN` | string | ""                                                                          | Auth token for Collections API.                                                  |
```
EdwinGuzman, line 28: Can you add that this is set in the repo's settings? To differentiate and know the source compared to terraform or task definitions.

app/src/utils/fetchApi.ts
```diff
@@ -23,8 +23,12 @@ export const fetchApi = async ({
     isRepoApi?: boolean;
   };
 }) => {
-  const apiKey = process.env.AUTH_TOKEN;
   const { method = "GET", params, body, isRepoApi = true } = options;
+  const apiKey = isRepoApi
+    ? process.env.AUTH_TOKEN
+    : process.env.NODE_ENV === "production"
```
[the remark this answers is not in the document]
EdwinGuzman: Right, not ideal. I think you should only have `COLLECTIONS_API_AUTH_TOKEN` as the env var and update it appropriately for each environment. You'd have to remember to update this locally to test either qa or prod, but it follows the same approach as `AUTH_TOKEN`. When the app gets built in docker, it's only really using `COLLECTIONS_API_AUTH_TOKEN` as well (`QA_COLLECTIONS_API_AUTH_TOKEN` is only read by the github action script and not the app).

EdwinGuzman (changes requested):
I think you do need two values set in github for the qa and prod values, but you can use only `COLLECTIONS_API_AUTH_TOKEN` in the repo.

More importantly, this is set up to pass the value from gha to docker, but docker isn't reading or setting it. Check the `Dockerfile` and read and set the env var there (similar to `AUTH_TOKEN`, for example).

app/src/utils/fetchApi.ts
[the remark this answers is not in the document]
7emansell: Ah i knew i must be forgetting it somewhere

7emansell:
@EdwinGuzman gotcha. So to be clear– as i have it now there's no specification in the repo but just in my/your personal `.env` i would have `COLLECTIONS_API_AUTH_TOKEN` set to the qa token. And then only the QA github workflow will use the QA token stored as a secret?

ENVIRONMENTVARS.md
7emansell: should I note that for all of the values set on the repo and used in Github actions?

EdwinGuzman: yes, that'd be helpful as part of this pr

EdwinGuzman:
>  So to be clear– as i have it now there's no specification in the repo but just in my/your personal .env i would have COLLECTIONS_API_AUTH_TOKEN set to the qa token. And then only the QA github workflow will use the QA token stored as a secret?

Yes, that's correct.

7emansell:
Have mentioned before but I don't think it makes sense to keep `API_URL` as a repo secret or as an arg in our Github actions.. not sure if this Pr is the place to discuss but just want to flag that cc @avertrees @EdwinGuzman

.github/workflows/deploy_qa.yml
```diff
@@ -37,9 +37,10 @@ jobs:
           NEW_RELIC_LICENSE_KEY: ${{ secrets.NEW_RELIC_LICENSE_KEY }}
           API_URL: ${{ secrets.API_URL }}
           AUTH_TOKEN: ${{ secrets.AUTH_TOKEN }}
+          COLLECTIONS_API_AUTH_TOKEN: ${{ secrets.QA_COLLECTIONS_API_AUTH_TOKEN }}
```
emu47, line 40: I think this should just be COLLECTIONS_API_AUTH_TOKEN again to keep the pattern the same. Generally I think the values changes and location of the values change but the variable names are the same?

7emansell: on qa this would need to be the qa token though is that correct? This is referencing the name of the secret in the facelift repo, not the environment variable

avertrees:
> Have mentioned before but I don't think it makes sense to keep `API_URL` as a repo secret or as an arg in our Github actions.. not sure if this Pr is the place to discuss but just want to flag that cc @avertrees @EdwinGuzman

@7emansell I'm a bit lost on whether or not we still need the API_URL as a repo secret. I think we should deff remove it if we don't need it, but would like if that work was separate from this PR.

.github/workflows/deploy_qa.yml
avertrees: Right. The environment variable is `COLLECTIONS_API_AUTH_TOKEN` in both environments but the name of the secret in the repo for qa is `QA_COLLECTIONS_API_AUTH_TOKEN` and for prod it's `COLLECTIONS_API_AUTH_TOKEN`. We need a secret for each environment and a way to differentiate the two. If anything `COLLECTIONS_API_AUTH_TOKEN` could be `PROD_COLLECTIONS_API_AUTH_TOKEN` but I think this is fine as it is.

7emansell:
@avertrees it just doesn't need to be a secret imo– but yes shouldn't happen as part of this work

EdwinGuzman (approved):
Approving but get more approvals before merging.

ENVIRONMENTVARS.md
```diff
@@ -19,16 +19,19 @@ If an environment variable is updated, make sure to restart the server for the a
...
+| `AUTH_TOKEN`                     | string | ""                                                                          | Auth token for Repo API. Set as a Github repo secret, used in Github action workflows (deploy QA, prod).                                                        |
+| `COLLECTIONS_API_URL`            | string | "https://qa-api-collections.nypl.org" or "https://api-collections.nypl.org" | Base url for Collections API.                                                                                                                                   |
+| `COLLECTIONS_API_AUTH_TOKEN`     | string | ""                                                                          | Auth token for Collections API. Set as a Github repo secret, used in Github action workflow (deploy prod).                                                      |
+| `QA_ COLLECTIONS_API_AUTH_TOKEN` | string | ""                                                                          | Auth token for QA Collections API. Set as a Github repo secret, used in Github action workflow (deploy QA).                                                     |
```
emu47, line 29: delete extra space between QA_ and COLLECTIONS

.github/workflows/deploy_qa.yml
emu47: Oh wait a second. Why are we doing variables like this? These shouldn't be set by gha, they should be using the variables setup in the task definition on aws, no? I mean, feel free to overrule me on this one. Don't mean to rock the boat here, but ... I'm not sold on this approach.

7emansell: cc @avertrees @EdwinGuzman

avertrees (approved):

.github/workflows/deploy_qa.yml
EdwinGuzman: There should be a ticket in the backlog about how next creates builds. If you take a look at the docker builds, next tries to create static pages by calling the API and generating the page. This should be turned off. So for now, the tokens are needed at build time. I do agree it's a separate ticket but it's not maintenance.

It should roughly be two tickets to:
- turn of static build generation in nextjs
- move token values to terraform and remove from github secrets and gha.

chrisb87: Yes I think the `COLLECTIONS_API_AUTH_TOKEN` isn't needed during build time, so not needed by github actions and not needed in the Dockerfile. But it will be needed by vercel, is that piece done?

chrisb87 (approved):

PR #318, opened 2025-04-15
DR-3416: Search page pt. 2

7emansell:
## Ticket:

- JIRA ticket [DR-3416](https://newyorkpubliclibrary.atlassian.net/browse/DR-3416)

- Resolves a11y, VQA, and functional issues with the search page that came out of [this review](https://docs.google.com/document/d/1MwhKG6YRd1yMerWtpL3-3BLxtVDl2biuNv9fwEj67Uo/edit?tab=t.0) and search testing session
- Notably adds focus management on the filters with a `lastFilterRef` now stored in the `searchManager`

### Checklist:

[DR-3416]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/search/filters/dateFilter.tsx
EdwinGuzman, line 1: Do you think the input fields should have a placeholder?

I think the form should submit by pressing enter on the "end year" input field.

app/src/components/search/filters/filters.tsx
```diff
@@ -15,38 +15,48 @@ import { SearchManager } from "@/src/utils/searchManager";
...
+  setFiltersExpanded,
+}: FiltersProps) => {
   const dateFilterRef = useRef<TextInputRefType | null>(null);
   const secondRowFilter = 4;
   useEffect(() => {
-    if (isExpanded) {
-      if (filterRefs.current[secondRowFilter]) {
-        filterRefs.current[secondRowFilter]?.focus();
+    if (filtersExpanded) {
+      // If there are 4> dropdown filters visible, focus the 4th.
```
EdwinGuzman, line 32: nit: it's 4th index but 5th dropdown.

app/src/components/search/filters/dateFilter.tsx
EdwinGuzman: The entire component should be wrapped in a form to submit on enter press.

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -49,14 +50,35 @@ const SearchPage = ({
...
+      const button = document.querySelector(
+        `button[aria-label="${searchManager.lastFilterRef?.current}"]`
+      );
+      if (button) {
+        (button as HTMLButtonElement).focus();
+      }
+      const input = document.querySelector(
+        `input[value="${searchManager.lastFilterRef?.current}"]`
+      );
+      if (input) {
```
EdwinGuzman, line 77: I guess there won't ever be a case when you focus on the button and then focus on the input element, right? It's either focus on the button or focus on the input?

app/src/components/search/filters/selectFilter.tsx
```diff
@@ -203,8 +200,12 @@ const SelectFilterComponent = forwardRef<
...
           {
             accordionType: "default",
-            buttonInteractionRef: mergedRef,
-            label: capitalize(filter.name),
+            buttonInteractionRef: accordionButtonRef,
+            label: (
+              <Box noOfLines={1}>{`${capitalize(filter.name)}${
+                selected ? `: ${capitalize(selected.name)}` : ``
+              }`}</Box>
+            ),
```
EdwinGuzman, line 208: As discussed in person, you can refactor the `FilterAccordionComponent` so that it only every takes one object value since it won't ever render more than one for DCF.

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -49,14 +50,35 @@ const SearchPage = ({
...
   useEffect(() => {
-    if (isFirstLoad.current) {
+    setIsLoaded(true);
+    if (
+      searchManager.lastFilterRef?.current &&
+      (searchManager.filters.length > 0 || searchManager.sort)
+    ) {
+      // Search for the button or input associated with the last used filter/sort
+      const button = document.querySelector(
+        `button[aria-label="${searchManager.lastFilterRef?.current}"]`
```
EdwinGuzman, line 69: This works well. You're righ that it's brittle. Can you try it this with ids? I don't imagine the value name will change but if you set an id for each filter every time it's updated, it might make it more stable. B

app/src/components/pages/searchPage/searchPage.tsx
7emansell: Yes only one of the (now three) options

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -127,6 +144,12 @@ const SearchPage = ({
               `}
             </Heading>
           )}
+          <Filters
```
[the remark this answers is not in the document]
7emansell: VQA approved this as the no results view– I was pro removing the panel altogether but was overruled

app/src/components/sortMenu/sortMenu.tsx
```diff
@@ -17,6 +24,9 @@ const SortMenu = ({ updateURL, searchManager, options }: SortMenuProps) => {
         id,
         label,
         onClick: () => {
+          if (setFiltersExpanded) {
+            setFiltersExpanded(false);
```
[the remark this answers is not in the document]
7emansell: VQA request that filters collapse on sort and paginate

app/src/utils/searchManager.tsx
```diff
@@ -265,7 +275,9 @@ export const transformToAvailableFilters = (
   availableFilters: Record<string, AvailableFilterOption[]>
 ): AvailableFilter[] => {
   return Object.entries(availableFilters)
-    .filter(([key]) => key !== "subcollection")
+    .filter(
+      ([key]) => key !== "subcollection" && key !== "language" && key !== "form"
```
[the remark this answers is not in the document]
7emansell: We don't display some of the available filters but they're still returned

app/src/utils/searchManager.tsx
```diff
@@ -56,6 +61,7 @@ abstract class BaseSearchManager implements SearchManager {
     this.currentAvailableFilters = transformToAvailableFilters(
       config.initialAvailableFilters ?? {}
     );
+    this.lastFilterRef = config.lastFilterRef!!;
```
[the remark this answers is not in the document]
7emansell: Typo, this is just a non null check

app/src/data/schemas/searchSchema.ts
```diff
@@ -17,10 +17,12 @@ const searchSchema = {
       {
         name: "string",
         count: "integer",
+        selected: "boolean",
```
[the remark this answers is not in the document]
7emansell: Yes– and like a step further into this mental model is that *only* the backend knows what value was selected, and is just telling the filter. For a long time, I was mocking this with state

app/src/components/card/searchCardImage.tsx
```diff
@@ -0,0 +1,52 @@
+import Image from "next/image";
+import React from "react";
+import { useState } from "react";
+import { Box } from "@nypl/design-system-react-components";
+
+export const SearchCardImage = ({ record }) => {
```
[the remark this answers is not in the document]
7emansell: There wasn't– there was `CardImage` but the search card image requires different dimensions/handling, and previously I was just using the DS Image

avertrees (approved):

7emansell:
A11y cleared by Clare, VQA cleared by Lily

app/src/components/pages/searchPage/searchPage.tsx
```diff
@@ -186,7 +189,9 @@ const SearchPage = ({
               <Heading
                 size="heading5"
                 ref={headingRef}
-                role="status"
+                aria-live="polite"
+                // @ts-ignore
```
EdwinGuzman, line 193: Why?

This is interesting. You don't need the tabindex attr if there's aria-live, and vice versa. So this is because sometimes you explicitly focus on the heading and other times you do not?

7emansell: Yes. There's a type error otherwise, which I assume is because `Heading` doesn't extend `tabIndex`?

EdwinGuzman (approved):

app/src/components/pages/searchPage/searchPage.tsx
EdwinGuzman: it's because it should be a number and not a string.

app/src/components/search/filters/activeFilters.tsx
```diff
@@ -48,8 +51,14 @@ const ActiveFilters = ({ searchManager }: ActiveFilterProps) => {
 
   return searchManager.filters.length > 0 ? (
     <>
-      <Flex alignContent="center" alignItems="center" gap="xs" flexDir="row">
-        <Text size="subtitle2" sx={{ margin: 0, fontWeight: 400 }}>
+      <Flex alignItems="flex-start" flexWrap="wrap" gap="xs" width="100%">
+        <Text
+          size="subtitle2"
+          // @ts-ignore
+          tabIndex={-1}
```
EdwinGuzman, line 58: In this case, `Text`'s prop does not extend Chakra's HTML props.

PR #321, opened 2025-04-24
No ref: Search filter redirects

7emansell:

- Handles _filter_ redirects within the search query, so that links to filters from the item page (ex. [item with physical description link](https://digitalcollections.nypl.org/items/c7438e90-ebed-013d-3ecf-0242ac110002)) as well as bookmarked filters on a search query will work
- See [this doc](https://docs.google.com/document/d/1_HDSzW5ToWf6ariAm7G0VWoUTvPBCqbq7q-9gInqQiM/edit?tab=t.0#heading=h.1ku3d96grql6) again for all the expected redirects and example URLs
- Lana has confirmed that for filters that do not appear as dropdowns on our new search page (namely physical description/form and language) the tag appearing in the "Applied filters:" is sufficient to indicate that you are filtering with that value, even if the dropdown isn't visible/selected
- Similar work will need to be done with `subcollection` but I'm considering that part of the collection landing page work.. I think (`collections/[someCollectionSlug]/?tab=navigation&roots=[someSubcollectionUuid]` needs to go to `collections/[someCollectionUuid]?filters=[subcollection=someSubcollectionTitle||Uuid]`)

## Open Questions

- In order to transform `root-collection = uuid` to a Collections API-acceptable `collection = title || uuid`, we have to request the collection title, making this middleware async. The `collections/:uuid` endpoint is so fast that I think this is worth it... what do you guys think

### Checklist:

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -2,6 +2,17 @@ import collectionSlugToUuidMapping from "@/src/data/collectionSlugUuidMapping";
 import { deSlugify } from "@/src/utils/utils";
 import { NextRequest, NextResponse } from "next/server";
 
+const filterMap = {
+  placeTerm_mtxt_s: "place",
+  genre_mtxt_s: "genre",
+  typeOfResource_mtxt_s: "type",
+  publisher_mtxt_s: "publisher",
+  namePart_mtxt_s: "name",
+  topic_mtxt_s: "topic",
+  languageTerm_mtxt_s: "language",
+  form_mtxt_s: "form",
+};
+
 async function getCollectionTitle(uuid): Promise<string> {
   try {
     const response = await fetch(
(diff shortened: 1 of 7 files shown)


middleware.ts
```diff
@@ -2,7 +2,27 @@ import collectionSlugToUuidMapping from "@/src/data/collectionSlugUuidMapping";
 import { deSlugify } from "@/src/utils/utils";
 import { NextRequest, NextResponse } from "next/server";
 
-export function middleware(req: NextRequest) {
+async function getCollectionTitle(uuid): Promise<string> {
```
7emansell, line 5: Necessary to duplicate `fetchApi` here because our logger doesn't work in edge runtime, and I don't want to remove the logs from all our requests

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
```
7emansell, line 3: Allows iteration over search params, weirdly

middleware.ts
```diff
@@ -2,7 +2,27 @@ import collectionSlugToUuidMapping from "@/src/data/collectionSlugUuidMapping";
...
+  try {
+    const response = await fetch(
```
avertrees, line 7: we talked about this and are in agreement that it's not ideal because this makes the middleware file async but is a faster solution to the problem than requesting an update to the collections filter. I'm ok with this for now

avertrees (approved):

middleware.ts
```diff
@@ -92,14 +112,51 @@ export function middleware(req: NextRequest) {
...
-  searchParams.forEach((value, key) => {
+  for (const [key, value] of searchParams.entries()) {
     const match = key.match(/^filters\[(.*?)\](?:\[\])?$/);
     if (match) {
       let filterKey = match[1];
       let filterValue = decodeURIComponent(value);
       if (filterKey === "rights" && filterValue === "pd") {
         filterValue = "publicDomain";
       }
+      switch (filterKey) {
```
EdwinGuzman, line 123: Would this work if you created a map lookup such as
```
const map = {
  placeTerm_mtxt_s: "place",
  genre_mtxt_s: "genre",
  typeOfResource_mtxt_s: "type",
  publisher_mtxt_s: "publisher",
  namePart_mtxt_s: "name",
  topic_mtxt_s: "topic",
  languageTerm_mtxt_s: "language",
  form_mtxt_s: "form",
};
```

and keep the additional logic updates? E.g.
```
if (filterKey === "rights" && filterValue === "pd") {
  filterValue = "publicDomain";
} else if (filterKey === "root-collection") {
  filterKey = "collection";
  filterValue = await getCollectionTitle(filterValue);
} else if (map[filterKey]) {
  filterKey = map[filterKey];
}
```

just an idea to group the filter key and their mappings together.

middleware.ts
EdwinGuzman: The console.error will still work. Any change that `collectionData.title` will be empty? If so, can you log that?

EdwinGuzman (commented):
Only comment is to try `esnext`. I'm not worried about making the middleware async for fetches, and the nextjs docs doesn't warn against making it async for this scenario.

middleware.ts
7emansell: Hmm could but not really an expected error– if `collectionData` returns anything, there'll be a title

EdwinGuzman: Ok, i would assume every collection always has a title.

EdwinGuzman (approved):

PR #326, opened 2025-05-01
DR-3598: Filter count display bug

7emansell:
## Ticket:

- JIRA ticket [DR-3598](https://newyorkpubliclibrary.atlassian.net/browse/DR-3598)

- If a dropdown filter has a selection, it does not display any of the other options
- This is because those other filter options (and their corresponding counts) are pretty much meaningless in the context of a search query that enforces one value per filter category. To get the filter option counts as if the current filter selection is not applied (an OR filter) we would need to revamp the backend of search which we are not gonna do rn

### Checklist:

[DR-3598]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed filters to be encoded as URI components (DR-3564)
+- Fixed dropdown filters to only display selection (DR-3598)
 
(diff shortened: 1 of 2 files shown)


yossariano (approved):
Seems simple enough!

PR #334, opened 2025-05-16
Hot fix: highlight adjustment

7emansell:

- Post search release, realized highlights are not returning full title field
- Updates the construction of the title on the search card to use full title field + highlights instead of just highlights

### Checklist:


7emansell:
kind of don't want to make a new version soooo

PR #348, opened 2025-06-05
create component for available on site materials

avertrees:
## Ticket:

- JIRA ticket [Conditionally render media viewer based on rights logic](https://newyorkpubliclibrary.atlassian.net/browse/DR-3645)

- builds a new component `ItemMediaViewerFallback` to handle what should render in place of the Universal Viewer when an Item is restricted OR when a manifest has metadata and no items in the items array.

## Open Questions

- divisions links.... currently I'm using the division information returned in the metadata under Library Locations but I'm pretty sure this links to a division in DCFL and not the division link on nypl.org.... Lana wants the banner to include the link to the division on the nypl.org site. I didn't include this logic in this PR but think I'll need to make an object in DCFL that maps DCFL slugs to NYPL Division urls.

locally with this uuid: 33e3aab0-16fc-013d-c413-0242ac110002

-

### Checklist:

### DIFF
--- a/app/src/components/items/viewer/fallback.tsx
+++ b/app/src/components/items/viewer/fallback.tsx
@@ -16,6 +16,8 @@ interface ItemProps {
 }
 
 const ItemMediaViewerFallback = ({ item }: ItemProps) => {
+  // hacky way to generate the link
+  // this assumes that the first location in the array is the Division link
   const divisionLink = item?.metadata?.locations?.split("<br>")[0];
   return (
     <>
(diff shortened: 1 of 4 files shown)


app/src/components/items/viewer/fallback.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Link,
+  Icon,
+} from "@nypl/design-system-react-components";
+
+interface ItemProps {
+  item: ItemModel;
+}
+
+const ItemMediaViewerFallback = ({ item }: ItemProps) => {
+  const divisionLink = item?.metadata?.locations?.split("<br>")[0];
```
sarangj, line 19: could we add a comment with that assumption, that the first location entry has to be the division link?

sarangj (approved):
Seems reasonable - thanks for calling out that assumption about the division link. Could we add a comment in the code stating that, just to make clear why we're splitting the string like that?

PR #356, opened 2025-06-11
vqa fixes and query params

avertrees:

- conditionally render "View this item elsewhere" links based on catalog and archives `identifiers` returned in manifest
- clean up some citation info
- add logic to conditionally render "order print" button based on `Content Type` ( we only want to render it for items... might conditionally render with rights info too)
- the `Order Print` button requires us to implement logic that gives us an accurate image id based on what the canvasIndex is in the viewer... this requires us to think about pagination with the UV. Pagination has been de-prioritized for now until we have a way to get a canvasIndex from a uuid to support the backwards compatibility of OGDC urls. Until then, I have commented out the `Order Print` button.
- add support for a `canvasIndex` query param on the Items page and start working through how to support using that to link directly to an item
- update middleware to support   `type`, `cv`,  and `uuid`  params just incase

### Checklist:

### DIFF
--- a/playwright/pages/dc_homepage.ts
+++ b/playwright/pages/dc_homepage.ts
@@ -34,6 +34,24 @@ export class DCHomepage {
   readonly nyplDigitalCollectionsApi: Locator;
   readonly digitalPublicLibraryOfAmerica: Locator;
 
+  //footer
+
+  readonly footerAccessibilityLink: Locator;
+  readonly footerPressLink: Locator;
+  readonly footerCareersLink: Locator;
+  readonly footerAboutNyplLink: Locator;
+  readonly footerSpaceRentalLink: Locator;
+  readonly footerPrivacyPolicyLink: Locator;
+  readonly footerOtherPoliciesLink: Locator;
+  readonly footerTermsAndConditionsLink: Locator;
+  readonly footerGovernanceLink: Locator;
+  readonly footerRulesAndRegulationsLink: Locator;
+  readonly footerLanguage: Locator;
+  readonly footerSocialFacebook: Locator;
+  readonly footerSocialTwitter: Locator;
+  readonly footerSocialInstagram: Locator;
+  readonly footerSocialYouTube: Locator;
+
   constructor(page: Page) {
     //navigation menu
     this.page = page;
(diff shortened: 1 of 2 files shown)


app/src/components/items/overview/external/overview.tsx
```diff
@@ -1,34 +1,13 @@
...
-  typeOfResource: "Type Of Resource",
+export const parseHtmlString = (html) => {
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(html, "text/html");
+  const anchor = doc.querySelector("a");
+  return {
+    href: anchor?.getAttribute("href") || undefined,
+    text: anchor?.textContent || undefined,
+  };
 };
```
sarangj, line 11: Is there any way we could share this function somewhere (I think I noticed a third version of it elsewhere). Doesn't have to be now, could just be a backlog ticket / TODO

app/src/components/items/uv/universalViewer.tsx
```diff
@@ -17,13 +17,15 @@ export type UniversalViewerProps = {
...
+    // console.log("canvasIndex is: ", canvasIndex)
+
     console.log("canvasIndex in UniversalViewer component is: ", canvasIndex);
     const ref = useRef<HTMLDivElement>(null);
     const lastIndex = useRef<number>();
     const options = useMemo(
       () => ({
         manifest: manifestId,
-        canvasIndex: canvasIndex || 0,
+        canvasIndex: 2, //canvasIndex || 0,
```
sarangj, line 28: Is this on purpose? If so, could you add a comment why?

app/src/components/items/viewer/fallback.tsx
```diff
@@ -18,8 +19,8 @@ interface ItemProps {
 const ItemMediaViewerFallback = ({ item }: ItemProps) => {
   // hacky way to generate the link
   // this assumes that the first Library Location in the array is the Division link
-  // TODO: map DCFL Division link to nypl.org Division link
-  const divisionLink = item?.metadata?.locations?.split("<br>")[0];
+  // const divisionLink =
+  //   item.divisionLink ? item.divisionLink : item?.metadata?.locations?.split("<br>")[0];
```
sarangj, line 23: can we just remove the commented out code?

app/src/models/item.ts
```diff
@@ -46,6 +50,23 @@ export class ItemModel {
...
+    // only used for order print button
+    this.isImage =
+      this.hasItems &&
+      (this.contentType === "singe image" ||
+        this.contentType === "multiple images");
+    // this.isImage = !!canvasAnnotations[0] && canvasAnnotations[0].body.type === "Image";
+
+    this.imageIDs = this.hasItems
+      ? canvases.map((canvas) => {
+          return canvas.id.split("/")[5];
```
sarangj, line 67: Might be useful to add a quick comment with an example url to explain why we're getting the specific index

app/src/models/item.ts
```diff
@@ -99,14 +129,15 @@ export class ItemModel {
       locations: manifestMetadataHash["Library Locations"]
         ? manifestMetadataHash["Library Locations"].join("<br>")
         : "",
-      subjects: manifestMetadataHash["Subjects"]
-        ? manifestMetadataHash["Subjects"].join("<br>")
-        : "",
+      subjects:
+        manifestMetadataHash["Subjects"].length > 2
+          ? manifestMetadataHash["Subjects"].join("<br>")
+          : "",
```
sarangj, line 135: Might be misunderstanding, but if there's only 1 subject, we'll set this to be empty

sarangj (approved):
approving to unblock, but a couple questions

app/src/components/items/uv/universalViewer.tsx
avertrees: nope this was to demo behavior and I accidentally left it in

app/src/components/items/viewer/fallback.tsx
avertrees: yup

PR #359, opened 2025-06-12
fix: Add keys to list elements

sarangj:
## Ticket:

- JIRA ticket [DR-xxxx](url)

Add keys to elements in the metadata and citations that are structured in a list.  Note I had to replace a couple instances of `<>` fragment syntax in favor of the explicit `<Fragment` to pass in a key prop.

This fixes *most* warnings for missing / duplicate keys - there are some edge cases I've seen that are a bit tricky for me, and would have to take more investigation

Opened an item page locally and verified the warnings go away
### Checklist:

### DIFF
--- a/app/src/components/items/overview/citations/overview.tsx
+++ b/app/src/components/items/overview/citations/overview.tsx
@@ -1,4 +1,5 @@
 import { Box, Flex, Heading, Text } from "@nypl/design-system-react-components";
+import { Fragment } from "react";
 import parse from "html-react-parser";
 
 // Helper function to parse anchor tag from HTML string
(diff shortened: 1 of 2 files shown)


app/src/components/items/overview/citations/overview.tsx
```diff
@@ -90,14 +91,12 @@ ${metadata.origin ? "(" + metadata.origin + ")" : ""}${
           const value = citationFormatDummyData[field];
           if (value !== "") {
             return (
-              <>
-                <Text key={index} size="overline1" marginBottom="xs">
+              <Fragment key={`citation-${field}`}>
```
avertrees, line 94: curious about why this needs a Fragment?

avertrees: I'll read the docs

avertrees (approved):

app/src/components/items/overview/citations/overview.tsx
sarangj: `<>` is just shorthand for a `Fragment` - if you want to add any props to a `<>` though, you have to explicitly use `Fragment`

PR #378, opened 2025-06-23
DR-3659: Playwright tests to click on items in search results

clarissarichard:
## Ticket:

- JIRA ticket [DR-3659](https://newyorkpubliclibrary.atlassian.net/browse/DR-3659)

- Adds Playwright tests to confirm the first result in both filtered and unfiltered search results navigates to an `/items/` page
- ~~Adds Playwright test to confirm the first result's Item page displays the selected filters~~

## Open questions/Notes
- ~~To confirm that the search results were filtered correctly, should the "confirms search results were filtered correctly" tests remain in `item.spec.ts` or be moved to `search.spec.ts`? I would like to eventually add tests filtering search results by dates and rights availability and confirming one of the search results matches and reflects those filters.~~
- Noting that there are additional tests that can be written for the Item page in `item.spec.ts`:
- [ ] displays image, video, audio, or on-site placeholder
- [ ] displays order print button
- [ ] displays item data
- [ ] displays cite this item

Run `npm run playwright` and observe the Playwright Tests check in the PR

N/A

### Checklist:

[DR-3659]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


playwright/pages/search.page.ts
```diff
@@ -70,6 +71,9 @@ export default class SearchPage {
       ),
     });
     this.firstResult = page
+      .locator("a[href*='/items/'], a[href*='/collections/']")
```
Toxiapo, line 74: Just out of my curiosity: When will the result page return a "collections" url item?

clarissarichard: If we wanted to change the search term, there's a chance a collection can be returned as the first result if it relates to the search term. In this particular case with the "map of scandinavia" search term, there actually aren't any collections in the search results. If it doesn't make sense to keep, let me know!

playwright/tests/search.spec.ts
```diff
@@ -174,3 +177,23 @@ test.describe("sorts search results", () => {
...
+  });
+
+  test("clicks on an item in filtered search results", async () => {
+    await expect(searchPage.refineHeading).toBeVisible();
+
+    await searchPage.filterSearchResults();
+
+    await expect(searchPage.firstResult).toBeVisible();
+    await searchPage.firstResult.click();
+    await expect(searchPage.page).toHaveURL(/\/(items)\//);
```
Toxiapo, line 197: Im wondering if we should also verify something on the page itself, rather than relying on the URL change. 🤔 In some cases, especially with React applications, a URL change doesn't necessarily mean the page has redirected successfully.

clarissarichard: That's true! The items page is still in development, but I could find something that's unlikely to change like the search bar or header navigation.

Toxiapo: If it's not possible, even something like `searchPage.refineHeading` is not visible should be enough!

playwright/pages/search.page.ts
Toxiapo: Hmmm, I think I'll just use "items" for now. If a search term change triggers the test to fail, that might actually be a good failure, at least it feels more deterministic imo! What do ya think?

Toxiapo (approved):

PR #389, opened 2025-06-26
use new permitted location field in manifest

avertrees:
- use Permitted Locations field from manifest to populate banner for restricted Items
- update GA4 data

### DIFF
--- a/app/src/models/item.ts
+++ b/app/src/models/item.ts
@@ -41,6 +41,7 @@ export class ItemModel {
   breadcrumbData: any;
   mediaFiles: string[];
   subcollectionName: string | null;
+  permittedLocationText: string | null;
 
   constructor(uuid: string, manifest: any) {
     const parser = new Maniiifest(manifest);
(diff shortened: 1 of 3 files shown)


app/src/models/item.ts
```diff
@@ -86,6 +87,11 @@ export class ItemModel {
         ? rawManifestMetadata["Division"].toString()
         : rawManifestMetadata["Library Locations"][0] || "";
 
+    this.permittedLocationText =
+      this.isRestricted && rawManifestMetadata["Permitted Locations"]
+        ? rawManifestMetadata["Permitted Locations"][0].toString()
```
sarangj, line 92: does the `toString` call need to be `?.toString()` in case the array is empty or null?

sarangj (commented):
just one question!

app/src/models/item.ts
avertrees: I'll add

PR #392, opened 2025-06-26
remove download option

avertrees:
disable download options for all AV

this uses the HTMLMediaElement `controlsList` feature which is partially supported across browsers: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList
thus, might want a better solution for this

additionally, Clare flagged this viewer behaving inconsistently across browsers and recommended a different viewer. contemplating spending time implementing that before diving deeper into this. 

open to thoughts and suggestions :)

### DIFF
--- a/app/src/components/items/react/player.tsx
+++ b/app/src/components/items/react/player.tsx
@@ -1,8 +1,12 @@
+import { useEffect } from "react";
+
 import dynamic from "next/dynamic";
 
 const Player = dynamic(() => import("react-player/lazy"), { ssr: false });
 
 export default function ReactPlayer({ src, type }) {
+  useEffect(() => {});
+
   return (
     <div>
       <Player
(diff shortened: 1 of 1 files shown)


sarangj (approved):

app/src/components/items/react/player.tsx
```diff
@@ -1,8 +1,17 @@
...
 import dynamic from "next/dynamic";
 
 const Player = dynamic(() => import("react-player/lazy"), { ssr: false });
 
 export default function ReactPlayer({ src, type }) {
+  useEffect(() => {
+    // disable download
+    const player = document.querySelector("video") as any;
+    player?.controls; // true
+    player?.controlsList == "nodownload";
```
7emansell, line 12: this needs to be `player?.setAttribute("controlslist", "nodownload");`

7emansell (changes requested):
It looks like Firefox and Safari already don't offer the download as a menu option so even though this `controlslist` fix is not compatible across all browsers, it works for Chrome and that's all we really need it for.

That said, is the expectation here that the video cannot be downloaded at all ? bc regardless of browser you can always right click and save it. If we just want to discourage it, then i think disabling the control works as well as anything

PR #394, opened 2025-06-27
[NO-REF] Change less to fewer because I am that guy.

emu47:
Just fixing a grammatical issue that bugs me.


roseb89:
There's no ticket for this, so no update to the changelog? Is that's not needed for a simple text change like this?

avertrees (approved):

avertrees:
> There's no ticket for this, so no update to the changelog? Is that's not needed for a simple text change like this?

@emu47 do you mind adding a quick changelog note for this?

PR #398, opened 2025-07-01
[WIP] Redirect capture2

sarangj:
## Ticket:

- JIRA ticket [DR-xxxx](url)

### Checklist:


sarangj:
I'm struggling a bit with tests here - but figured I'd have this up for review while I work on that

playwright/tests/item-page.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { test, expect } from "@playwright/test";
+import ItemPage from "../pages/item-page";
+
+let itemPage: ItemPage;
+
+// Probably a nicer way to do this than hardcoding but :shrug?
+const CAPTURE_UUID = "e36b2b90-740d-0136-0517-476ebf952da4";
+const ITEM_UUID = "7cd3acc0-5d7f-0134-12ab-00505686a51c";
```
samanthaandrews, line 8: i don't mind the hardcoding at this phase tbh. if we decide to do some more serious mocking in these tests then we can definitely go that route, but for now seems fine.

my suggestion is to move these consts into the page file and make them part of the ItemPage class. something like:

```
static captureUuid: string = "e36b2b90-740d-0136-0517-476ebf952da4";
etc.
```

playwright/pages/item-page.ts
samanthaandrews, line 1: let's rename this file to `item.page.ts`

app/src/components/items/uv/universalViewer.tsx
```diff
@@ -12,13 +12,28 @@ export type UniversalViewerProps = {
...
+    // we strip off the first 3 chars
+    const hash = window.location.hash.slice(3);
+    const captureUuid = new URLSearchParams(hash).get("uuid");
+    if (captureUuid) {
+      const captureIdx = captureUuidToIdx[captureUuid];
+      if (captureIdx) {
+        window.location.replace(
+          window.location.pathname + `?canvasIndex=${captureIdx}`
+        );
+      }
```
samanthaandrews, line 34: this is a little bit of a drive by comment, but generally speaking if possible we would try to use a more declarative approach here. Nextjs has a `useRouter` method that could work. something like this:

```
// this import goes at the top of the file
import { useRouter } from 'next/router'

// at the top of the UniversalViewer component you'd put this
const router = useRouter();

if (captureIdx) {
  const urlSearchParams = new URLSearchParams(window.location.search);
  urlSearchParams.set("canvasIndex", captureIdx.toString());
  const newUrl = `${window.location.pathname}?${urlSearchParams.toString()}`;
  router.push(newUrl);
}
```
the nice part about Nextjs router is that it handles edge cases like URL encoding and browser-specific quirks, making it (usually) more robust than manual URL manipulation

samanthaandrews: ah nice, i didn't know this part too:

> pushState and replaceState calls integrate into the Next.js Router, allowing you to sync with [usePathname](https://nextjs.org/docs/app/api-reference/functions/use-pathname) and [useSearchParams](https://nextjs.org/docs/app/api-reference/functions/use-search-params).

you're right @avertrees; better to use `window.location.replaceState` in this case.

avertrees (approved):
approving for now. have one small comment but it's not a blocker

PR #403, opened 2025-07-03
DR-3748 render order print button only if item is marked buyable

chrisb87:
https://newyorkpubliclibrary.atlassian.net/browse/DR-3748

### DIFF
--- a/app/src/components/items/overview/overview.tsx
+++ b/app/src/components/items/overview/overview.tsx
@@ -30,7 +30,9 @@ const ItemOverview = ({ item }) => {
           catalogLink={item.catalogLink}
           archivesLink={item.archivesLink}
         />
-        {item.isImage && <PrintOverview imageIDs={item.imageIDs} />}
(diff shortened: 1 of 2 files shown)


app/src/components/items/overview/overview.tsx
```diff
@@ -30,7 +30,9 @@ const ItemOverview = ({ item }) => {
           catalogLink={item.catalogLink}
           archivesLink={item.archivesLink}
         />
-        {item.isImage && <PrintOverview imageIDs={item.imageIDs} />}
+        {item.isImage && item.buyable && (
+          <PrintOverview imageIDs={item.imageIDs} />
```
avertrees, line 34: this will hide both the order Print button and the Image ID (not ideal but that's how the component is structured atm). I don't want to ask you to break up the component, so easiest solution is to add a "buyable" param to the `PrintOverview` component, pass in `item.buyable` as the value, and then conditionally render the button in the `PrintOverview` component itself.

PR #418, opened 2025-07-16
switch image source to images.nypl.org

avertrees:


emu47:
We could probably close this PR, as I can (happily) say that I don't think we need to do this.

PR #432, opened 2025-08-07
Dr 3796/test search filters for Place and Collection

jbdalton:
## Ticket:

- JIRA ticket [DR-DR3796](https://newyorkpubliclibrary.atlassian.net/browse/DR-3796)

Adds Place and Collection Filters to the search.  The first, default row of filters is now complete, and I'll start on the 2nd row next.  

## Open questions

I tried (with AI assistance) to refactor these to Accessibility API matchers for the _selected_ filters.  But there are some factors which I think make it more straightforward to go with a getByText matcher (but scoped within its appropriate selector) for these filter tests.   For example, the link that Playwright sees contains a span around the place|name|collection etc., and then a count (which can change) after that, but all in the same link.  Also, there are piped UUIDs existing at the end of "collection-filter" strings that, while not always visible, do display in shorter collection-titles.   By using getByText here, along with substring matching, we're able to directly test the _functionality_ (does the filter work on selected choices?) while not getting in the weeds on the type of matchers of the string-text itself.   

Ran suite of filter-tests in playwright.

### Checklist:


playwright/pages/search.page.ts
```diff
@@ -107,7 +107,22 @@ export default class SearchPage {
     this.collectionFilter = this.page.getByRole("button", {
       name: "Collection",
     });
+    this.collectionOption = this.page
+      .locator("#select-collection")
+      .getByText("Portolan atlas", { exact: true });
+    this.collectionSelected = this.page
+      .locator("#select-collection")
+      // regex sub-string (or exact:false) is necessary for this match right now because there is a
+      // bug in the selected collection title filter, where the piped uuid that displays at end of title
```
sarangj: oooh interesting catch 👀 is there a ticket for this?

playwright/tests/search.spec.ts
```diff
@@ -124,6 +124,30 @@ test.describe("displays specific filter options", () => {
...
+
+    await expect(searchPage.placeFilter).toBeVisible();
+    await searchPage.placeFilter.click();
+    await expect(searchPage.placeOption).toBeVisible();
+    await searchPage.placeOption.click();
+    await expect(searchPage.applyFilterButton).toBeVisible();
+    await searchPage.applyFilterButton.click();
+    await expect(searchPage.placeSelected).toBeVisible();
+  });
+
```
sarangj: Very cool, I was able to click around the site for these and follow the steps!

sarangj (approved):

playwright/pages/search.page.ts
jbdalton: Yes, I put this comment in there as a TO-DO to remind me to make the ticket, and to remind me to remove regex method, if possible, once the issues's been addressed.

Toxiapo (approved):
> For example, the link that Playwright sees contains a span around the place|name|collection etc., and then a count (which can change) after that

Interesting, I thought the name selectors are fuzzy searches so `getByLabel('radio', { name: "Portolan atlas"})` should match either `Portolan atlas` or `Portolan atlas 1`

It's unfortunate that we can't use the label selector , but everything else looks good to me!

playwright/pages/search.page.ts
samanthaandrews: FWIW the code that is appending the uuid seems to be on line 56 in the `selectFilter.tsx` component:

```
    <Radio
      key={`${option.name}-${index}`}
      id={`${option.name}-${index}`}
      labelText={label}
      value={option.name}
    />
```
I'm not sure if this is intentional, but from an accessibility standpoint it doesn't seem right to me at first glance. Maybe @avertrees has more context though.

playwright/pages/search.page.ts
```diff
@@ -107,7 +107,22 @@ export default class SearchPage {
     this.collectionFilter = this.page.getByRole("button", {
       name: "Collection",
     });
+    this.collectionOption = this.page
+      .locator("#select-collection")
```
samanthaandrews: I just checked this myself and it looks like `this.collectionOption = this.page.getByLabel("collection filter options").getByText("Portolan atlas");` works, which means it's possible to use labels for some of these selectors instead of grabbing the id, which would be preferrable

jbdalton: Were you able to run the test with that replacement?

Where I was having trouble was on the "after-selected" part of the collectionOption choice.  You can see it on the page itself too. Where, after selecting "Portolan atlas" you'll see "Portolan atlas||collection-uuid" in the selected-collection slot in that filter   (I don't have the uuid in front of me rn, so I'm using "collection-uuid" as a placeholder here).   So when we are checking if the "selected text" part of the test works, simply checking the label wasn't finding the "text" Portolan atlas, because the text of the box in that case is actually displaying "Portolan atlas||collection-uuid"

samanthaandrews: Yeah I ran the test with this replacement and it passes. Made a PR off of yours to see it working: https://github.com/NYPL/digital-collections/pull/441

Also updated the workflow to make sure Playwright is still running on all PRs. It looks like changing the base branch to `qa` broke this.

jbdalton:
I've added Sam's suggestions, and added a revised comment on the UUID bug with Jira ticket reference.

I've updated changelog, re-pushed to origin, had to do one more fix to a broken About Page test, and I THINK now this PR is finally ready to be merged!?

PR #437, opened 2025-08-11
Track AV progress and send to google analytics

sarangj:
## Ticket:

- JIRA ticket [DR-3593](https://newyorkpubliclibrary.atlassian.net/browse/DR-3593)
- JIRA ticket [DR-3594](https://newyorkpubliclibrary.atlassian.net/browse/DR-3593)

Use the plyr on 'timeupdate' event to track when specific milestones are reached when playing an AV file. The idea here is to track when a 'milestone' percentage gets crossed, and send an event for those milestones. I basically followed the same logic as in OG DC: https://github.com/NYPL/digitalreadingroom/blob/9c01d999ef8085b867869f65d87cbbc2bbaedeca/app/assets/javascripts/ga-plugin.js#L199-L217 where we wait until a given milestone is passed, but if we're more than 5% past the milestone, we don't trigger an event, to account for the case where a user is clicking around (ie, if the user skips straight to 40% into the video, we don't want to trigger an event for 25% progress reached).

Note, I had to split the plyr out into a child component to cleanly use the `usePlyr`, as in this example from the plyr-react library: https://github.com/chintan9/plyr-react/blob/eb579712a2e3a7b871a24daa7241acf01856699d/example/src/custom-audio-player.tsx#L18-L50

Also note that for the 100& milestone, I simply use the playback `end` event.

I basically tested this by playing some AV material locally and inspecting console logs, this will need to be further tested in QA.

### Checklist:

[DR-3593]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DR-3594]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/items/plyr/player.tsx
```diff
@@ -74,20 +75,10 @@ const Player = ({ title, sources, type }: PlyrProps) => {
   return (
     <div>
       {sources.length === 1 ? (
-        <Plyr
-          source={source}
-          options={undefined}
-          height={playerHeight}
-          width="100%"
-        />
+        <CustomPlyr ref={plyrRef} source={source} options={undefined} />
```
sarangj: I don't need to pass `options` here

app/src/components/items/plyr/player.tsx
```diff
@@ -115,4 +106,43 @@ const Player = ({ title, sources, type }: PlyrProps) => {
...
+            console.log(`${milestone}%`);
+            loggedProgressEvents.current.add(milestone);
+          }
+          break;
+        }
+      }
+    };
+    const api = current as { plyr: PlyrInstance };
+    api.plyr.on("timeupdate", handleTimeUpdate);
+    api.plyr.on("ended", () => trackAVProgress(source.type, source.title, 100));
```
sarangj: I should make sure this only fires once in case the user navigates back and ends the playback multiple times.

app/src/components/items/plyr/player.tsx
```diff
@@ -115,4 +106,43 @@ const Player = ({ title, sources, type }: PlyrProps) => {
...
+  });
+  return (
+    <video
```
sarangj: This seems to work for audio, but I suppose I should do this correctly

sarangj: hmmm actually this seems to be the way - the ply-react provided audio example still uses a video element: https://github.com/chintan9/plyr-react/blob/eb579712a2e3a7b871a24daa7241acf01856699d/example/src/custom-audio-player.tsx#L43-L46

sarangj:
i've been testing this audio for anyone who wants to look: https://digital-collections-git-videoanalytics-nypl.vercel.app/items/aa56ae00-b884-0133-3277-60f81dd2b63c

you'll see the milestone percentages logged in the console

app/src/components/items/plyr/player.tsx
```diff
@@ -115,4 +106,48 @@ const Player = ({ title, sources, type }: PlyrProps) => {
   );
 };
 
+const CustomPlyr = forwardRef<APITypes, any>((props, ref) => {
```
avertrees: nit but can we make this it's own separate component?

sarangj: done!

avertrees (approved):

PR #438, opened 2025-08-12
[DR-3819] Removing whole image download options.

emu47:
## Ticket:

- JIRA ticket [DR-3819](https://newyorkpubliclibrary.atlassian.net/browse/DR-3819)

Removes the whole image download options from the download dialogue box.

## Open questions

Theoretically, this should be simpler. It should be a config change. The follow would work to remove the whole image download options: 

`{
  "modules": {
    "downloadDialogue": {
      "options": {
        "downloadWholeImageHighResEnabled": false,
        "downloadWholeImageLowResEnabled": false
      }
    }
  }
}`

The problem is that this does not let our custom sizes through. I've yet to find a way to adjust the manifest and the UV configs to work quite how I want to. 

Go to the download options dialogue for any image in DC and check the links provided. Make sure they work, and make sure only 'current view' (if zoomed in) and our custom derivative sizes, and a tiff for any public domain images, show up. 

None that I can think of. Link style remains as it's been. This removes two links. 

### Checklist:

[DR-3819]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chrisb87 (approved):
This is amusing but gets the job done!

PR #441, opened 2025-08-12
DR-3796: Search filters updates suggestions

samanthaandrews:
Follow up to https://github.com/NYPL/digital-collections/pull/432

### DIFF
--- a/playwright/tests/search.spec.ts
+++ b/playwright/tests/search.spec.ts
@@ -130,7 +130,8 @@ test.describe("displays specific filter options", () => {
     await expect(searchPage.collectionFilter).toBeVisible();
     await searchPage.collectionFilter.click();
     await expect(searchPage.collectionOption).toBeVisible();
-    await searchPage.collectionOption.click();
+    await searchPage.collectionOption.check();
+    await expect(searchPage.collectionOption).toBeChecked();
     await expect(searchPage.applyFilterButton).toBeVisible();
     await searchPage.applyFilterButton.click();
     await expect(searchPage.collectionSelected).toBeVisible();
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,8 +1,5 @@
 name: Playwright Tests
-on:
-  pull_request:
-    branches:
-      - main
+on: pull_request
 jobs:
   test:
     timeout-minutes: 60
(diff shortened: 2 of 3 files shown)


jbdalton:
Although this matcher scenario allows the test to pass, it does require `exact: true` to not be set.  Otherwise, the Collection filter, unlike the other filters (Publisher, etc.) which are all using the `exact: true` setting, will fail the `getByText` assertion when it encounters the visible string "Portolan atlas||`<collection-uuid>`."  The fuzzy default of `getByText` allows it to still pass in this version. 

Not sure if I should A.) leave it this way while I add tests for the rest of the filters, B.) add back the `exact: true` to `getByText` (in which case it will fail) ,  C.) do the latter but tag the test as wip/failing so it doesn't run until the display bug is fixed, or D.) do the first choice and add a comment that we'll circle back to strict `exact: true` matching once bug is fixed.

jbdalton:
@samanthaandrews - For now, I think the most expedient way to unblock this is to go ahead and merge your PR into my active test branch, and add a comment for the `getByText` line noting the bug, so we don't forget to change it back to "strict" matching once that's addressed.   Do you think this approach make sense?

samanthaandrews:
> @samanthaandrews - For now, I think the most expedient way to unblock this is to go ahead and merge your PR into my active test branch, and add a comment for the `getByText` line noting the bug, so we don't forget to change it back to "strict" matching once that's addressed. Do you think this approach make sense?

Yup I think that's a good move. In the comment can you also add the link to the ticket that references this bug fix? Thanks!

playwright/tests/search.spec.ts
```diff
@@ -130,7 +130,8 @@ test.describe("displays specific filter options", () => {
     await expect(searchPage.collectionFilter).toBeVisible();
     await searchPage.collectionFilter.click();
     await expect(searchPage.collectionOption).toBeVisible();
-    await searchPage.collectionOption.click();
+    await searchPage.collectionOption.check();
```
samanthaandrews, line 133: @jbdalton want to call out the shift from using `click()` to `check()` here which is preferable for radio buttons like we are using in this case. Link to Playwright docs about check() https://playwright.dev/docs/input#checkboxes-and-radio-buttons

jbdalton: Noticed that, and was wondering which to go with.   I'll refactor those existing calls in a separate PR that has the remaining filter-tests in it.

PR #443, opened 2025-08-15
Dr 3796 finish final test search filters

jbdalton:
## Ticket:

Adds the remaining filters -- genre, division and type -- to the search-filter tests.  Also some refactoring to get radio check assertions correct, and to try to get things to run a little faster.

## Open questions

I still would like to clean up some of the testing language to make it easier to follow in reports, but these tests do work now, and the whole suite, when run in chromium, currently takes 1.6 min with these changes.  I plan to focus the next PR on timeouts and flakey or brittle tests.

Ran `npx playwright test --project=chromium `

### Checklist:

-  X  All new and existing tests passed.

### DIFF
--- a/playwright/pages/search.page.ts
+++ b/playwright/pages/search.page.ts
@@ -125,6 +125,12 @@ export default class SearchPage {
       .locator("#select-place")
       .getByText("Place: England", { exact: true });
     this.genreFilter = this.page.getByRole("button", { name: "Genre" });
+    this.genreOption = this.page
+      .locator("#select-genre")
+      .getByText("Maps", { exact: true });
+    this.genreSelected = this.page
+      .locator("#select-genre")
+      .getByText("Genre: Maps", { exact: true });
     this.publisherFilter = this.page.getByRole("button", { name: "Publisher" });
     this.publisherOption = this.page
       .locator("#select-publisher")
@@ -133,7 +139,23 @@ export default class SearchPage {
       .locator("#select-publisher")
       .getByText("Publisher: Printed at the Theater,", { exact: true });
     this.divisionFilter = this.page.getByRole("button", { name: "Division" });
+    this.divisionOption = this.page
+      .locator("#select-division")
+      .getByText("Lionel Pincus and Princess Firyal Map Division", {
+        exact: true,
+      });
+    this.divisionSelected = this.page
+      .locator("#select-division")
+      .getByText("Division: Lionel Pincus and Princess Firyal Map Division", {
+        exact: true,
+      });
     this.typeFilter = this.page.getByRole("button", { name: "Type" });
+    this.typeOption = this.page
+      .locator("#select-type")
+      .getByText("Cartographic", { exact: true });
+    this.typeSelected = this.page
+      .locator("#select-type")
+      .getByText("Type: Cartographic", { exact: true });
     this.startYear = this.page.getByRole("textbox", { name: "Start year" });
     this.endYear = this.page.getByRole("textbox", { name: "End year" });
     this.applyDates = this.page.getByRole("button", { name: "Apply dates" });
--- a/playwright/pages/search.page.ts
+++ b/playwright/pages/search.page.ts
@@ -34,6 +34,7 @@ export default class SearchPage {
   readonly publisherSelected: Locator;
   readonly divisionFilter: Locator;
   readonly divisionOption: Locator;
+  readonly divisionSelected: Locator;
   readonly typeFilter: Locator;
   readonly typeOption: Locator;
   readonly typeSelected: Locator;
--- a/playwright/tests/search.spec.ts
+++ b/playwright/tests/search.spec.ts
@@ -6,11 +6,15 @@ let searchPage: SearchPage;
 test.beforeEach(async ({ page }, testInfo) => {
   const excludedTests = [
     "find item with a keyword search",
+    "enter keyword and submit",
+    "display search results with a clickable item",
     "after opening 2nd row filters",
     "choose publisher",
     "choose genre",
     "choose division",
     "choose type",
+    "filter search results by date",
+    "filter search results by availability",
   ];
 
   if (!excludedTests.includes(testInfo.title)) {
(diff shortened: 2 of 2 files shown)


playwright/pages/search.page.ts
```diff
@@ -133,7 +140,23 @@ export default class SearchPage {
       .locator("#select-publisher")
       .getByText("Publisher: Printed at the Theater,", { exact: true });
     this.divisionFilter = this.page.getByRole("button", { name: "Division" });
+    this.divisionOption = this.page
+      .locator("#select-division")
```
samanthaandrews, line 144: I'd like to move away from using ids as locators here since they can be brittle. I pulled down your branch and was able to use `getByLabel("division filter options")` here and I believe it's possible to do something similar across most of the locators in this PR.

playwright/tests/search.spec.ts
```diff
@@ -143,24 +183,91 @@ test.describe("displays specific filter options", () => {
     await expect(searchPage.placeFilter).toBeVisible();
     await searchPage.placeFilter.click();
     await expect(searchPage.placeOption).toBeVisible();
-    await searchPage.placeOption.click();
+    await searchPage.placeOption.check();
```
samanthaandrews, line 186: nice!

playwright/tests/search.spec.ts
```diff
@@ -143,24 +183,91 @@ test.describe("displays specific filter options", () => {
...
-    await searchPage.showFilters.click();
-    await expect(searchPage.publisherFilter).toBeVisible();
-    await searchPage.publisherFilter.click();
-    await expect(searchPage.publisherOption).toBeVisible();
-    await searchPage.publisherOption.click();
-    await expect(searchPage.applyFilterButton).toBeVisible();
-    await searchPage.applyFilterButton.click();
-    await expect(searchPage.publisherSelected).toBeVisible();
+  // Go to search page once, open 2nd row, and check filters, date range and availability
+  test.describe.serial("after opening 2nd row filters", () => {
```
samanthaandrews, line 193: it's not clear to me from the description of the test what this is doing exactly. the comment is somewhat more useful, so let's rework to remove the comment and just use that as the description of the test.

playwright/tests/search.spec.ts
```diff
@@ -265,24 +372,13 @@ test.describe("sorts search results", () => {
...
-
-    await expect(searchPage.firstItemResult).toBeVisible();
-    await searchPage.firstItemResult.click();
-    await expect(searchPage.page).toHaveURL(/\/(items)\//);
-    await expect(searchPage.refineHeading).not.toBeVisible();
-  });
+  await expect(searchPage.firstItemResult).toBeVisible();
+  await searchPage.firstItemResult.click();
+  await expect(searchPage.page).toHaveURL(/\/(items)\//);
+  await expect(searchPage.refineHeading).not.toBeVisible();
```
samanthaandrews, line 383: let's add one more positive assertion here to confirm that the page navigated as expected. is there a heading or something on the new page that we can verify is visible?

playwright/tests/search.spec.ts
```diff
@@ -143,24 +183,91 @@ test.describe("displays specific filter options", () => {
...
+
+    test.beforeAll(async ({ browser }) => {
+      // Perform one-time setup here
+      browserContext = await browser.newContext();
+      page = await browserContext.newPage();
+
+      searchPage = await SearchPage.loadPage(SearchPage.searchResultsUrl, page);
+      await expect(searchPage.showFilters).toBeVisible();
+      await searchPage.showFilters.click();
+    });
```
samanthaandrews, line 206: i don't understand why we need this setup in the beforeAll hook when there is another beforeAll hook at the very top of the file. we either need to clean this up or make it clear why we need it.

in either case, we shouldn't need to launch a new browser context at all here. loading the search page should be enough.

playwright/tests/search.spec.ts
```diff
@@ -143,24 +183,91 @@ test.describe("displays specific filter options", () => {
...
+
+    test.afterAll(async () => {
+      // Close the page and context
+      await page.close();
+      await browserContext.close();
+    });
```
samanthaandrews, line 212: i don't think this is necessary since we don't need to open a new browser context. i had to do some digging into the Playwright docs as well, but i don't think the page close is necessary either. none of the Playwright docs recommend this or show it in their examples.

playwright/tests/search.spec.ts
```diff
@@ -1,39 +1,79 @@
...
+    "enter keyword and submit",
+    "display search results with a clickable item",
+    "after opening 2nd row filters",
+    "choose publisher",
+    "choose genre",
+    "choose division",
+    "choose type",
+    "filter search results by date",
+    "filter search results by availability",
+  ];
```
samanthaandrews, line 18: this variable name is confusing for folks who might not be familiar for these tests. why are we excluding tests?

playwright/tests/search.spec.ts
```diff
@@ -1,39 +1,79 @@
...
+    await searchPage.showFilters.click();
   });
 
-  await expect(page).toHaveTitle("Search results - NYPL Digital Collections");
-});
+  test.afterAll(async () => {
+    // Close the page and context
+    await page.close();
+    await browserContext.close();
+  });
```
samanthaandrews, line 44: I believe Playwright does this cleanup for us automatically, so we can remove it.

samanthaandrews (changes requested):
it's looking good; a few changes from me. in the future i'd recommend even smaller PRs which will make them easier to review.

playwright/tests/search.spec.ts
```diff
@@ -1,39 +1,79 @@
...
+    "filter search results by date",
+    "filter search results by availability",
+  ];
+
+  if (!excludedTests.includes(testInfo.title)) {
     searchPage = await SearchPage.loadPage(SearchPage.searchResultsUrl, page);
   }
 });
 
-test("searches for a keyword from homepage", async ({ page }) => {
```
clarissarichard, line 12: Will you be testing search from the homepage elsewhere, like in `homepage.spec.ts`?

ETA: Unless it's just been renamed to `test("enter keyword and submit")` below!

playwright/tests/search.spec.ts
```diff
@@ -1,39 +1,79 @@
...
-  searchPage = await SearchPage.loadPage("/", page);
+// Do a new search from a results-page
+test.describe("find item with a keyword search", () => {
+  let searchPage: SearchPage;
```
clarissarichard, line 27: This is already set in line 4 above!

PR #453, opened 2025-09-04
Pass items endpoint data into item model [DR-3855]

sarangj:
## Ticket:

- JIRA ticket [DR-3855](https://newyorkpubliclibrary.atlassian.net/browse/DR-3855)

We want to fully remove these manifest fetches, but in order to do so, I'll need to add more data to the existing items endpoint. In the meantime, get the ball rolling by using the items endpoint where I can and adding an interface to make the data types clearer.

This also uses the new endpoint for all the fields we plan on removing from the manifest, so this will unblock that work

Updated existing jest tests + clicked around

### Checklist:

[DR-3855]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/types/CollectionsAPI.ts
+++ b/app/src/types/CollectionsAPI.ts
@@ -1,0 +1,17 @@
+export interface APICapture {
+  uuid: string;
+  imageId: string | null;
+  orderInSequence: number;
+}
+
+export interface APIItem {
+  uuid: string;
+  buyable: boolean;
+  isRestricted: boolean;
+  permittedLocationText: string | null;
+  contentType: string | null;
+  yearStart: string | null;
+  yearEnd: string | null;
+  citationResourceType: string | null;
+  captures: APICapture[];
+}
--- a/app/src/utils/apiClients/apiClients.tsx
+++ b/app/src/utils/apiClients/apiClients.tsx
@@ -16,6 +16,7 @@ import {
 } from "../../config/constants";
 import { fetchApi } from "../fetchApi/fetchApi";
 import { Filter } from "../../types/FilterType";
+import { APIItem } from "@/src/types/CollectionsAPI";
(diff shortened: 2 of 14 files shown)


app/src/types/CollectionsAPI.ts
```diff
@@ -0,0 +1,17 @@
+export interface APICapture {
+  uuid: string;
+  imageId: string | null;
+  orderInSequence: number;
+}
+
+export interface APIItem {
```
emu47, line 7: nice

app/src/utils/apiClients/apiClients.tsx
```diff
@@ -153,7 +154,7 @@ export class CollectionsApi {
     });
   }
 
-  static async getItemData(uuid: string) {
+  static async getItemData(uuid: string): Promise<APIItem> {
```
emu47, line 157: like it ... just wondering if we're SURE we have ALL the fields we need in the ApiItem interface

emu47:
I was able to click around a lot without any apparent issues. Not the most confident at understanding all the implications, but it looks good to me.

emu47 (approved):

app/src/utils/apiClients/apiClients.tsx
sarangj: we don't yet! The page is still getting hydrated with both this and the manifest, this PR was just preliminary to split up some of the work. Once https://github.com/NYPL/collections-api/pull/243, I'll add a followup to this PR to completely cut out the manifest calls.

PR #464, opened 2025-09-17
NO-REF: Fix Playwright test timeouts

samanthaandrews:
## Ticket:

- No Jira ticket

- Making Playwright tests run after a production nextjs build & server start in CI
- Removing firefox and webkit browsers so that tests only run on chrome (should greatly reduce time it takes tests to run)
- Remove most of the footer assertions on the homepage test. We are still doing a sanity check to make sure the footer appears, but we should be testing the footer more thoroughly in the source code.
- Random updates to fix flaky tests (honestly idk if any of this worked)
- Code cleanup & file renaming

## Open questions

- Looks like we do have parallelization enabled, which is good. We can look into sharding, but ultimately I think the changes made here make noticeable improvements.

Tests run in CI on PRs, so we should be able to see if this fixes issues we are seeing. Locally, I've been running test suites to ensure they run and pass, which has been going fairly well. There are still some tests that are flaky for me when running locally. In CI we have retries enabled, which should address this flake.

### Checklist:


sarangj (approved):

playwright.config.ts
```diff
@@ -73,12 +73,9 @@ export default defineConfig({
...
-        // only run in local environment not CI
-        command: "npm run dev",
-        url: "http://localhost:3000",
-        reuseExistingServer: !process.env.CI,
-      },
+  webServer: {
+    command: "npm run start",
+    url: "http://localhost:3000",
+    reuseExistingServer: !process.env.CI,
+  },
```
samanthaandrews, line 80: @jbdalton one important change in this PR is that in order to run the tests you should make a production build of the app first, using the command `npm run build`. This will build the app for "prod" locally in your `dist` folder. After that, you can run `npm run playwright` which will start that prod build. This should make testing a lot easier. `npm run dev` is running a dev build which is much less performant.

PR #465, opened 2025-09-18
DR-3790: Fix timeouts on homepage localhost tests

jbdalton:
## Ticket:

- JIRA ticket [DR-3790](https://newyorkpubliclibrary.atlassian.net/browse/DR-3790)

I've added a few things, after trying several alternatives, to try to address some of the continuing timeout issues, especially for the DC homepage tests.

The two main changes I've made are: 

- added route-filters on 3rd-party calls (New Relic, adobe, google, etc.) that normally wouldn't be running on dev testing
- changed wait-until to "load"

This builds on the improvements made recently that allow us to run our playwright tests on localhost against production builds for better performance and stability during testing.

## Open questions

When this is run in CI, I assume that New Relic, google, adobe etc. are not being called against the prod site.  Or maybe this is a minimal overhead in that case.  I found removing it from localhost calls against the prod build keeps it from being part of our analytics and runs faster, at least when the waitUntil-load method is used on the page. 

I found that currently, while homepage.spec.ts would _sometimes_ run without failing, if it were run a few times -- or especially if run within the entire suite, we could replicate failures.  These were usually on the "verify feedback form" scenario in the homepage test.

I compared several different configurations until I found a combination that seems to be 0% flakey, at least during my testing.  Specifically I've found that the visual overlay that uses IIIF's "main-feature" splash image is loaded twice (a 900 px and 1600 px version) to produce the main visual element on the home page.  And whenever it fails (especially on localhost/dev, but even when run against a prod build) the feedback-form tests also fail.  Overall, this is about 10-20% of the time.  My theory is it's bottlenecking other assertions that Playwright might be waiting for on the "verify feedback form" functionality, which does happen to appear at the bottom of a DC homepage-load.  Even after changing to `{ waitUntil: "domcontentloaded" }` I found that the feedback-form test, no matter how large the timeout, would still  sometimes fail.

My goal was to find some combination that would run quickly, at least 7 times in a row, without a homepage.spec.ts fail, both as part of the entire test suite and when run individually.

The results of my testing look like this (if something failed over 30% of the time, I stopped  before 7 tries).  "Timeout" here refers to the timeout set in the specific feedback scenario via the "test" and "page" set methods.  Default.jpg is the visual overlay from IIIF.  There is a route-filter on it, as well as the 3rd-party calls.

Running suite:

|pass/fail|timeout|default.jpg|page-load-wait|avg-total|
|:--------|:--------|:--------|:--------|:--------|
|7/0 | 30000 | filter-out | "load" | 35-41s|
|1/3 | 30000 | filter-out | "domcontentloaded"| |
|1/3 | 60000 (test.set) | filter-out | "domcontentloaded" | |
|0/2 | 60000 (page.set) | filter-out | "domcontentloaded" | |

Single test only: homepage.spec.ts

|pass/fail|feedback-timeout|default.jpg|page-load-wait|avg-test-time|
|:--------|:--------|:--------|:--------|:--------|
|1/2 | 60000 (test.set) | filter | "domcontentloaded"| 9.1s|
|7/0| 30000 | filter | "load" | 10.6s|

### Checklist:

[DR-3790]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/tests/homepage.spec.ts
+++ b/playwright/tests/homepage.spec.ts
@@ -18,7 +18,7 @@ test.beforeEach(async ({ page }) => {
   await page.route("**/default.jpg", (route) => route.abort());
 
   // Navigate to the page after setting up the routing rules.
-  await page.goto("/", { waitUntil: "load" });
+  await page.goto("/");
 });
 
 test("verify navigation menu is displayed (items, collections, divisions, about)", async ({
--- a/playwright/tests/homepage.spec.ts
+++ b/playwright/tests/homepage.spec.ts
@@ -1,8 +1,10 @@
 import { test, expect } from "@playwright/test";
 import { DCHomepage } from "../pages/homepage.page";
 
+//test.setTimeout(60000);
+
 test.beforeEach(async ({ page }) => {
-  await page.goto("/", { waitUntil: "domcontentloaded" });
+  await page.goto("/", { waitUntil: "networkidle" });
 });
 
 test("verify navigation menu is displayed (items, collections, divisions, about)", async ({
@@ -73,7 +75,10 @@ test("verify footer links are visible", async ({ page }) => {
 
 test("verify feedback button is visible", async ({ page }) => {
   // this test is flaky; increasing timeout
-  test.setTimeout(120000);
+  // test.setTimeout(120000);
+  // a timeout set wihin the block won't affect this test, so using page timeout override instead
+  page.setDefaultTimeout(60000); // 60 seconds
+
   const dchomepage = new DCHomepage(page);
   await expect(dchomepage.feedbackButton).toBeVisible();
   await dchomepage.feedbackButton.click();
--- a/playwright/tests/homepage.spec.ts
+++ b/playwright/tests/homepage.spec.ts
@@ -85,13 +85,13 @@ test("verify footer links are visible", async ({ page }) => {
   page.setDefaultTimeout(30000); // 30 seconds
   const dchomepage = new DCHomepage(page);
   // the full footer content should be tested in the footer repo, not here in DC
+
   await expect(dchomepage.footerAccessibilityLink).toBeVisible();
 });
 
 test("verify feedback button is visible", async ({ page }) => {
-  // With route-filtering on and domcontentload as a default,
-  // extending timeouts might not be necessary anymore for feedback button.
-  //page.setDefaultTimeout(60000); // 60 seconds
+  // With route-filtering on, extending timeouts might not be necessary
+  // for feedback button tests.
   test.setTimeout(60000);
 
   const dchomepage = new DCHomepage(page);
--- a/playwright/tests/homepage.spec.ts
+++ b/playwright/tests/homepage.spec.ts
@@ -1,10 +1,24 @@
 import { test, expect } from "@playwright/test";
 import { DCHomepage } from "../pages/homepage.page";
 
-//test.setTimeout(60000);
-
 test.beforeEach(async ({ page }) => {
-  await page.goto("/", { waitUntil: "networkidle" });
+  // Block analytics, tracking, and third-party domains
+  await page.route(/.*googletagmanager\.com.*/, (route) => route.abort());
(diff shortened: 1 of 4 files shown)


playwright/tests/homepage.spec.ts
```diff
@@ -2,7 +2,23 @@ import { test, expect } from "@playwright/test";
...
+  await page.route(/.*omappapi\.com.*/, (route) => route.abort());
+  await page.route(/.*google-analytics\.com.*/, (route) => route.abort());
+
+  // If necessary, block the main-image overlay from iiif
+  // When running the whole suite, feedback on the homepage will often
+  // timeout when default img overlays are slow
+  await page.route("**/default.jpg", (route) => route.abort());
+
+  // Navigate to the page after setting up the routing rules.
+  await page.goto("/", { waitUntil: "load" });
```
samanthaandrews: The default behavior for Playwright's `.goto()` method is `waitUntil: "load"`, so this option can be removed.

playwright/tests/homepage.spec.ts
```diff
@@ -2,7 +2,23 @@ import { test, expect } from "@playwright/test";
...
-  await page.goto("/", { waitUntil: "domcontentloaded" });
+  // Block analytics, tracking, and third-party domains
+  await page.route(/.*googletagmanager\.com.*/, (route) => route.abort());
+  await page.route(/.*demdex\.net.*/, (route) => route.abort());
+  await page.route(/.*adobedtm\.com.*/, (route) => route.abort());
+  await page.route(/.*everesttech\.net.*/, (route) => route.abort());
+  await page.route(/.*ipify\.org.*/, (route) => route.abort());
+  await page.route(/.*google\.com.*/, (route) => route.abort());
+  await page.route(/.*omappapi\.com.*/, (route) => route.abort());
+  await page.route(/.*google-analytics\.com.*/, (route) => route.abort());
```
samanthaandrews, line 13: nice! if we find this helpful on the homepage, we could make a change to abort these calls for all tests on all pages using a global setup file. definitely doesn't need to happen on this PR.

samanthaandrews (changes requested):
One small change; otherwise this looks good. Once the change is made, you can re-request review from me so I'm notified to look at the PR again.

playwright/tests/homepage.spec.ts
jbdalton: removed the explicit "load"

Toxiapo:
Not sure if this will be helpful, but when I try to debug local versus virtual environment tests, I try to match the configurations as closely as possible. For instance, I noticed that the worker settings run with 2 workers in GH but 0 locally (which, I believe, will turns off parallelism). To make the environments more consistent, I would set a constant number of workers locally to match the GH setup.

samanthaandrews:
> Not sure if this will be helpful, but when I try to debug local versus virtual environment tests, I try to match the configurations as closely as possible. For instance, I noticed that the worker settings run with 2 workers in GH but 0 locally (which, I believe, will turns off parallelism). To make the environments more consistent, I would set a constant number of workers locally to match the GH setup.

This is good insight. I think our config currently is set to use 2 workers on GH Actions and "undefined" in local testing. In this context "undefined" actually means "use the max number of workers" which is pretty confusing. For me has been defaulting to 4. I think due to GH Actions concurrency limitations, 2 is what Playwright recommends because if you default to `undefined` in CI, then GH might default to only 1 which is less ideal.

samanthaandrews (approved):

jbdalton:
Yes, on my machine the undefined seems to default to 6 workers.

jbdalton:
I see at the top of the changelog an unreleased entry for "Fix Playwright test timeouts."  Should I be adding another entry for this specific merge, like  "DR-3790-Fix timeouts ...." or does that previous entry cover it?

PR #470, opened 2025-09-26
use Next Image component on About page

avertrees:
attempt to fix broken image on QA by explicitly importing the image via relative path and and then passing the image into the Next Image component (following the pattern that we use on the homepage for local assets)


sarangj (approved):
looks good in vercel, let's see if it works on qa! weird that we *have* to use next/image 🤔  Is it because this is one of those build-time rendered static pages?

PR #472, opened 2025-09-29
Explicitly avoid items page cache

sarangj:
Currently, we use `revalidatePath` to wipe the cache from the items page
- reading a bite more, it seems like a simpler approach would be to just not cache the data used to load the items page, and then avoid calling revalidate.  In practice, the items endpoint is already not cached anyway, since nextJS has some underlying magic to not cache any of the data if 'dynamic' functions like `headers()` are called (we call this method to fetch the client side IP address).

In practice, this should do nothing, except maybe reduce a tiny bit of overhead in calling `revalidatePath`. I think this might just make some of the cache behavior a bit clearer. Additionally, I think the purpose of `revalidatePath` is to call it when data is mutated, ie, if you have a form submission or button that triggers data updates, and you want to wipe the cache so that the next GET isn't stale.  Since DC is read-only, seems like we can avoid that step.

If this looks okay on QA, I may reevaluate our other uses of revalidatePath as well.

## Ticket:

- JIRA ticket [DR-xxxx](url)

### Checklist:

### DIFF
--- a/app/src/utils/fetchApi/fetchApi.ts
+++ b/app/src/utils/fetchApi/fetchApi.ts
@@ -25,6 +25,7 @@ export const fetchApi = async ({
     clientIP?: string | null;
     isRepoApi?: boolean;
     next?;
+    cache?: string;
   };
 }) => {
   const {
@@ -34,6 +35,7 @@ export const fetchApi = async ({
     clientIP = null,
     isRepoApi = true,
     next,
+    cache = "force-cache",
   } = options;
 
   const headers: Record<string, string> = {};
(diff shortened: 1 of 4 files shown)


app/src/utils/fetchApi/fetchApi.ts
```diff
@@ -34,6 +35,7 @@ export const fetchApi = async ({
     clientIP = null,
     isRepoApi = true,
     next,
+    cache = "force-cache",
```
sarangj, line 38: note - I set this as the default since this is the default in nextJS 14 per the docs

next.config.js
```diff
@@ -40,6 +40,12 @@ const nextConfig = {
   // the modules that newrelic supports should not be mangled by webpack. Thus,
   // we need to "externalize" all of the modules that newrelic supports.
 
+  logging: {
+    fetches: {
+      fullUrl: true,
+    },
+  },
```
sarangj, line 47: this utility only runs in dev mode, but is handy to see when data is *actually* being fetched

keithbauer (approved):

PR #474, opened 2025-10-03
Dr 3796 Refactor search filter tests

jbdalton:
## Ticket:

This replaces the previous PR "Dr 3796 finish final test search filters #443", which I've closed.  The changes here were built on top of 443, but are also adding some additional tests that weren't in that filter-tests set.

The main things this PR does are:

* Separate async and serial process into their own files.
* Replace the monolithic search.spec.ts  with search-items-result, search-filters and search-filters-modal.
* All of the tests that were in search.spec are in one of those 3 files now, with search-filters-modal and search-items-result being devoted only to a single, serialized function.

The other change is that the 3 files all use the base.ts import in order to not hit 3rd party metrics when running on localhost.     

**Smaller changes:**
* changed select id matchers to getByRole
* use Name filter in cleared-filter test and in search item-results
* fixed 'Maps in Education' matcher
* added division and type filter tests

In a future PR, I'll be adding the import from base.ts to the rest of the files in playwright/tests to get global route-filters applied to all of our tests.

## Open questions

The reason I refactored the 2nd-row dropdown clicks to their  own -modal test was that this made it easier manage that "open" modal state which displays the "2nd row and beyond" for the various choices under it that are being tested.  I did have it running successfully earlier in the large search.spec file, but when changing to the base.ts framework and importing route-filters, it became increasingly complex to account for serial vs async sections within the same long-running test.

Ran tests individually, in playwright GUI for debugging, and ran the entire suite after merging qa and doing a new build on this branch.

### Checklist:


playwright/base.ts
```diff
@@ -1,26 +1,20 @@
-import { test as base } from "@playwright/test";
+import { test as base, expect, Page, TestInfo } from "@playwright/test"; // Add Page and TestInfo types
```
clarissarichard: This comment may not be needed as it's implied 🙂

playwright/tests/search-filters-modal.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { test, expect } from "../base";
+import SearchPage from "../pages/search.page";
+import { applyRouteFilters } from "../utils/routeFilters";
+import { Browser, Page } from "@playwright/test"; // <-- Import is only for types
```
clarissarichard, line 4: These `{ Browser, Page }` imports are not being referenced later in the code, so I think this line can be removed!

playwright/tests/search-filters-modal.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  test.describe.serial("choose specific filter options", () => {
+    // Go to search page once, open 2nd row, and check filters, date range and availability
+
+    test.beforeEach(async () => {
+      if (await searchPage.showFilters.isVisible()) {
+        await searchPage.showFilters.click();
+      }
+      // The state is now guaranteed to be 'open' for the test.
+      await expect(searchPage.refineHeading).toBeVisible();
```
clarissarichard, line 37: Would the `hideFilters` button be a better locator here? It serves as another check to confirm the second row of filters are visible.

playwright/tests/search-items-result.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { test, expect } from "../base";
+import SearchPage from "../pages/search.page";
+import { applyRouteFilters } from "../utils/routeFilters";
+import { Browser, Page } from "@playwright/test"; // <-- Import is only for types
```
clarissarichard, line 4: These `{ Browser, Page }` imports are also not being referenced later in the code, so this line can be removed!

playwright/tests/search-filters-modal.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    const browserContext = await browser.newContext();
+    const page = await browserContext.newPage();
+
+    // Apply global filters manually
+    await applyRouteFilters(page);
+
+    // Assign to global variable and load the page object
+    searchPage = new SearchPage(page);
+    await searchPage.loadPage(SearchPage.searchResultsUrl);
+  });
```
clarissarichard, line 21: Just so I understand this better, why would we need to manually set up the browser if it is something that Playwright could do automatically?

playwright/tests/search-filters.spec.ts
```diff
@@ -62,8 +30,7 @@ test.describe("displays search results filters", () => {
     await expect(searchPage.typeFilter).not.toBeVisible();
   });
 
-  test("displays second row of drop-down filters", async ({ page }) => {
-    const searchPage = new SearchPage(page);
+  test("display second row of drop-down filters", async () => {
```
clarissarichard, line 31: Thoughts on removing this test and the date and availability tests below since they are accounted for in `search-filters-modal.spec.ts`?

playwright/tests/search-filters.spec.ts
```diff
@@ -96,8 +61,7 @@ test.describe("displays search results filters", () => {
     await expect(searchPage.availableOnsite).toBeVisible();
   });
 
-  test("hides second row of drop-down filters", async ({ page }) => {
-    const searchPage = new SearchPage(page);
+  test("hide second row of drop-down filters", async () => {
```
clarissarichard, line 62: Thoughts on moving this test to `search-filters-modal.spec.ts`?

playwright/tests/search-filters-modal.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      await expect(searchPage.applyDates).toBeVisible();
+      await searchPage.applyDates.click();
+      await expect(searchPage.startYear).toHaveValue("1700");
+      await expect(searchPage.endYear).toHaveValue("1800");
+    });
+
+    test("choose filter results by availability", async () => {
+      await expect(searchPage.refineHeading).toBeVisible();
+      await expect(searchPage.availablePublicDomain).toBeVisible();
+      await searchPage.availablePublicDomain.click();
```
clarissarichard, line 95: This can also be updated to `check()` since it's a radio button. I like how it adds extra checks to ensure it's visible first!

playwright/tests/search-filters.spec.ts
```diff
@@ -1,47 +1,18 @@
...
-    timeout: 60000,
-  });
+let searchPage: SearchPage; // Global variable declared here
 
-  await expect(page).toHaveTitle("Search results - NYPL Digital Collections");
+test.beforeEach(async ({ page }) => {
+  // 1. Instantiate the Page Object using the provided fixture
+  searchPage = new SearchPage(page);
+  // 2. Call the instance method to load the page
+  await searchPage.loadPage(SearchPage.searchResultsUrl);
```
clarissarichard: These are pretty standard calls that we've used in other tests, so the comments are not needed 🙂

playwright/pages/search.page.ts
```diff
@@ -90,50 +91,74 @@ export default class SearchPage {
...
       .getByText("Maps in education", { exact: true });
     this.topicSelected = this.page
-      .locator("#select-topic")
-      .getByText("Topic: Maps in education", { exact: true });
+      //this matcher gets two hits, could this be because getByLabel is not
+      // limiting the match to within the filter's select area?  So we're going to restrict
+      // to first match for this method, since the 2nd match seems to be erroneously matching
+      // on the global, deletable "filter tags" below these search filter rows.
+      .getByText("Topic: Maps in education")
+      .nth(1);
```
clarissarichard: This is actually matching the text in one of the results!

image

clarissarichard (commented):
Some thoughts and questions!

playwright/pages/search.page.ts
clarissarichard: I'm looking into this right now and an alternative to finding a locator for it could be to reopen the filter and `expect(searchPage.topicOption).toBeChecked();`

playwright/tests/search-items-result.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // Assign to global variable and load the page object
+    searchPage = new SearchPage(page);
+    await searchPage.loadPage(SearchPage.searchResultsUrl);
+  });
+
+  // TEARDOWN: close the entire context.
+  test.afterAll(async () => {
+    // Closes the BrowserContext, freeing up resources.
+    await searchPage.page.context().close();
```
Toxiapo, line 27: Interesting. Should we always manually close resources for the tests? Or this is one-off case for the search page.

playwright/pages/search.page.ts
Toxiapo: We can probably further scoped this to the dropdown container

```ts
this
  .page
  .getByRole('radiogroup', { name: "topic filter options" }
  .getByText("Topic: Maps in education")
```

playwright/tests/search-filters.spec.ts
jbdalton: These don't "operate" the 2nd-row filters, they just verify that the filter choices "display."   Whereas the modal-filter tests take actions on all the "2nd row" (and 3rd, row etc.) filter choices.  It's probably an over zealous distinction to just check whether they display (because it's true that they have to display in order to take an action on them), but that's why I had left it in here.

playwright/tests/search-filters-modal.spec.ts
jbdalton: If the session is set up through the default page, I found that it wouldn't stay in a serialized mode even though that's declared in the test.describe wrapper.  But instead it would default back to truly asyc and try to reload the page for each open-modal locator.  Whereas our goal is for those modal tests to all run serially while the modal stays open instead of getting a new page load for each of these.  Calling browser seems to allow this non-asyc override to "take." 

But this could also be a left-over artifact from an earlier point in their development.   I'll give a shot in my next PR at seeing if the default page inherited from playwright can in fact stay in a serialized, no-reload mode for the open-modal tests.

playwright/pages/search.page.ts
jbdalton: Got this to work

```
    this.topicSelected = this.page
      .getByRole('button', { name: "Select topic" })
      .getByText("Topic: Maps in education")
```

Happy to leave the brittle nth-style locator behind.

playwright/tests/search-filters-modal.spec.ts
jbdalton: There was some issue with making sure Browser is available from default/playwright at one point, but this also could be left over from and earlier development state.  I'll look at pulling this out in the next refactor.

playwright/tests/search-items-result.spec.ts
jbdalton: Not sure if this was for a certain browser/state or just boilerplate.  So I'll look into whether it's already done by default.

playwright/tests/search-filters-modal.spec.ts
jbdalton: I agree!  I'll add it in a cleanup PR.

playwright/tests/search-items-result.spec.ts
jbdalton: I _think_ this might've also been related to the Browser vs Page scenario that came up for serial.test scenarios.  But will look at this again in next round of cleanup too.

playwright/tests/search-filters.spec.ts
jbdalton: removed.

samanthaandrews:
Hey @jbdalton it looks like this was PR merged to the base branch without an approval from a reviewer, which goes against our [PR review guidelines and best practices](https://github.com/NYPL/engineering-general/blob/7d55e2f67871ffc4e0f99ba809e3d87e971c1dbf/standards/peer-review.md). I can see from the comments that you intend to create a follow-up PR to address some of the remaining questions and issues in this PR, which is ok, but you should instead create that PR immediately as a follow up to this one and not merge this one until those issues have been addressed.

I'm going to [revert](https://github.com/NYPL/digital-collections/pull/478) these changes since they were not approved. Please make another PR based on this branch and request review from @clarissarichard. FYI to the team, @clarissarichard will be reviewing and approving Joe's PRs for the time being.

PR #479, opened 2025-10-17
DR-3904 - Checks existence of correct media viewer for images

jbdalton:
## Ticket:

This PR checks whether an image's item page is displaying the correct Universal Viewer media-type controls.

## Open questions

Currently, we are only checking that an type of "still image" does not display inadvertently an audio/video player UV modal.  In future PRs, this can be extended to the reverse: that a video displays its correct controls.   There is a sample UUID for both image and audio types in a utils helper "mediaData.ts" that can be used for extending to video checks.

Right now the UUID variable takes a sample item selected to be in both the qa and prod api data.  In future iterations, it would be nice for this test go be able to take the first random image UUID it finds, and pass that to the "control-checker," and then do the same for a random video item that it finds.  

But as it is, I think the test still fulfills the primary goal:  which is to make sure that the wrong type of UV modal is not being used for an item's stated media type.

Ran tests against both types of UUID (video and image).  Also ran the test suite, and checked with both qa-api and prod-api  settings in the local env.

### Checklist:

### DIFF
--- a/playwright/pages/item-media.page.ts
+++ b/playwright/pages/item-media.page.ts
@@ -34,7 +34,7 @@ export default class ItemMediaPage {
   }
 
   static get itemVideoURL(): string {
-    return `/items/${ItemMediaPage.IMAGE_UUID}`;
+    return `/items/${ItemMediaPage.VIDEO_UUID}`;
   }
 
   constructor(page: Page, expectedType: "IMAGE" | "VIDEO") {
--- a/playwright/pages/item-media.page.ts
+++ b/playwright/pages/item-media.page.ts
@@ -6,23 +6,40 @@ export default class ItemMediaPage {
   static readonly IMAGE_UUID = "4387c9f0-c53c-012f-9924-58d385a7bc34";
   static readonly VIDEO_UUID = "8820d790-e50c-0130-3a92-3c075448cc4b";
 
-  // Structural locators
+  // Set the appropriate content type
+  static get CONTENT_MAP() {
+    return {
+      [ItemMediaPage.IMAGE_UUID]: "IMAGE",
+      [ItemMediaPage.VIDEO_UUID]: "VIDEO",
+    } as const;
+  }
+
+  // Store the expected content-type
+  readonly expectedContentType: "IMAGE" | "VIDEO";
+
+  //Viewer locator
   readonly viewerHeading: Locator;
 
-  // Image Control Locators (Unique to IMAGE viewer mode)
+  // Image control locators (unique to IMAGE viewer mode)
   readonly zoomInButton: Locator;
   readonly rotateRightButton: Locator;
   readonly fullScreenButton: Locator;
 
-  // Video/Audio Control Locators (Unique to VIDEO viewer mode)
+  // Video/Audio control locators (unique to VIDEO viewer mode)
   readonly playButton: Locator;
+  readonly scrubBar: Locator;
 
-  static get itemMediaURL(): string {
+  static get itemImageURL(): string {
     return `/items/${ItemMediaPage.IMAGE_UUID}`;
   }
 
-  constructor(page: Page) {
+  static get itemVideoURL(): string {
+    return `/items/${ItemMediaPage.VIDEO_UUID}`;
+  }
+
+  constructor(page: Page, expectedType: "IMAGE" | "VIDEO") {
     this.page = page;
+    this.expectedContentType = expectedType;
 
     this.viewerHeading = this.page.getByRole("heading", {
       name: "Media Viewer",
@@ -32,17 +49,48 @@ export default class ItemMediaPage {
     this.zoomInButton = page.getByRole("button", { name: "Zoom In" });
     this.rotateRightButton = page.getByRole("button", { name: "Rotate Right" });
     this.fullScreenButton = page.getByRole("button", { name: "Full Screen" });
-
-    this.playButton = page.getByRole("button", { name: "Play" });
+    // this.playButton = page.getByRole("button", { name: "Play" });
+    this.playButton = page.locator("button").filter({ hasText: /^Play$/ });
+    this.scrubBar = page.getByRole("slider", { name: "seek slider" });
   }
 
   async loadPage(gotoPage: string): Promise<void> {
     await this.page.goto(gotoPage);
-    await this.viewerHeading.waitFor({ state: "visible" });
+    // remove this possibly redundant wait?
+    // await this.viewerHeading.waitFor({ state: "visible" });
+  }
+
+  // Waits for the correct viewer type (Image or Video) to stabilize in the DOM.
+  async stabilizeViewer(): Promise<void> {
+    // check for VIDEO content
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible", timeout: 90000 });
+    }
+    // check for IMAGE content
+    else if (this.expectedContentType === "IMAGE") {
+      await this.viewerHeading.waitFor({ state: "visible", timeout: 90000 });
+    }
+    // throw an error if an unknown type is passed
+    else {
+      // If it's neither a video or an image, it SHOULD be an audio, which defaults to the video player.
+      // Later, we can test for audio and throw and error below if an unexpected type rears its head.
+      // throw new Error(`Unknown content type: ${this.expectedContentType}. Cannot determine viewer stabilization locator.`);
+      await this.playButton.waitFor({ state: "visible", timeout: 90000 });
+    }
   }
 
   async verifyViewerIsReady(): Promise<void> {
-    // this will be called in spec to confirm the stability step succeeded.
-    await expect(this.viewerHeading).toBeVisible();
+    // check the appropriate anchor based on the content type
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible" });
+    } else if (this.expectedContentType === "IMAGE") {
+      await this.viewerHeading.waitFor({ state: "visible" });
+    }
+    // If it's not an video or an image, it should be an audio file, which also uses the player
+    else {
+      await this.playButton.waitFor({ state: "visible" });
+      // Or, should we also specify audio, and throw an error if an unknown content type slips in there?
+      // throw new Error(`Unknown content type: ${this.expectedContentType}. Cannot determine viewer stabilization locator.`);
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,49 @@
...
+  readonly zoomInButton: Locator;
+  readonly rotateRightButton: Locator;
+  readonly fullScreenButton: Locator;
+
+  // Video/Audio Control Locators (Unique to VIDEO viewer mode)
+  readonly playButton: Locator;
+
+  static get itemMediaURL(): string {
+    return `/items/${IMAGE_UUID}`;
+  }
```
clarissarichard: Is this function being used somewhere?

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    // CONTROLS: Standard accessible locators for image-specific controls
+    this.zoomInButton = page.getByRole("button", { name: "Zoom In" });
+    this.rotateRightButton = page.getByRole("button", { name: "Rotate Right" });
+    this.fullScreenButton = page.getByRole("button", { name: "Full Screen" });
+
+    //  VIDEO CONTROL (we want to confirm is NOT visible on image pages)
+    this.playButton = page.getByRole("button", { name: "Play" });
+  }
```
clarissarichard: It looks like there are two play buttons in the video viewer that could both be  located with this `getByRole`. Do you happen to know if this still correctly fails if the video viewer were to display?

playwright/tests/item-media.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { test, expect } from "../base";
+import ItemMediaPage from "../pages/item-media.page";
+
+let itemMediaPage: ItemMediaPage;
+
+// configure  suite's timeout settings before the describe block.
+test.describe.configure({ timeout: 90000 });
```
clarissarichard, line 7: Were you intending to use the `waitForSlowResource` function or `SLOW_RESOURCE_TIMEOUT` constant here?

playwright/utils/slowWaitHelpers.ts
```diff
@@ -0,0 +1,18 @@
+import { Page, Locator } from "@playwright/test";
+
+const SLOW_RESOURCE_TIMEOUT = 90000;
```
clarissarichard, line 3: Out of curiosity, why 90s here?

playwright/utils/mediaData.ts
```diff
@@ -0,0 +1,2 @@
+export const IMAGE_UUID = "4387c9f0-c53c-012f-9924-58d385a7bc34";
+export const VIDEO_UUID = "8820d790-e50c-0130-3a92-3c075448cc4b";
```
sarangj: i think if these are just used in one place, we can just keep em in the `page.ts` module for now - that will result in less indirection for the person reading the code.

sarangj (commented):
Niiiice, this is a great test to have

clarissarichard (commented):
Just some questions on declared but unused functions!

playwright/pages/item-media.page.ts
jbdalton: It's going to be.  I've just pushed up another PR cut off of this one that checks both iMAGE_UUID and VIDEO_UUID

playwright/pages/item-media.page.ts
jbdalton: Yes, there are two play buttons.  Good catch!   This is addressed in that next PR already using playwright's suggested locator for it.

playwright/tests/item-media.spec.ts
jbdalton: I found that once the video waits appropriately for the right Play button, I don't really need a waitForSlowResource util. So am thinking of removing that one.   The current timeouts are set in the page-loads i the page-object file in the latest PR.

playwright/utils/slowWaitHelpers.ts
jbdalton: It was just a ballpark number, based on watching how slow the image/universal-viewer seemed to load in the playwright GUI.  It's a failure timeout, so generally that 90s will never be hit during the test run.  Secondly. the test itself doesn't need the IIIF image to actually load, just the Universal Viewer to show up in the DOM.  So this might work ok at 60.  It was failing regularly at 30s in the early going, so I had just tripled it at that point.

playwright/utils/mediaData.ts
jbdalton: I'm imagining maybe one more line for audio will be the extent of this test-data.   Overall, I don't like how brittle these could turn out to be.  There's an opportunity to instead have a test that just grabs a random page off of the "all-items' function, clicks the first result that claims to be an "image" and runs the test against that instead of these pre-canned uuids  But it would need more complex logic to deal with a page that might randomly be all videos, etc.

When I originally set up this data helper, i was imagining a whole long set of UUIDs, but that doesn't really seem necessary, so I'll move them to the page object file.

playwright/pages/item-media.page.ts
jbdalton: This was actually broken out into itemImageURL and itemVideoURL in the followup PR-480 that checks image and video urls.

playwright/pages/item-media.page.ts
jbdalton: I'm going to need to verify this.  I'll feed the "itemImageURL" a "bad" UUID, and see what happens.  I'm realizing that since the previous check for image controls WILL fail if a video player appears here, it also kind of makes this negative assertion test redundant.

playwright/utils/mediaData.ts
jbdalton: moved UUIDs back into the `page.ts` and removed helper util.

playwright/utils/slowWaitHelpers.ts
clarissarichard: Ah I see. If the default timeout for the test is still set to 30s, then setting the action timeout to 90s or even 60s is not needed, since it will never hit that threshold. 

It also doesn't look like we're using the `waitForSlowResource` function in the tests and the Playwright tests are passing on GHA without it, so I think we should remove this helper for the time being and investigate slow resources in a separate PR.

playwright/tests/item-media.spec.ts
clarissarichard: Yes, let's remove the util!

playwright/tests/item-media.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+test.describe("Verify Image Viewer Controls", () => {
+  test.beforeEach(async ({ page }) => {
+    itemMediaPage = new ItemMediaPage(page);
+
+    await itemMediaPage.loadPage(ItemMediaPage.itemMediaURL);
+  });
+
+  test("all image-specific controls are visible", async () => {
+    await itemMediaPage.verifyViewerIsReady();
```
clarissarichard, line 17: You can just assert `await expect(itemMediaPage.viewerHeading).toBeVisible();` here! No need to create a new function if it is just one line.

playwright/tests/item-media.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+test.describe.configure({ timeout: 90000 });
+
+test.describe("Verify Image Viewer Controls", () => {
+  test.beforeEach(async ({ page }) => {
+    itemMediaPage = new ItemMediaPage(page);
+
+    await itemMediaPage.loadPage(ItemMediaPage.itemMediaURL);
+  });
+
+  test("all image-specific controls are visible", async () => {
```
clarissarichard: This test is currently named in a misleading way since there are a few more controls in the image viewer. Either rename this for now or confirm the remaining controls are visible.

playwright/tests/item-media.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    // check if IMAGE Controls are visible
+    await expect(itemMediaPage.zoomInButton).toBeVisible();
+    await expect(itemMediaPage.rotateRightButton).toBeVisible();
+    await expect(itemMediaPage.fullScreenButton).toBeVisible();
+  });
+
+  test("video controls are NOT visible", async () => {
+    // check that no AUDIO/VIDEO controls are appearing (verifying correct UV mode)
+    await expect(itemMediaPage.playButton).toBeHidden();
```
clarissarichard, line 28: You can move this assertion to the test above since they share the same context.

clarissarichard (changes requested):

playwright/pages/item-media.page.ts
clarissarichard: > it also kind of makes this negative assertion test redundant.

I know what you mean, but in the odd case that the play button were to also display with the image buttons, we can catch that glitch. I also recommended in my latest review to just move `expect(playButton).toBeHidden()` to `test("image-specific controls are visible")`.

playwright/tests/item-media.spec.ts
jbdalton: removed

playwright/pages/item-media.page.ts
jbdalton: added as a negative assertion to the relevant image|video 'primary' controls check.

playwright/pages/item-media.page.ts
jbdalton: these two functions are consolidated into a single function in the latest PR: "DR-3904 - Final UV/Media check" (pending)

playwright/utils/slowWaitHelpers.ts
jbdalton: removed

playwright/tests/item-media.spec.ts
jbdalton: This function intention was more for making sure the UV modal itself has actually loaded, since playwright is so quick on the draw during its DOM-based locator matching.  So we're trying to establish "yes, the viewer is ready" then move on to the assertions for the controls, since those often wont be available until the UV modal has fully loaded.  The function also establishes the right "checker" (play for video, viewer-heading for image) to test for whether the modal has fully loaded.

playwright/tests/item-media.spec.ts
jbdalton: renamed to "primary image|video-specifc controls are visible"

playwright/tests/item-media.spec.ts
jbdalton: moved into the shared context.

jbdalton:
Closing this PR.

All required functional code and review comments have been implemented and consolidated into a new branch for final approval.

Reason for Closure: This branch was closed and retired to resolve a complex history divergence and ensure a stable, clean merge path. The code has been migrated after a clean rebase to a fresh branch.

This PR is retired and closed. Please direct all further review and approval to the new PR:

https://github.com/NYPL/digital-collections/pull/481

PR #480, opened 2025-10-17
DRAFT - Dr-3904 - test media buttons for Video

jbdalton:
## Ticket:

## This DRAFT PR does the following:

This PR refactors the work in PR #479 

It adds video to the image checks, and accounts for the brittle play button locator.

I'm going to refactor this based on  PR 479's feedback.

## Open questions

Questions were in PR #479.  This is mainly just a refactor/cleanup PR.

Ran the test suite locally.

### Checklist:

### DIFF
--- a/playwright/tests/item-media.spec.ts
+++ b/playwright/tests/item-media.spec.ts
@@ -13,7 +13,7 @@ test.describe("Verify Image Viewer Controls", () => {
     await itemMediaPage.loadPage(ItemMediaPage.itemImageURL);
   });
 
-  test("all image-specific controls are visible", async () => {
+  test("primary image-specific controls are visible", async () => {
     await itemMediaPage.verifyViewerIsReady();
 
     // check if IMAGE Controls are visible
@@ -22,7 +22,8 @@ test.describe("Verify Image Viewer Controls", () => {
     await expect(itemMediaPage.fullScreenButton).toBeVisible();
   });
 
-  test("video controls are NOT visible", async () => {
+  // flakey test because it checks before ANY controls are visible, so will always pass
+  test.skip("video controls are NOT visible", async () => {
     // check that no AUDIO/VIDEO controls are appearing (verifying correct UV mode)
     await expect(itemMediaPage.playButton).toBeHidden();
   });
@@ -35,7 +36,7 @@ test.describe("Verify Video Viewer Controls", () => {
     await itemMediaPage.loadPage(ItemMediaPage.itemVideoURL);
   });
 
-  test("all video-specific controls are visible", async () => {
+  test("primary video-specific controls are visible", async () => {
     await itemMediaPage.verifyViewerIsReady();
 
     // check if VIDEO Controls are visible
@@ -43,8 +44,18 @@ test.describe("Verify Video Viewer Controls", () => {
     await expect(itemMediaPage.playButton).toBeVisible();
   });
 
-  test("image controls are NOT visible", async () => {
-    // check that no IMAGE controls are appearing (verifying correct UV mode)
+  // unnecessary assertions, which also will get false-positives in async mode
+  test.skip("image controls are NOT visible", async () => {
+    // This will get a false positive in the DOM for the same reasons as
+    // the negative-assertion about video-controls will similarly fail in
+    // the main image-controls test.
+    // We could serially wait for the answer to the previous
+    // assertion (is media-play-button visible?) here, first.  But we
+    // can probably leave these negative assertions below out, because if
+    // you are on a video player, it's all or nothing:  you're not going to see
+    // ANY of the image controls we are checking
+    // for below on the video-player UV.  And the reverse holds true
+    // for the image-viewer UV.
     await expect(itemMediaPage.zoomInButton).toBeHidden();
     await expect(itemMediaPage.rotateRightButton).toBeHidden();
     await expect(itemMediaPage.fullScreenButton).toBeHidden();


samanthaandrews:
Since this PR builds off of #479, you should change the base branch of this PR to make that clear. I've gone ahead and done it in this one, but in the future @jbdalton you should do this to make review easier for teammates. Screenshot of how to change the base branch in this PR ui:
Screenshot 2025-10-21 at 3 18 25 PM

playwright/pages/item-media.page.ts
```diff
@@ -1,49 +1,94 @@
 import { Locator, Page, expect } from "@playwright/test";
```
clarissarichard, line 1: `expect` is not being used and should be removed.

playwright/pages/item-media.page.ts
```diff
@@ -1,49 +1,94 @@
...
 
-  static get itemMediaURL(): string {
+  static get itemImageURL(): string {
     return `/items/${IMAGE_UUID}`;
   }
 
-  constructor(page: Page) {
+  static get itemVideoURL(): string {
+    return `/items/${VIDEO_UUID}`;
+  }
```
clarissarichard, line 38: These are nearly identical functions and should be condensed into one function. Something like:
```
static itemURL(uuid: string): string {
  return `/items/${uuid}`;
}
```

playwright/pages/item-media.page.ts
```diff
@@ -1,49 +1,94 @@
...
-    await this.viewerHeading.waitFor({ state: "visible" });
+    // remove this possibly redundant wait?
+    // await this.viewerHeading.waitFor({ state: "visible" });
+  }
+
+  // Waits for the correct viewer type (Image or Video) to stabilize in the DOM.
+  async stabilizeViewer(): Promise<void> {
+    // check for VIDEO content
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible", timeout: 90000 });
```
clarissarichard, line 67: If the test is still set to the default 30s, this button will also be capped at 30s.

playwright/tests/item-media.spec.ts
```diff
@@ -27,3 +27,26 @@ test.describe("Verify Image Viewer Controls", () => {
...
+    // check if VIDEO Controls are visible
+
+    await expect(itemMediaPage.playButton).toBeVisible();
+  });
+
+  test("image controls are NOT visible", async () => {
+    // check that no IMAGE controls are appearing (verifying correct UV mode)
+    await expect(itemMediaPage.zoomInButton).toBeHidden();
+    await expect(itemMediaPage.rotateRightButton).toBeHidden();
+    await expect(itemMediaPage.fullScreenButton).toBeHidden();
```
clarissarichard: These assertions can be moved to the test above since they also share the same context.

playwright/pages/item-media.page.ts
```diff
@@ -1,49 +1,94 @@
...
+  // Set the appropriate content type
+  static get CONTENT_MAP() {
+    return {
+      [IMAGE_UUID]: "IMAGE",
+      [VIDEO_UUID]: "VIDEO",
+    } as const;
+  }
+
+  // Store the expected content-type
+  readonly expectedContentType: "IMAGE" | "VIDEO";
```
clarissarichard, line 18: I think this content map is overcomplicating the tests at this moment in time since we only have one image UUID and one video UUID.

playwright/utils/mediaData.ts
```diff
@@ -1,2 +1,3 @@
 export const IMAGE_UUID = "4387c9f0-c53c-012f-9924-58d385a7bc34";
-export const VIDEO_UUID = "8820d790-e50c-0130-3a92-3c075448cc4b";
+// export const VIDEO_UUID = "8820d790-e50c-0130-3a92-3c075448cc4b";
+export const VIDEO_UUID = "c57ddb40-e50c-0130-575a-3c075448cc4b";
```
clarissarichard: Since we only have two constants, I think we can move these back to `item-media.page.ts` for simplicity.

playwright/tests/item-media.spec.ts
```diff
@@ -27,3 +27,26 @@ test.describe("Verify Image Viewer Controls", () => {
...
+    await itemMediaPage.loadPage(ItemMediaPage.itemVideoURL);
+  });
+
+  test("all video-specific controls are visible", async () => {
+    await itemMediaPage.verifyViewerIsReady();
+
+    // check if VIDEO Controls are visible
+
+    await expect(itemMediaPage.playButton).toBeVisible();
+  });
```
clarissarichard: Similar to the image test, this test is currently named in a misleading way since there are a couple more controls in the video viewer. Either rename this for now or confirm the remaining controls are visible.

playwright/pages/item-media.page.ts
```diff
@@ -1,49 +1,94 @@
...
 
-    // CONTROLS: Standard accessible locators for image-specific controls
     this.zoomInButton = page.getByRole("button", { name: "Zoom In" });
     this.rotateRightButton = page.getByRole("button", { name: "Rotate Right" });
     this.fullScreenButton = page.getByRole("button", { name: "Full Screen" });
-
-    //  VIDEO CONTROL (we want to confirm is NOT visible on image pages)
-    this.playButton = page.getByRole("button", { name: "Play" });
+    // this.playButton = page.getByRole("button", { name: "Play" });
+    this.playButton = page.locator("button").filter({ hasText: /^Play$/ });
```
clarissarichard, line 53: Interesting how these are somehow not the same selector! A way to select the play button next to the slider would be
```this.playButton = page.locator(".plyr__controls").getByRole("button", { name: "Play", exact: true });```

I recommend choosing one and noting which play button it is either by renaming the locator or adding a small comment.

File: playwright/pages/item-media.page.ts [hunk 9]
```diff
@@ -1,49 +1,94 @@
...
+    this.scrubBar = page.getByRole("slider", { name: "seek slider" });
   }
 
   async loadPage(gotoPage: string): Promise<void> {
     await this.page.goto(gotoPage);
-    await this.viewerHeading.waitFor({ state: "visible" });
+    // remove this possibly redundant wait?
+    // await this.viewerHeading.waitFor({ state: "visible" });
+  }
```
clarissarichard, line 61: Do you happen to still find this `loadPage` function helpful? 

Additionally, if we remove the `waitFor`, then I think we can remove the entire `loadPage` function and just use `goto`.

playwright/tests/item-media.spec.ts
```diff
@@ -27,3 +27,26 @@ test.describe("Verify Image Viewer Controls", () => {
...
+
+test.describe("Verify Video Viewer Controls", () => {
+  test.beforeEach(async ({ page }) => {
+    itemMediaPage = new ItemMediaPage(page, "VIDEO");
+
+    await itemMediaPage.loadPage(ItemMediaPage.itemVideoURL);
+  });
+
+  test("all video-specific controls are visible", async () => {
+    await itemMediaPage.verifyViewerIsReady();
```
clarissarichard, line 40: Could we just assert something like `expect(viewerHeading).toBeVisible()` instead of calling this function?

playwright/pages/item-media.page.ts
```diff
@@ -1,49 +1,94 @@
...
+
+  // Waits for the correct viewer type (Image or Video) to stabilize in the DOM.
+  async stabilizeViewer(): Promise<void> {
```
clarissarichard, line 64: Is this function being used?

clarissarichard (changes requested):
I know this is still in draft, but I wanted to share my initial thoughts before I'm OOO. I think a lot of the functions can be replaced by a single line in the tests. While there may be a use case for them sometime in the future, let's choose the simplest route first!

PR #481, opened 2025-10-24
DR-3904 - Final UV/Media check

jbdalton:
## Ticket:

This PR checks whether an image's item page will display the expected Universal Viewer for its media-type.  It does a similar UV/Player check for a publicly-viewable video item.

## Open questions

This PR contains stable code for the UV Controls test, migrated from the working code previously in DR-3904/test-media-buttons-ami.   I needed to rebase from the previous branch when history diverged after its remote tracking was changed (tl;dr I made some commits before a pull  and found myself in a long, tedious rebase).  So closing the previous PR and opening this new one seemed to be the most direct path forward in this case.

These tests have been migrated to a new branch to resolve the history divergence and ensure a clean merge path. The entire working code from the previous branch is present here.

All functional changes, including the fixes based on comments in the previous PRs, have been incorporated and verified in this branch. 

Any new questions/comments should go here, and I'll close the previous PR.

All current tests were run on localhost.

### Checklist:


playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    this.viewerHeading = this.page.getByRole("heading", {
+      name: "Media Viewer",
+      level: 2,
+    });
+
+    this.zoomInButton = page.getByRole("button", { name: "Zoom In" });
+    this.rotateRightButton = page.getByRole("button", { name: "Rotate Right" });
+    this.fullScreenButton = page.getByRole("button", { name: "Full Screen" });
+    // this.playButton = page.getByRole("button", { name: "Play" });
```
samanthaandrews: i prefer not to include commented out code. if this isn't needed, we should remove it entirely so as to reduce confusion for anyone working on these tests in the future.

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,87 @@
...
+    // this.scrubBar = page.getByRole("slider", { name: "seek slider" });
+  }
+
+  async loadPage(uuid: string): Promise<void> {
+    const gotoPage = `/items/${uuid}`;
+    await this.page.goto(gotoPage);
+  }
+
+  // Waits for the correct viewer type (Image or Video) to stabilize in the DOM.
+  async stabilizeViewer(): Promise<void> {
```
samanthaandrews: it doesn't look like this function is being used anywhere in this PR so i don't understand why it's here. if it's not being used, we should remove it.

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,87 @@
...
+  readonly rotateRightButton: Locator;
+  readonly fullScreenButton: Locator;
+
+  // Video/Audio control locators (unique to VIDEO viewer mode)
+  readonly playButton: Locator;
+
+  // Would the scrubBar appear if the video is an on-site, restricted item?
+  // We should have other tests that check this under Rights-Checks conditions.
+  // Here we are only concerned with the Play-button on a known public-video.
+  // readonly scrubBar: Locator;
```
samanthaandrews: If these are questions you have about how the content should display, you should ask them in a comment in this PR and tag an engineer who might know the answer and remove them from the code comments as they make things more confusing to people reading the tests in the future

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,87 @@
...
+  async loadPage(uuid: string): Promise<void> {
+    const gotoPage = `/items/${uuid}`;
+    await this.page.goto(gotoPage);
+  }
+
+  // Waits for the correct viewer type (Image or Video) to stabilize in the DOM.
+  async stabilizeViewer(): Promise<void> {
+    // check for VIDEO content
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible", timeout: 90000 });
```
samanthaandrews: It looks like we are putting 90000 timeouts both in this file and in the spec file, which seems like unnecessary duplication to me. Can you refactor slightly to make this into a constant that can be used across both files?

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,87 @@
...
+    // check the appropriate anchor based on the content type
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible" });
+    } else if (this.expectedContentType === "IMAGE") {
+      await this.viewerHeading.waitFor({ state: "visible" });
+    }
+    // If it's not an video or an image, it should be an audio file, which also uses the player
+    else {
+      await this.playButton.waitFor({ state: "visible" });
+      // Or, should we also specify audio, and throw an error if an unknown 4th content type slips in there?
```
samanthaandrews: Once again, let's get clarity here from an engineer instead of keeping this comment in the code

samanthaandrews (changes requested):
This PR is a step in the right direction as it is small and easy for a reviewer to comment on. Some small changes requested.

playwright/pages/item-media.page.ts
jbdalton: will remove.  They were more comments/questions for me to keep track of in code. Normally I clean out these types of internal comments for the last push that pulls in the latest QA branch before the PR-merge/deploy.

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,87 @@
...
+      await this.playButton.waitFor({ state: "visible", timeout: 90000 });
+    }
+  }
+
+  async verifyViewerIsReady(): Promise<void> {
+    // check the appropriate anchor based on the content type
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible" });
+    } else if (this.expectedContentType === "IMAGE") {
+      await this.viewerHeading.waitFor({ state: "visible" });
```
clarissarichard: Does the `viewerHeading` also display for videos? If so, is there a different element unique to images that you could look for to ensure ~~that it doesn't find the `viewerHeading` before the `playButton`~~ the image UUID isn't accidentally displaying the video player?

playwright/pages/item-media.page.ts
jbdalton: Using only one function now.

Removed the `verifyViewerIsReady()` function below this one.  It was from an earlier attempt to separate the "loading guarantee" (UV fully loads so that relevant locators are visible) from the "locator checking."  But in reality the `stabilizeViewer` function does both: 1.) makes sure the UV loads and doesn't just rely on a DOM display of "Media Viewer," and 2.) it does this by waiting for at least one of the UV IMAGE|VIDEO control-tools to display.

Also renamed `stabilizeViewer` to `getViewerControls` to make more clear what it does.

playwright/pages/item-media.page.ts
jbdalton: removed.

playwright/pages/item-media.page.ts
jbdalton: After testing on localhost in both prod and dev builds, I feel pretty confident about removing _all_ of the 90 sec timeouts in these tests.   They were from a period where I thought the UV needed that time to load.  It did need _some_ time, and it did need some appropriate locators to allow for for it not to fail on waits, but I'm finding now that the default 30 sec timeout suffices.

playwright/pages/item-media.page.ts
jbdalton: removed comments.

playwright/pages/item-media.page.ts
jbdalton: Changed this to `await this.zoomInButton.waitFor` ...  which is a unique control for the `IMAGE` content type.

jbdalton:
I think this PR is ready for another look?  Ran it against prod build, and in a slower dev-build mode to verify that timeouts are sufficient.  Also tested negative cases by temporarily adding in purposely bad UUIDs (to "test the test").  This verified that the logic and locators relevant to their content types are both correct for passing tests.

playwright/pages/item-media.page.ts
jbdalton: commented-out code removed.

samanthaandrews (approved):
LGTM!

playwright/pages/item-media.page.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (this.expectedContentType === "VIDEO") {
+      await this.playButton.waitFor({ state: "visible" });
+    }
+    // check for IMAGE content
+    else if (this.expectedContentType === "IMAGE") {
+      await this.zoomInButton.waitFor({ state: "visible" });
+    }
+    // check for 3rd content-type fallback (currently AUDIO)
+    else {
+      await this.playButton.waitFor({ state: "visible" });
```
clarissarichard, line 59: I replaced the video link with `28a04050-c32b-0133-1433-60f81dd2b63c` to see how this would fallback and it was unable to find the play button.

ETA: I wonder if it's because the audio button has the "Play" text _**and**_ the "Pause" text.

jbdalton: I'm going to add an Audio-specific content-type check, based on this.  Then each type should be able to test exclusively that it's not showing the "wrong" viewer.   The current tests I'll push up in this PR will at least for now cover the Prod bug we had seen recently where all images were showing a "play" button.

The audio-check will be in this new Jira ticket: [DR-3935](https://newyorkpubliclibrary.atlassian.net/browse/DR-3935)

PR #483, opened 2025-10-31
DR-3796 fix revert of search filters

jbdalton:
## Ticket:

This PR updates the reverted [PR-478](https://github.com/NYPL/digital-collections/pull/478).  I've also added a missing conditional check on "Applied Filters" to verify the correct filter remains "active/visible" in the "Applied Filters" section after the other is closed.

Otherwise this is the same set of tests that were added in the original PR.   I'm mainly just "reverting the revert" on a new branch from qa, and pulling in the test-files from the original PR-478.  Any questions/comments either carried over from the closed PR-478 or new can go in this PR.

## Open questions

Mostly covered in the approved, prior [PR-478](https://github.com/NYPL/digital-collections/pull/478). 

It should be noted there is a manual page/browser object required to enforce the serial behavior that we are using in two of the spec files (search-filters-modal.spec.ts and search-items-result.spec.ts).  The other "main" spec file, search.spec.ts, uses the POM page import conventionally since the serialization is isolated to the two other "search-<type>" specs.

Ran the suite against qa-collections-api and ran it on a prod build and a dev build.

### Checklist:

### DIFF
--- a/playwright/pages/search.page.ts
+++ b/playwright/pages/search.page.ts
@@ -50,6 +50,7 @@ export default class SearchPage {
   readonly clearFilterButton: Locator;
   readonly clearNameFilterApplied: Locator;
   readonly clearTopicFilterApplied: Locator;
+  readonly clearPublisherFilterApplied: Locator;
   readonly clearAllFilters: Locator;
 
   // sort search results
@@ -184,6 +185,9 @@ export default class SearchPage {
     });
     this.clearNameFilterApplied = this.page.getByRole("button", {
       name: "Sheldonian Theatre, click to remove filter",
+    });
+    this.clearPublisherFilterApplied = this.page.getByRole("button", {
+      name: "Printed at the Theater,, click to remove filter",
     });
     this.clearAllFilters = this.page
       .locator("#search-filter-tags")
--- a/playwright/pages/search.page.ts
+++ b/playwright/pages/search.page.ts
@@ -34,6 +34,7 @@ export default class SearchPage {
   readonly publisherSelected: Locator;
   readonly divisionFilter: Locator;
   readonly divisionOption: Locator;
+  readonly divisionSelected: Locator;
   readonly typeFilter: Locator;
   readonly typeOption: Locator;
   readonly typeSelected: Locator;
@@ -47,6 +48,7 @@ export default class SearchPage {
   readonly hideFilters: Locator;
   readonly applyFilterButton: Locator;
   readonly clearFilterButton: Locator;
+  readonly clearNameFilterApplied: Locator;
   readonly clearTopicFilterApplied: Locator;
   readonly clearAllFilters: Locator;
 
@@ -90,20 +92,22 @@ export default class SearchPage {
     this.refineHeading = this.page.getByRole("heading", {
       name: "Refine your search",
     });
+
     this.topicFilter = this.page.getByRole("button", { name: "Topic" });
     this.topicOption = this.page
-      .locator("#select-topic")
+      .getByLabel("topic filter options")
       .getByText("Maps in education", { exact: true });
     this.topicSelected = this.page
-      .locator("#select-topic")
-      .getByText("Topic: Maps in education", { exact: true });
+      .getByRole("button", { name: "Select topic" })
+      .getByText("Topic: Maps in education");
+
     this.nameFilter = this.page.getByRole("button", { name: "Name" });
     this.nameOption = this.page
-      .locator("#select-name")
+      .getByLabel("name filter options")
       .getByText("Sheldonian Theatre", { exact: true });
-    this.nameSelected = this.page
-      .locator("#select-name")
-      .getByText("Name: Sheldonian Theatre", { exact: true });
+    this.nameSelected = this.page.getByText("Name: Sheldonian Theatre", {
+      exact: true,
+    });
     this.collectionFilter = this.page.getByRole("button", {
       name: "Collection",
     });
@@ -111,7 +115,7 @@ export default class SearchPage {
       .getByLabel("collection filter options")
       .getByText("Portolan atlas");
 
-    // The selected filter check below currently uses a fuzzy match
+    // The selected collection-filter check below currently uses a fuzzy match
     // because the collection UUID is appended to the
     // collectionSelected display text, which is visible to users when
     // collections have short titles. (DR-3838)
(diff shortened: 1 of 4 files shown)


samanthaandrews:
Hey @jbdalton thanks for opening this up again. Can you explain why we're using `.serial` in these tests? It's not quite clear to me from reading the code but I am also not super familiar with this part of the Playwright API.

playwright/tests/search-items-result.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      await searchPage.searchButton.click();
+
+      // Page actions use searchPage.page
+      await searchPage.page.waitForURL("/search/**", {
+        waitUntil: "load",
+      });
+
+      await expect(searchPage.page).toHaveTitle(
+        "Search results - NYPL Digital Collections"
+      );
```
clarissarichard, line 61: Could we also confirm the page loads by asserting an element displays on the page?

playwright/tests/search-items-result.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      await searchPage.searchBar.fill(searchPage.searchKeyword);
+      await expect(searchPage.searchBar).toHaveValue(searchPage.searchKeyword);
+      await expect(searchPage.searchButton).toBeVisible();
+
+      await searchPage.searchButton.click();
+
+      // Page actions use searchPage.page
+      await searchPage.page.waitForURL("/search/**", {
+        waitUntil: "load",
+      });
```
clarissarichard: Do we already have a function for this?

playwright/pages/search.page.ts
```diff
@@ -216,16 +242,16 @@ export default class SearchPage {
   }
```
clarissarichard, line 242: We can just call `goto()` directly in the tests with the same effect.

playwright/tests/search-filters-modal.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      await searchPage.startYear.fill("1700");
+      await searchPage.endYear.fill("1800");
+      await expect(searchPage.applyDates).toBeVisible();
+      await searchPage.applyDates.click();
+      await expect(searchPage.startYear).toHaveValue("1700");
+      await expect(searchPage.endYear).toHaveValue("1800");
+    });
+
+    test("choose filter results by availability", async () => {
+      await expect(searchPage.refineHeading).toBeVisible();
```
clarissarichard: This assertion is already in the `beforeEach()` above.

playwright/tests/search-filters-modal.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      await expect(searchPage.availablePublicDomain).toBeVisible();
+      await searchPage.availablePublicDomain.click();
```
clarissarichard: Could we use `check()` for this?

clarissarichard (commented):
Don't forget to update the changelog!

jbdalton:
> Hey @jbdalton thanks for opening this up again. Can you explain why we're using `.serial` in these tests? It's not quite clear to me from reading the code but I am also not super familiar with this part of the Playwright API.

I'm using serial in the two additional filter-tests because we would like the context to remain on the same session for those tests.  And without `.serial` the test will default to the `async` mode.   The "serial" isn't so much to run the checks one after another, as it is to run them _within that same page visit_ in this case.   Playwright's default is to reload a new page for every test within a spec; but we want the extra filters, for example, in search-filters-modal.spec.ts to all run while the modal for those choices is only opened once.

Without using the serial attribute on the describe wrapper for search-filters-modal-spec.ts, I found that playwright defaults to async calls:

1. load the page for test-1, 
2. open modal-1
3. check test/filter-1
4. load another page for test-2 (via another worker)
5. open modal-1
6. check test/filter-2

... etc ...

17. open modal-1
18. check test/filter-6



With `serial.test` it does this (using a single worker, and with one page load):

1. load the page
2. open modal-1 one time
3. check test-1/filter-1
4. check test-2/filter-2
5. check test-3/filter-3
6. check test-4/filter-4
7. check test-4/filter-5
8. check test-4/filter-6

Both methods will verify the tests, but the second one will take fewer workers, and not require 5 more new page loads for a check of 6 filters in a single open modal.

As the test numbers increase in the overall suite, it should also be faster. Although the main benefit is in being able to load a single page, open 1 modal, and check the filters in that open modal within a single operation/session.

In terms of speed/timings, I ran 6 sample tests with `serial` and without it.   The serialized tests ran an average of 16.57 sec. while the `async` (default) tests without `serial` ran an average of 26.23 seconds (used 6 workers), for an almost 60% increase in speed.

jbdalton:
> Don't forget to update the changelog!

I think the changelog is now automatically updated for digital-collections now?   😅

playwright/pages/search.page.ts
jbdalton: I agree that loadPage() is verbose for just one line. 

But I've put it in the Page Object file specifically so that if we ever need to add, for example, some prerequisite before navigation (e.g., setting a user session cookie, checking for a preload script, etc.), we'll only have to edit the logic in the Page Object once. Currently there are 3 spec files using the search.page POM file.

playwright/tests/search-items-result.spec.ts
jbdalton: Can those elements be inferred  in the subsequent test block, which is checking for clickable items?  The basic scenario in this block is:  

1. the result-page should have a title that says it's a "search-results-page," (there isn't a specific describe.title to call this out, because it's mainly to pre-check that we got to the right a landing place -- if this assertion fails something has seriously gone wrong ... or someone has changed the result titles `<head>` behind our back 😅 ), 
2. there should be clickable images on that sample results page.
3. those clicks should take a user to items.

clarissarichard: Ah I see what you mean and that makes sense to me with these serial tests.

playwright/tests/search-filters-modal.spec.ts
jbdalton: removed.

playwright/tests/search-filters-modal.spec.ts
jbdalton: changed to `check()`

jbdalton:
Pushed commits with the following changes: add check() and remove duplicate filter assertion for heading.

I think this is ready for final review now?

samanthaandrews:
> I'm using serial in the two additional filter-tests because we would like the context to remain on the same session for those tests. And without `.serial` the test will default to the `async` mode. The "serial" isn't so much to run the checks one after another, as it is to run them _within that same page visit_ in this case. Playwright's default is to reload a new page for every test within a spec; but we want the extra filters, for example, in search-filters-modal.spec.ts to all run while the modal for those choices is only opened once.
> 
> Without using the serial attribute on the describe wrapper for search-filters-modal-spec.ts, I found that playwright defaults to async calls:
> 
> 1. load the page for test-1,
> 2. open modal-1
> 3. check test/filter-1
> 4. load another page for test-2 (via another worker)
> 5. open modal-1
> 6. check test/filter-2
> 
> ... etc ...
> 
> 17. open modal-1
> 18. check test/filter-6
> 
> With `serial.test` it does this (using a single worker, and with one page load):
> 
> 1. load the page
> 2. open modal-1 one time
> 3. check test-1/filter-1
> 4. check test-2/filter-2
> 5. check test-3/filter-3
> 6. check test-4/filter-4
> 7. check test-4/filter-5
> 8. check test-4/filter-6
> 
> Both methods will verify the tests, but the second one will take fewer workers, and not require 5 more new page loads for a check of 6 filters in a single open modal.
> 
> As the test numbers increase in the overall suite, it should also be faster. Although the main benefit is in being able to load a single page, open 1 modal, and check the filters in that open modal within a single operation/session.
> 
> In terms of speed/timings, I ran 6 sample tests with `serial` and without it. The serialized tests ran an average of 16.57 sec. while the `async` (default) tests without `serial` ran an average of 26.23 seconds (used 6 workers), for an almost 60% increase in speed.

Thank you @jbdalton this is more clear to me now. An alternate option that might be clearer from a code readability perspective would be to combine all of these filter tests into one larger test. I am ok with keeping it this way, but let's add a comment docstring within the `.serial` code blocks to explain this clearly. Something like this:

```
JUSTIFICATION FOR .serial USAGE:
These tests intentionally use .serial to maintain browser state across multiple test steps. This deviates from Playwright's recommendation for isolated tests for the following reasons:
1. Performance: Avoids 6 separate page loads (60% faster: 16.57s vs 26.23s)
2. Resource efficiency: Uses single worker instead of 6 parallel workers
3. User journey testing: Simulates realistic workflow of opening modal once and applying multiple filters in sequence
4. State accumulation: Each filter builds on previous selections

Alternative considered: Single comprehensive test, but rejected because individual filter failures would be harder to debug.
```

samanthaandrews:
@jbdalton just add that comment block to the spots where we are using `.serial` and I'll approve!

clarissarichard (approved):
LGTM!

samanthaandrews (approved):
🚀

jbdalton:
Merged in the latest from qa before this final push.   Tests all passed on localhost ... (against prod build), but failed in CI here.  Fun times.

samanthaandrews:
> Merged in the latest from qa before this final push. Tests all passed on localhost ... (against prod build), but failed in CI here. Fun times.

I clicked the button to re-run them. Hopefully they pass this time around! 🤞Otherwise might just need to address whatever is flaky.

jbdalton:
Second time's the charm.  Tests passed on re-try, and I'm going to finally merge these filter tests.

PR #484, opened 2025-11-06
fix "see more" links with ampersands on division page

avertrees:
## Ticket:

- JIRA ticket [DR-3912](https://newyorkpubliclibrary.atlassian.net/browse/DR-3912)

updates the method used to convert division names to dc links to url encode &

### Checklist:

[DR-3912]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/utils/utils.ts
+++ b/app/src/utils/utils.ts
@@ -89,7 +89,16 @@ export const stringToSlug = (string: string = ""): string => {
 };
 
 export const titleToDCParam = (string: string = ""): string => {
-  return string?.replace(/\s+/g, "+"); // replace spaces with +
+  // replace space with + and & with '%26'
+  return string?.replace(/(\s+)|(&)/g, (match, whitespace, ampersand) => {
+    if (ampersand) {
+      return "%26";
+    }
+    if (whitespace) {
+      return "+";
+    }
+    return match;
+  });
 };
 
 export const totalNumPages = (numResults: string, perPage: number): number => {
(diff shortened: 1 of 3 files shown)


app/src/utils/utils.ts
```diff
@@ -89,7 +89,16 @@ export const stringToSlug = (string: string = ""): string => {
...
+  // replace space with + and & with '%26'
+  return string?.replace(/(\s+)|(&)/g, (match, whitespace, ampersand) => {
+    if (ampersand) {
+      return "%26";
+    }
+    if (whitespace) {
+      return "+";
+    }
+    return match;
+  });
```
sarangj: this works, but I wonder if we could just use the JS builtin [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)?

sarangj (approved):

app/src/utils/utils.ts
avertrees: I was thinking that as well bc it would be cleaner but was worried it would break the filters (since they require the +) but I can try it and see if it breaks anything?

avertrees: tried it and it didn't break anything obvious so pushing it up. think we should try it in vercel before qa (and also before fixing this broken test)

PR #485, opened 2025-11-06
DR-3943 - Add base fixture import to all spec files

jbdalton:
## Ticket:

- JIRA ticket [DR-3943](https://newyorkpubliclibrary.atlassian.net/browse/DR-3943)

The base fixture framework (it's used to import routeFilters that block external links like adobe, etc.) was set up during work on the timeouts issue PR earlier.  But it hadn't been linked to all spec files yet.   Given that we have seen occasional flakiness recently on the CI, I think now is a good time to make sure all of our specs are at a minimum pulling from the base import so they can all block on the same routeFilters list.

## Open questions

This required a 1-line change to four spec files, so hopefully not too much to review here.

I also checked the network console for each spec in the playwright GUI, to make sure there weren't any new 3rd party links. 

Ran affected specs in network console, and ran the whole suite.

### Checklist:

[DR-3943]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/tests/about.spec.ts
+++ b/playwright/tests/about.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "../base";
 import { AboutPage } from "../pages/about.page";
 
 test.beforeEach(async ({ page }) => {
--- a/playwright/tests/divisions.spec.ts
+++ b/playwright/tests/divisions.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "../base";
 import { DivisionsPage } from "../pages/divisions.page";
 
 test.beforeEach(async ({ page }) => {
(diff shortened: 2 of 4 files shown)


jbdalton:
Tests passed in CI on the 2nd run, which seems to be the theme of the day.   My research is showing that we may need to modify the CI processes via `playwright.config` by setting up some warming so that playwright doesn't fire off until the environment is truly stable.

I think this standardized base import can be reviewed separate from that, and I'll make another PR for stabilizing the testing CI.

And while we're in there, it might also be a good time to have the CI not download and package all three browsers, if we're only planning on using chromium for the foreseeable future.

clarissarichard (approved):
Sorry it took so long for me to get to this - looks good!

Toxiapo (approved):

PR #486, opened 2025-11-12
[DRAFT] NO-REF: Stabilize Playwright tests on ci

jbdalton:
## Ticket:

- No JIRA ticket 

This PR addresses persistent test flakiness and timeout failures observed in the CI pipeline, especially under resource-constrained (`--workers=2`) environments.

**1. Synchronization of waits**: Adds targeted `waitFor({ state: 'hidden' })` calls after key data-altering clicks (e.g., sorting). This forces Playwright to wait for the UI to settle after an asynchronous action, eliminating timeouts observed on the 2-worker test-suite runs for the existing search.spec file.

**2. Standardize Timeout Buffer**: Increases the global `expect` action timeout in playwright.config.ts to 45000ms. This prevents the entire test run from failing due to the slow server response time experienced under low CI concurrency. Overall timeout also increased to 60000ms.  This is not a hard wait, it just allows the test-run to have a better chance at succeeding in the _first_ run so that time-consuming retries are not occurring as often on the CI.

**3. Code Cleanup**: Removes conflicting, hardcoded local timeouts from search.spec.ts.

Run this on localhost with 2 workers (eg. `npx playwright test --workers=2`) and with 6 workers against a prod build (`npm run build`).  Before the code-changes, the 2 worker test-runs consistently were timing out/failing on the sort-result blocks in the search.spec file, whether it was ran alone or as part of a suite.

### Checklist:

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -14,8 +14,8 @@ import { defineConfig, devices } from "@playwright/test";
 
 export default defineConfig({
   testDir: "./playwright",
-  timeout: 30000, // default timeout for each test
-  expect: { timeout: 20000 }, // default timeout for each expect assertion
+  timeout: 60000, // default timeout for each test
+  expect: { timeout: 45000 }, // default timeout for each expect assertion
   /* Run tests in files in parallel */
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
--- a/playwright/base.ts
+++ b/playwright/base.ts
@@ -23,4 +23,15 @@ export const test = base.extend<RouteFilterFixtures>({
   ],
 });
 
+// Teardown with global afterEach cleanup hook
+test.afterEach(async ({ page }) => {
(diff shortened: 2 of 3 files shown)


jbdalton:
Going to switch this PR to DRAFT, because I can see, based on the _actual CI run_ that we need an adjusted "wait for" strategy a little more for those data-clicks.  I have something in mind, but it may not be ready until tomorrow, hence the switch to "draft" on this PR.

The impediment here is that it acts differently on localhost, even with 2 workers, than in the CI with 2 workers.  So we're going to need to try something, then see what it really does in the CI in order to get this to a more stable place.  My definition of a stable place is: "doesn't require jobs retry."

jbdalton:
This issue was addressed by the merged PR [Dr-3796 Fix failing tests on CI](https://github.com/NYPL/digital-collections/pull/487).  Closing this Draft PR.

PR #487, opened 2025-11-14
Dr 3796 Fix failing tests on CI

jbdalton:
## Ticket:

This PR moves the brittle (when run on the CI) sort-filter tests into their own test file, s**earch-filters-sort.spec.ts**.  Although I've attached it to the DR-3796 ticket (where the sort-filters revisions originated), the purpose of this work is to stabilize our CI runs, which have been consistently failing this week only on the sort tests.

After attacking this from a timing angle first, I found that what was really going on was that the CI "viewport" is often pretty crummy in a headless environment, so Playwright _thinks_ it's not seeing visible states that are really there in the DOM.   And so a `.click()` event can fail because Playwright doesn't "see" the locator on the CI.

It turns out that using `click( {force: true} )` is sort of a magic bullet when applied in these scenarios.   After I added this to the sort filter tests, they immediately stopped failing under the `--workers=2` flag that the CI runs under.

I'm going to keep the sort-results isolated to their own file, because they do also do some heavy-lifting in terms of their conditional checks that rely on a fully loaded new results set each time.  But with the `force: true` in my toolkit now, these kinds of brittle CI scenarios are hopefully something we'll see less of.

### Checklist:

### DIFF
--- a/playwright/tests/search-filters-sort.spec.ts
+++ b/playwright/tests/search-filters-sort.spec.ts
@@ -1,0 +1,54 @@
+import { test, expect } from "../base";
+import SearchPage from "../pages/search.page";
+
+let searchPage: SearchPage;
+
+test.describe.serial("choose specific sort options", () => {
+  test.beforeEach(async ({ page }) => {
+    searchPage = new SearchPage(page);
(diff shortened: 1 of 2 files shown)


samanthaandrews (approved):
I like this!

PR #488, opened 2025-11-20
Dr 3900 - Item Fields: Identifiers Tests

jbdalton:
## Ticket:

Adds tests for identifiers under the "Identifiers" heading on the item page.

These include: UUID, RLIN/OCLC numbers, and catalog link for this PR.

Also included is a check for the Shelf Locator (call number) content and link, which doesn't appear under the "Identifiers" heading but still would be considered an "identifier" normally.

This also serves as the first section of an overall item Page Object build-out.   The other Field-headings/content should follow a similar pattern.  (eg. _item-fields-topic_, etc.)

## Open questions

The identifiers under the "Identifiers" heading are checked as a group.   And I'm grouping other "identifiers" that are displayed elsewhere (Shelf Locator, Image ID, etc.) under the test-wrapper "Other Identifiers."

The data being checked against comes from a known record. The identifier values are stored in the POM.  The goal here would be to check every type of identifier, so this pass uses Record A's identifier data and checks against it.  Record B will hold any other identifiers, and so on.   Since there is no one Record that has all possible identifiers in it, this is the approach we're planning to use for the various field-check specs.

Ideally, we will build out a random smoke test later on that might check a range of random records for _expected_ required fields, or to surface edge-cases that might be hidden for some optional fields.   But this current PR is for the happy path:  field is known, check that it is displaying correctly and in the correct place.

Ran new tests locally against a prod build, and ran with whole suite at CI-level constraints (workers=2).

### Checklist:

### DIFF
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -205,14 +205,6 @@ export default class ItemMetadataPage {
     await expect(catalogLinkLocator).toHaveText(
       ItemMetadataPage.EXPECTED_BNUMBER_VALUE
     );
-
-    //  // Get URL the link points to before clicking it
-    // const expectedUrl = (await catalogLinkLocator.getAttribute('href'))!;
-
-    // await catalogLinkLocator.click();
-    // // Check for correct catalog page
-    // const verificationText = "The New York Public Library"; // Example verification text
-    // await expect(this.page.locator('body')).toContainText(verificationText, { timeout: 15000 });
   }
 
   async verifyShelfLocatorIsPresent(): Promise<void> {
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -49,26 +49,6 @@ export default class ItemMetadataPage {
   readonly dataSourceHeading: Locator;
   readonly dataSourceLink: Locator;
 
-  // METADATA:
-  // Data Source link to catalog (optional)
-  // Item Data heading (req)
-
-  // Title (req)
-  // Collection (req)???
-  // Dates/Origin heading
-  // Library location’s heading and text (req)
-  // Library shelf locator (optional, under location heading)
-  // Topics and links
-  // Genre and link
-  // Notes and text
-  // Physical Description and text
-  // Description and text
-  // Type of Resource (req)?
-  // Identifiers and text (req - at minimum there will be an item uuid)
-  // Rights Statement heading and text (req)
-  // Name heading and text
-  // Place/Geographic heading and text
-
   constructor(page: Page) {
     this.page = page;
 
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -41,7 +41,7 @@ export default class ItemMetadataPage {
   static readonly EXPECTED_UUID_VALUE = "8b2b3160-c5d5-012f-d95c-58d385a7bc34";
   static readonly EXPECTED_OCLC_VALUE = "24501668";
   static readonly EXPECTED_BNUMBER_VALUE = "b14924644";
-  static readonly EXPECTED_SHELFLOCATOR_VALUE = "Map Div. 97-6199 [LHS 839]";
+  static readonly EXPECTED_SHELF_LOCATOR_VALUE = "Map Div. 97-6199 [LHS 839]";
   readonly rightsHeading: Locator;
   readonly rightsText: Locator;
   readonly dataSourceHeading: Locator;
@@ -94,8 +94,17 @@ export default class ItemMetadataPage {
     }); // Assuming full heading text
     this.libraryLink = this.libraryHeading.locator("+ p > a"); // Assumes link is inside the next paragraph
 
-    // Library shelf locator (optional)
-    this.shelfLocatorText = this.libraryHeading.getByText(/Shelf locator:/i);
+    // Library shelf locator
+    // this.shelfLocatorText = this.libraryHeading.getByText(/Shelf locator/i);
+
+    // Above locator doesn't work due to non-descendant, non-semantic tags used in front end.
+
+    this.shelfLocatorText = this.libraryHeading
+      .page()
+      .locator(
+        'div:has-text("Library location")' // Find the heading container
+      )
+      .getByText(/Shelf locator:/i);
 
     // Topics
     this.topicHeading = this.page.getByText("Topic", { exact: true });
@@ -209,11 +218,10 @@ export default class ItemMetadataPage {
       'span:has-text("RLIN/OCLC:")'
     );
 
-    if ((await oclcLocator.count()) > 0) {
-      await expect(oclcLocator).toHaveText(
-        `RLIN/OCLC: ${ItemMetadataPage.EXPECTED_OCLC_VALUE}`
-      );
-    }
+    await expect(oclcLocator).toBeVisible();
+    await expect(oclcLocator).toHaveText(
+      `RLIN/OCLC: ${ItemMetadataPage.EXPECTED_OCLC_VALUE}`
+    );
   }
 
   async verifyCatalogLinkIsPresent(): Promise<void> {
@@ -222,12 +230,11 @@ export default class ItemMetadataPage {
       .locator('span:has-text("NYPL Catalog ID")')
       .locator("a");
 
-    if ((await catalogLinkLocator.count()) > 0) {
-      await expect(catalogLinkLocator).toBeVisible();
-      await expect(catalogLinkLocator).toHaveText(
-        ItemMetadataPage.EXPECTED_BNUMBER_VALUE
-      );
-    }
+    await expect(catalogLinkLocator).toBeVisible();
+    await expect(catalogLinkLocator).toHaveText(
+      ItemMetadataPage.EXPECTED_BNUMBER_VALUE
+    );
+
     //  // Get URL the link points to before clicking it
     // const expectedUrl = (await catalogLinkLocator.getAttribute('href'))!;
 
@@ -238,6 +245,34 @@ export default class ItemMetadataPage {
     // // Check for correct catalog page
     // const verificationText = "The New York Public Library"; // Example verification text
     // await expect(this.page.locator('body')).toContainText(verificationText, { timeout: 15000 });
+  }
+
+  // async verifyShelfLocatorIsPresent(): Promise<void> {
+  //   const shelfLocator = this.shelfLocatorText.getByText(/Shelf locator:/i);
+
+  //   await expect(shelfLocator).toBeVisible();
+
+  //   await expect(shelfLocator).toHaveText(
+  //     ItemMetadataPage.EXPECTED_SHELF_LOCATOR_VALUE,
+  //   );
+  // }
+
+  async verifyShelfLocatorIsPresent(): Promise<void> {
+    // // --- DEBUG INSTRUMENTATION ---
+    // // Read the locator count and content *before* any assertions run.
+    // const locatorCount = await this.shelfLocatorText.count();
+    // const rawText = await this.shelfLocatorText.textContent();
+
+    // console.log(`[DEBUG] Shelf Locator Count: ${locatorCount}`);
+    // console.log(`[DEBUG] Raw Text Content: "${rawText}"`);
+    // // -----------------------------
+
+    await expect(this.shelfLocatorText).toBeVisible();
+
+    // confirms the text value includes the Call Number.
+    await expect(this.shelfLocatorText).toContainText(
+      ItemMetadataPage.EXPECTED_SHELF_LOCATOR_VALUE
+    );
   }
 
   async verifyRightsContent(): Promise<void> {
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -172,10 +172,9 @@ export default class ItemMetadataPage {
 
   async verifyUUIDIdentifierIsPresent(): Promise<void> {
     // 1. Locate the element using the scoped locator
-    const uuidLocator = this.identifiersText.locator(
-      'span:has-text("Universal Unique Identifier (UUID):")'
-    );
-
+    const uuidLocator = this.identifiersText
+      .locator("span")
+      .filter({ hasText: /Universal Unique Identifier \(UUID\):/ });
     // 2. Assert the element is visible and contains the exact required content
     await expect(uuidLocator).toBeVisible();
     await expect(uuidLocator).toHaveText(
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -18,6 +18,7 @@ export default class ItemMetadataPage {
   readonly datesLink: Locator;
   readonly libraryHeading: Locator;
   readonly libraryLink: Locator;
+  readonly shelfLocatorText: Locator;
   readonly topicHeading: Locator;
   readonly topicText: Locator;
   readonly nameHeading: Locator;
@@ -40,6 +41,7 @@ export default class ItemMetadataPage {
   static readonly EXPECTED_UUID_VALUE = "8b2b3160-c5d5-012f-d95c-58d385a7bc34";
   static readonly EXPECTED_OCLC_VALUE = "24501668";
   static readonly EXPECTED_BNUMBER_VALUE = "b14924644";
+  static readonly EXPECTED_SHELFLOCATOR_VALUE = "Map Div. 97-6199 [LHS 839]";
   readonly rightsHeading: Locator;
   readonly rightsText: Locator;
   readonly dataSourceHeading: Locator;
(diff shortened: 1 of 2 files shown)


playwright/tests/item-metadata.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+test.beforeEach(async ({ page }) => {
+  itemMetadataPage = new ItemMetadataPage(page);
+  await itemMetadataPage.loadPage(ItemMetadataPage.itemResultURL);
+});
+
+// --- Tests for Required Fields ---
+
+test.describe("Metadata Fields", () => {
+  test.skip('should display main "Item Data" header', async () => {
```
clarissarichard: Why are we skipping this? I see that `itemDataHeader` isn't fully defined in `item-metadata.page.ts`, is that why?

playwright/tests/item-metadata.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Verifies the main required heading is present.
+    // why is this  passing?  Where is the "Item data" text called from?
+    await expect(itemMetadataPage.itemDataHeader).toBeVisible();
+  });
+
+  test("should display Title heading and corresponding text", async () => {
+    // Checks the required Title heading and its text content.
```
clarissarichard: No need to include comments like this if the test name is descriptive enough!

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+    this.titleHeading = this.page.getByText("Title", { exact: true });
+
+    // Defines the TARGET (the content right after the header)
+    // This is the relative locator that enforces the structural relationship.
+    this.titleText = this.titleHeading.locator("+ p");
+
+    // Collection
+    this.collectionHeading = this.page.getByText("Collection", { exact: true });
+    this.collectionText = this.collectionHeading.locator("+ p");
+    this.collectionLink = this.collectionText.locator("a").first(); // Target the link within the following paragraph
```
clarissarichard: Could we replace `locator("a")` with `getByRole("link")` here?

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    // Library Location
+    this.libraryHeading = this.page.getByText("Library location", {
+      exact: true,
+    }); // Assuming full heading text
+    this.libraryLink = this.libraryHeading.locator("+ p > a"); // Assumes link is inside the next paragraph
+
+    // Library shelf locator
+    this.shelfLocatorText = this.libraryHeading
+      .page()
```
clarissarichard, line 83: What is this `page()` function for?

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+    this.identifiersText = this.identifiersHeading.locator("+ p");
+
+    // Rights Statement
+    this.rightsHeading = this.page.getByText("Rights Statement", {
+      exact: true,
+    });
+    this.rightsText = this.rightsHeading.locator("+ p");
+  }
+
+  async verifyTitleTextContent(): Promise<void> {
```
clarissarichard, line 140: Out of curiosity, why are we creating new functions when the assertions can exist directly in the test? Are we intending to reuse this in more than one test?

playwright/tests/item-metadata.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    expect(itemMetadataPage.titleHeading).toBeVisible;
```
clarissarichard: Let's add `await` to this.

playwright/tests/item-metadata.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    test("should include RLIN/OCLC identifier if present", async () => {
+      // This test checks OCLC field (which is conditionally checked inside the method).
+      await itemMetadataPage.verifyOclcIdentifierIsPresent();
+    });
+
+    test("should include the NYPL Catalog Link if present", async () => {
+      // This test checks the Catalog Link field.
+      await itemMetadataPage.verifyCatalogLinkIsPresent();
+    });
```
clarissarichard, line 34: I found an item that has more than one RLIN/OCLC and catalog ID: https://digitalcollections.nypl.org/items/6ebd9820-c5d3-012f-e2b2-58d385a7bc34?canvasIndex=0. I know we're navigating to a specific item page, but would these functions account for that scenario?

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+    );
+
+    await expect(oclcLocator).toBeVisible();
+    await expect(oclcLocator).toHaveText(
+      `RLIN/OCLC: ${ItemMetadataPage.EXPECTED_OCLC_VALUE}`
+    );
+  }
+
+  async verifyCatalogLinkIsPresent(): Promise<void> {
+    // Check existence first (Conditional Check)
```
clarissarichard, line 178: This doesn't appear conditional to me since it will fail if it is not visible.

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+    );
+
+    //  // Get URL the link points to before clicking it
+    // const expectedUrl = (await catalogLinkLocator.getAttribute('href'))!;
+
+    // await catalogLinkLocator.click();
+    // // Check for correct catalog page
+    // const verificationText = "The New York Public Library"; // Example verification text
+    // await expect(this.page.locator('body')).toContainText(verificationText, { timeout: 15000 });
+  }
```
clarissarichard: Let's remove commented code if it's not being used!

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+    await expect(this.titleText).toContainText(
+      "To His Excellency Sr. Henry Moore, Bart"
+    );
+  }
+
+  async verifyUUIDIdentifierIsPresent(): Promise<void> {
+    // 1. Locate the element using the scoped locator
+    const uuidLocator = this.identifiersText.locator(
+      'span:has-text("Universal Unique Identifier (UUID):")'
+    );
```
clarissarichard: Thoughts on using `getByText("Universal Unique Identifier (UUID):")` instead? It's better supported by Playwright than a CSS locator.

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+  async verifyTitleTextContent(): Promise<void> {
+    // Since 'this.titleText' was defined using the adjacent selector '+ p',
+    // this assertion inherently validates the structure AND the content.
+
+    // Structural check: Ensure the element exists and is visible
+    await expect(this.titleText).toBeVisible();
+
+    // Content check: Assert the specific expected text
+    await expect(this.titleText).toContainText(
+      "To His Excellency Sr. Henry Moore, Bart"
```
clarissarichard: This can probably be a constant too!

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  async verifyRightsContent(): Promise<void> {
+    // Structural check: Ensure the element exists and is visible
+    await expect(this.rightsHeading).toBeVisible();
+
+    // Content check: Assert the specific expected text
+    await expect(this.rightsText).toHaveText(
+      "SOME RIGHTS TEXT WILL GO HERE, AND IT VARIES"
+    );
+  }
```
clarissarichard, line 206: Did we want to add this to a test?

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  async verifyCatalogLinkIsPresent(): Promise<void> {
+    // Check existence first (Conditional Check)
+    const catalogLinkLocator = this.identifiersText
+      .locator('span:has-text("NYPL Catalog ID")')
+      .locator("a");
+
+    await expect(catalogLinkLocator).toBeVisible();
+    await expect(catalogLinkLocator).toHaveText(
+      ItemMetadataPage.EXPECTED_BNUMBER_VALUE
```
clarissarichard, line 185: We may not need to include `ItemMetadataPage` here! It should be able to refer to its definition above.

clarissarichard (commented):
I'd be interested to learn if we can add ARIA labels or test IDs to these elements at some point! These elements are super finnicky to locate in their current state and may be quick to break. Potential conversation to have with the devs!

playwright/tests/item-metadata.spec.ts
jbdalton: Yes, I was trying to keep the spec focused on the identifiers, but had started it as a "check all the required fields" (some of the identifiers are indeed required) test initially.  

The other non-identifier test, "check that title is there," is sort of a sanity-test.  But I think I should just remove this particular Item Data header test in the identifiers spec.  I'm not entirely sure whether all the fields will ultimately be compartmentalized out into specs, or will become part of one gigantic "fields" spec.

playwright/tests/item-metadata.spec.ts
jbdalton: Comments will be coming out before the PR is merged.  Just there as a roadmap currently.  That tends to be my process.

playwright/pages/item-metadata.page.ts
jbdalton: I think the `locator("a").first()` is there for checking the "top-level" (first) collection link.  But maybe `getByRole(Link)` can take a `.first()` attribute too?

playwright/pages/item-metadata.page.ts
jbdalton: hmmm.  that one does look like a typo 👀 .... except the tests did pass with it, so I'm going to dig further on this one.

playwright/pages/item-metadata.page.ts
jbdalton: Yes, I am planning for its reuse.  I'd like to be able to run these against a random checker, for example.  

However, I do see that I'm actually not using anything like the "verifyIdentiferIsPresent" format in the "identifiers" tests here in this PR.  Because we are testing against _static_ data in this particular test, it's not really relevant.

The "Verify" methods were ultimately intended for running against an item where we may not know which field is present or not.  But that's not what we're doing in the current spec.

playwright/pages/item-metadata.page.ts
jbdalton: Yes, this sounds good, as long as it doesn't scope to the entire page.  The goal is to make sure fields have the correct content, and that they appear in the"right spot"  (ie in this case under the "Identifiers" heading).

I'll switch it and see what it does.

playwright/pages/item-metadata.page.ts
jbdalton: Yes, we do.  That will be rolled out in the "rights" spec/section.

playwright/tests/item-metadata.spec.ts
jbdalton: Actually now that i have the file open, I think I'll leave the title-check in. But for now will continue to skip the Item heading check, since it's not fully defined yet.

playwright/pages/item-metadata.page.ts
jbdalton: Actually, this one was a doozie,  Since the markup is `h2` followed by additional SIBLING tags, the nested locator off of "libraryHeading" failed here, and necessitated the counter-intuitive use of the .page constraint.

By calling `.page()` and then using the subsequent `locator('div:has-text("Library location")')`, we're in effect telling Playwright: "Find the large container (the div) that encompasses the heading and all its siblings, and then look for the Shelf Locator text inside that box."

When the locator was constrained to the heading's subtree, since the heading is an `<h2>` its subtree is just the heading text itself.

playwright/pages/item-metadata.page.ts
jbdalton: correction:  I am using the "verify" style methods in the identifier fields too.  The title-check is using an assertion.  I would like to keep the "verify" methods in the the POM for now, for reuse in this case.  But normally it does look like our "assertions-in-the-spec" style is how I should be doing this in other specs.

playwright/pages/item-metadata.page.ts
jbdalton: Are you saying that the `toBeVisible` is all that's necessary, or to remove the `itemMetadataPage.` suffix?

In my IDE, it's complaining if those constants aren't scoped to `itemMetadataPage`.

playwright/pages/item-metadata.page.ts
jbdalton: Bad news: implementing the `identifiersText.getByText()` method for the UUID locator broke the existing test.

Since the application's non-semantic DOM structure places the label and the value in two separate elements, the simple `getByText()` is resolving only to the static label (`<b>Universal Unique Identifier (UUID):</b>`) and not capturing the sibling UUID value.

For now I've reverted to the stable CSS selector (`span:has-text(...)`). It's the only locator so far that has reliably anchored to the unique label element.

playwright/pages/item-metadata.page.ts
jbdalton: replaced `a` with `getByRole(link)`

playwright/tests/item-metadata.spec.ts
jbdalton: That's a good find!  I will check that scenario using a different constant.  This is also why I'm very interested in prototyping a truly random fuzz-test solution that will be much more likely to find these types of scenarios, and test them on the fly.

playwright/tests/item-metadata.spec.ts
jbdalton: comments removed.

playwright/pages/item-metadata.page.ts
jbdalton: removed.

playwright/pages/item-metadata.page.ts
jbdalton: added constant.

jbdalton:
@clarissarichard - I think this is ready for another pass?

playwright/pages/item-metadata.page.ts
clarissarichard: I was wondering why `ItemMetadataPage` was needed, but it must be how it is defined in the page object because I also see the IDE error. 🤷

playwright/tests/item-metadata.spec.ts
clarissarichard: I'm hesitant including incomplete tests like this, since skipped tests should only be temporary for known failures and not placeholders for unfinished work. I recommend either defining this header and unskipping the test or removing the test entirely.

jbdalton: Removed the placeholder (Item Data Heading) test.

playwright/pages/item-metadata.page.ts
clarissarichard: Figured out the Playwright equivalent of the CSS selector:
```
const uuidLocator = this.identifiersText.locator('span')
    .filter({ hasText: /Universal Unique Identifier \(UUID\):/ });

await expect(uuidLocator).toBeVisible();
await expect(uuidLocator).toContainText(ItemMetadataPage.EXPECTED_UUID_VALUE);
```

You could also consider finding both the label and UUID:
```
await expect(this.identifiersText).toContainText(
    `Universal Unique Identifier (UUID): ${ItemMetadataPage.EXPECTED_UUID_VALUE}`
  );
```

playwright/pages/item-metadata.page.ts
clarissarichard: What are your thoughts on just using `this.shelfLocatorText = this.page.getByText(/Shelf locator:/i);`? This is assuming it's the only one on the page, but that does appear to be the case!

playwright/pages/item-metadata.page.ts
```diff
@@ -0,0 +1,232 @@
...
+  // Topics and links
+  // Genre and link
+  // Notes and text
+  // Physical Description and text
+  // Description and text
+  // Type of Resource (req)?
+  // Identifiers and text (req - at minimum there will be an item uuid)
+  // Rights Statement heading and text (req)
+  // Name heading and text
+  // Place/Geographic heading and text
```
clarissarichard: Is this still needed?

playwright/pages/item-metadata.page.ts
jbdalton: Language like "shelf locator" could definitely appear elsewhere, like in the notes field.   Even though for _this_ record it doesn't, I would like to avoid page-find-anywhere type locators for these field names if it is possible so that we can re-use the POM for more dynamic checks (of random items, etc.).

playwright/pages/item-metadata.page.ts
jbdalton: changed to `.filter({hasText: <uuid-text>})` style

playwright/pages/item-metadata.page.ts
jbdalton: removed WIP field guide for next-steps.  I'll keep these elsewhere.

jbdalton:
@clarissarichard  - Ready for another pass, I hope?   Pushed latest commits to address the accessibility locator and remove WIP comments.

clarissarichard (approved):
Looks good to go!

PR #491, opened 2025-12-09
Dr 3900 - Item fields: names

jbdalton:
## Ticket:

Added tests to check Name and Name-Role values on the item page.  These tests also verify that Name is linked and Role is plain-text. 

## Open questions

 I started with a simple "NAME_ONE_VALUE," "NAME_TWO_ROLE," type of assertion.  But I refactored this to take a list of name-role combinations instead.  Also, it is a known feature of Solr that values in multi-valued fields can sort randomly, so I did not want the test to have a brittle dependence on _which_ name appeared first in the UX field list. 
 
 Because they aren't semantically marked up, we're relying on spans to break out the name-value nodes.  Tried this first with a "scope >" attribute in Playwright, but was foiled there by some `<br>` tags hanging out below the spans.  Span seems to do the trick here, and can also be utilized for Topic, or other multiValued field-checks.

I've left some comments in-place to help in reviewing this more complex test interaction, but these will be removed before the final-merge.

Ran test suite on localhost against a prod build with the `--workers=2` flag.

### Checklist:

### DIFF
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -52,6 +52,7 @@ export default class ItemMetadataPage {
   static readonly EXPECTED_COLLECTION_ROOT_VALUE =
     "Lawrence H. Slaughter Collection of English maps, charts, globes, books and atlases";
   static readonly EXPECTED_COLLECTION_LEVEL_ONE_VALUE = "Charts and maps";
+  static readonly EXPECTED_NAME_COUNT = 2;
   static readonly EXPECTED_NAME_ONE_VALUE = "Ratzer, Bernard";
   static readonly EXPECTED_NAME_ONE_ROLE_VALUE = "(Cartographer)";
   static readonly EXPECTED_NAME_TWO_ROLE_VALUE = "Kitchin, Thomas, 1718-1784";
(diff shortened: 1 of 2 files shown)


playwright/pages/item-metadata.page.ts
```diff
@@ -225,6 +262,47 @@ export default class ItemMetadataPage {
...
+      const { locator: nameContainer, pattern: fullExpectedPattern } =
+        this.locatorService.getNameLocatorAndPattern(
+          allNameEntryContainers,
+          expectedName
+        );
+
+      // Verify the full text (including the non-clickable role) is present.
+      await expect(nameContainer).toHaveText(fullExpectedPattern);
+    }
+  }
```
clarissarichard, line 298: Thoughts on combining these two functions? You may need to then combine the tests, but I think it would still make sense to have one test to check the Names section.

playwright/pages/item-metadata.page.ts
```diff
@@ -1,7 +1,34 @@
...
+
+    // Filter the general locator to find the specific container
+    const locator = allEntryContainers.filter({
+      hasText: fullExpectedPattern,
+    });
+
+    // return locator values and full-text for assertions
+    return { locator: locator, pattern: fullExpectedPattern };
+  }
+}
```
clarissarichard, line 27: Are you planning to use a function like this for the other metadata fields? Is it at all possible to make it more general for plugging in other metadata fields?

playwright/pages/item-metadata.page.ts
```diff
@@ -48,10 +79,16 @@ export default class ItemMetadataPage {
   static readonly EXPECTED_COLLECTION_ROOT_VALUE =
     "Lawrence H. Slaughter Collection of English maps, charts, globes, books and atlases";
   static readonly EXPECTED_COLLECTION_LEVEL_ONE_VALUE = "Charts and maps";
-  readonly rightsHeading: Locator;
-  readonly rightsText: Locator;
-  readonly dataSourceHeading: Locator;
-  readonly dataSourceLink: Locator;
+  static readonly EXPECTED_NAMES = [
+    { name: "Ratzer, Bernard", role: "Cartographer" },
+    { name: "Kitchin, Thomas, 1718-1784", role: "Engraver" },
```
clarissarichard, line 84: Do the names always have a role?

playwright/pages/item-metadata.page.ts
jbdalton: I want them to give specific error messages, so each of those `verifyBlaBlaBla` functions is called by a different, specific spec describe.block.  This way we have more granular traceability for failures in the test-results report.

playwright/pages/item-metadata.page.ts
jbdalton: Yes it's going to be used for Topics/Subjects next (that code, for `getTopicsLocator`, will be in the next PR), and  that method will check for slightly different patterns.  And for some other, simpler fields, a `getBasicLocator`, which is also coming in that next PR, should suffice.

playwright/pages/item-metadata.page.ts
jbdalton: Yes, I believe they do, but I will verify this with a `solr.facets` query on all names.   And  if there are some names that don't have a role, I'll need to add an item UUID to the test-data for an example of that condition.

jbdalton: In doing a facet analysis of the API's solr index, I've found that any "name-ish" term without a corresponding role (for example, corporate names like "New York Central Railroad Company, " or "New York Shakespeare Festival Public Theater") gets filed under Subject on the item fields.   Based on that, I'm pretty confident that we shouldn't have any names appearing under the "Names" heading without a role.

jbdalton:
@clarissarichard I've removed the deprecated extra name-N-value and name-count vars, and I've double-checked on the "Roles" question (see comments in reply).  I think this should be ready for another pass?

clarissarichard (approved):

PR #497, opened 2026-01-08
grid view

avertrees:
## Tickets:
- [GRID: Implement View Toggle with Grid and List Icons](https://newyorkpubliclibrary.atlassian.net/browse/DR-3977)
- [GRID: Grid View Card](https://newyorkpubliclibrary.atlassian.net/browse/DR-3978)
- [GRID: Apply Thumbnail Scaling and Aspect Ratio in Grid View](https://newyorkpubliclibrary.atlassian.net/browse/DR-3979)
- [GRID: Implement Grid View Layout and Responsiveness Logic](https://newyorkpubliclibrary.atlassian.net/browse/DR-3980)

- Creates a viewingOptionsMenu component to render the grid and list view buttons
- updates query parameter based on button selected
- loads corresponding card type

VQA - Round 1 & 2 Resolved Issues
- Default View: All pages now default to grid view when first landing on the page.
- Toggle Active State/Design: The view toggle now shows an active state (blue background + white icon), the toggle icons are enclosed in a container, the corners are rounded to match the "sort by" filter, and the button outline for the inactive view has been addressed.
- Toggle Stacking: Toggle buttons stacking at mobile viewports has been resolved.
- Status Badge Wrapping: Blue status badges wrap once they are wider than the card.
- Grid View Description: The description is now hidden in grid view.
- Search Results Responsiveness: The grid column display for search results is now responsive, going from 4 columns (desktop) to 2 columns (tablet) to 1 column (mobile).
- Collection w/ Structure Padding: The extra 24px left/right padding has been removed.
- Collection w/o Structure Padding: The extra 24px left/right padding has been removed.
- Highlighter Color Contrast: The highlighter color's contrast issue has been confirmed as fixed.
 
A11y Resolved Issues
- Grid/List Button Labels: The programmatic labels for the grid view and list view buttons were updated to remove the unnecessary word "icon."
- Grid/List Button State: The grid view and list view buttons now programmatically convey their state using aria-pressed (true/false).
- Sort By Button Label: The aria-label for the "Sort By" button has been fixed to remove unnecessary repetition of the selected value.

Locally and in vercel. 

- Focus management is not working, but it's also not working in Prod. I created this ticket to fix it https://newyorkpubliclibrary.atlassian.net/browse/DR-4030 and it's a high priority item post launch. 

### Checklist:


app/src/utils/searchManager/searchManager.tsx
```diff
@@ -180,25 +205,34 @@ export class GeneralSearchManager extends BaseSearchManager {
       sort: sort,
       page: this.currentPage,
       filters: filterToString(this.filters),
+      viewMode: this.currentViewMode,
     });
   }
 
   getQueryString = (paramsObject: Record<string, any>) => {
```
7emansell, line 212: ah i think there's like one line when the non-default values aren't getting assigned back to `newParams` (all filters and everything aren't working rn too bc of it).  
```
getQueryString = (paramsObject: Record<string, any>) => {
  const newParams: Record<string, string> = {};

  Object.entries(paramsObject).forEach(([key, value]) => {
    if (value == null) return; 

    let isDefault = false;

    switch (key) {
      case "q":
        isDefault = value === DEFAULT_SEARCH_TERM;
        break;
      case "page":
        isDefault = value === DEFAULT_PAGE_NUM;
        break;
      case "sort":
        isDefault = value === this.defaultSort;
        break;
      case "filters":
        isDefault = value === "" 
        break;
      case "viewMode":
        isDefault = value === DEFAULT_VIEW_MODE;
        break;
    }

    if (!isDefault) {
      newParams[key] = value;
    }
  });
```

7emansell (commented):
works perfect with the searchManager getQueryString fix!! Dk about other implementation details but looks good too

app/src/components/grids/searchCardsGrid.tsx
```diff
@@ -8,13 +8,18 @@ import { SearchResultType } from "@/src/types/SearchResultsType";
...
+  viewMode,
+  type,
 }: {
   results: SearchResultType[];
   keywords: string;
+  viewMode: "grid" | "list";
+  type: "collection" | "search";
 }) => {
   const { isLargerThanLargeTablet } = useBreakpoints();
+  const numColumns = type == "collection" ? 3 : 4;
```
avertrees: @7emansell I thought this would work but turns out the collections page doesn't always have the collection structure component and in that case... I imagine a four card grid would be better. is this something for the search manager to handle or is there a better way to handle this case that I am missing?

app/src/components/grids/searchCardsGrid.tsx
```diff
@@ -8,13 +8,17 @@ import { SearchResultType } from "@/src/types/SearchResultsType";
...
 }: {
   results: SearchResultType[];
   keywords: string;
+  viewMode: "grid" | "list";
+  numColumns: number;
 }) => {
   const { isLargerThanLargeTablet } = useBreakpoints();
   return (
-    <SimpleGrid columns={1} gap="grid.l">
+    <SimpleGrid columns={viewMode === "list" ? 1 : numColumns} gap="grid.l">
```
sarangj: small thing but it *could* be a little confusing for `numColumns` to not be respected when we're doing list view.  Would it make sense to push this logic up to the collection page and search page, and pass 1 for `numColumns` if we're in list view?

sarangj (commented):
I know this is in VQA, but just thought I'd give a look as well, one small question / suggestion

sarangj (approved):

PR #506, opened 2026-01-14
Dr 3900- item fields physical description and notes

jbdalton:
## Ticket:

Adds Physical Description and Notes fields tests to POM and Spec files.

Most of this is based on the same framework that name, genre, etc. use:  it allows checking multi-valued fields, and uses a `scenario()` method to handle different sample records instead of hardcoding the UUIDs into the field var names.  This also allows one spec (item-metadata.spec) to easily access the different sample items that are required for various fields.

## Open questions

There are many, many different types of Notes, but the display of them  ("`<type>: <note-text>`") is the same: one note per new line-break on the front-end.   At a later date, we should return to this if we want to verify that all of the various "note-types" display their correct "type" and values.  This would also be checked in the API upstream, and since the note is handed off to the front-end intact, there might not be be any parsing/decision-trees in the UX for various note-types.  If the entire Notes "blob" is just being passed to the UX, then checking the various types, imo, would be overkill for the front-end playwright tests.

The Physical Description test/check is much more straightforward.  And since the sample record is showing only one, we are just checking for that value.

All tests were run locally against the production build.

`npm run build`
`npx playwright test --workers=2`

### Checklist:

### DIFF
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -218,6 +218,10 @@ export default class ItemMetadataPage {
     this.typeHeading = this.page.getByText("Type of Resource", { exact: true });
     this.typeText = this.typeHeading.locator("+ p");
 
+    // Languages
+    this.languageHeading = this.page.getByText("Languages", { exact: true });
+    this.languageText = this.languageHeading.locator("+ p");
+
     // Rights Statement
     this.rightsHeading = this.page.getByText("Rights Statement", {
       exact: true,
--- a/playwright/pages/item-metadata.page.ts
+++ b/playwright/pages/item-metadata.page.ts
@@ -327,17 +327,14 @@ export default class ItemMetadataPage {
     }
   }
 
-  async verifyNotesCount(): Promise<void> {
-    const rawText = await this.notesText.innerText();
-    // Playwright's innerText() converts <br> tags to newlines
-    const actualNotes = rawText
-      .split("\n")
-      .filter((line) => line.trim().length > 0);
-
-    expect(actualNotes.length).toEqual(ItemMetadataPage.EXPECTED_NOTES.length);
+  async verifyPhysicalDescriptionContent(): Promise<void> {
+    await expect(this.physicalHeading).toBeVisible();
+    await expect(this.physicalText).toContainText(
+      ItemMetadataPage.EXPECTED_PHYSICAL_DESCRIPTION_VALUE
+    );
   }
 
-  async verifyNotesValues(): Promise<void> {
+  async verifyNotesText(): Promise<void> {
     const rawText = await this.notesText.innerText();
 
     // Normalize whitespace and split by newline to verify each distinct note
@@ -347,6 +344,16 @@ export default class ItemMetadataPage {
       .filter((val) => val.length > 0);
 
     expect(actualNotes).toEqual(ItemMetadataPage.EXPECTED_NOTES);
+  }
+
+  async verifyNotesCount(): Promise<void> {
+    const rawText = await this.notesText.innerText();
+    // Playwright's innerText() converts <br> tags to newlines
+    const actualNotes = rawText
+      .split("\n")
+      .filter((line) => line.trim().length > 0);
+
+    expect(actualNotes.length).toEqual(ItemMetadataPage.EXPECTED_NOTES.length);
   }
 
   async verifyRightsContent(): Promise<void> {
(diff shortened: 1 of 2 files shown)


playwright/pages/item-metadata.page.ts
```diff
@@ -1,5 +1,8 @@
 import { Locator, Page, expect } from "@playwright/test";
 
+// sets which UUID is to be used for sample record navigation
+export type MetadataScenario = "DEFAULT" | "SAMPLE1" | "SAMPLE2";
```
sarangj, line 4: nice! Would it make sense to rename 'SAMPLE1' and 'SAMPLE2' to 'TOPICS' and 'GENRES', since those are commented below?

playwright/pages/item-metadata.page.ts
```diff
@@ -297,6 +327,35 @@ export default class ItemMetadataPage {
...
+
+  async verifyNotesCount(): Promise<void> {
+    const rawText = await this.notesText.innerText();
+    // Playwright's innerText() converts <br> tags to newlines
+    const actualNotes = rawText
+      .split("\n")
+      .filter((line) => line.trim().length > 0);
+
+    expect(actualNotes.length).toEqual(ItemMetadataPage.EXPECTED_NOTES.length);
+  }
```
sarangj, line 382: hmm, do we need this? It seems like the assertion in the `verifyNotesText()` method will also only pass if the lengths are equal.

playwright/tests/item-metadata.spec.ts
```diff
@@ -49,29 +52,58 @@ test.describe("Verify Metadata Fields", () => {
...
 
-  test("should display correct name and role values", async () => {
-    await itemMetadataPage.verifyNameDataValues();
+    test("should include correct notes values", async () => {
+      await itemMetadataPage.verifyNotesText();
+    });
+
+    test("should contain the correct number of notes", async () => {
+      await itemMetadataPage.verifyNotesCount();
+    });
```
sarangj, line 130: same question, re: the notes text test seems to also implicitly test the count

sarangj (commented):
nice! Just a couple small questions / suggestions

playwright/pages/item-metadata.page.ts
jbdalton: I've actually (in another PR) got multiple field-checks happening in some of the "SAMPLE1" etc. records.  It's not ideal, in terms of how the Spec `describe()` reads "Check Sample 2" > Notes > should contain expected tex"t, etc.  But I'm trying to limit the number of "sample" records" overall needed for this item-metadata.spec.  I might end up making the wrapper `describe()` empty or blank to Hide that "Check Sample 2" part.   I think I'll wind up needing only 3-5 "sample" records max when it's all done, since got only 2 or 3 fields remaining to check.

sarangj: got it, that makes sense!

playwright/tests/item-metadata.spec.ts
jbdalton: yes, the "text" test does implicitly test the count.  I'm trying to also call it out as its own assertion in the Spec language in a "Fail First" way.  So that if the count is off for some reason (like an extra note suddenly shows up), the failure shows up in the `describe("should have accurate count")` spec line.  If the count-failure relies only on the "Text" test not passing, it'll be harder imo to debug from the Spec feedback itself since there'll be a much longer failure diff in a consolidated assertion that tests both at the same time.

I was thinking of moving that normalization logic to its own helper, though, which should remove a lot of the duplication here at least:

Example:

```
// Handles <br> tags (via innerText) and normalizes whitespace.
private async getNormalizedLines(locator: Locator): Promise<string[]> {
  const rawText = await locator.innerText();
  return rawText
    .split("\n")
    .map((val) => val.trim().replace(/\s+/g, " "))
    .filter((val) => val.length > 0);
}

// Count assertion :
async verifyNotesCount(): Promise<void> {
  const actual = await this.getNormalizedLines(this.notesText);
  expect(actual.length, "Notes count mismatch").toEqual(ItemMetadataPage.EXPECTED_NOTES.length);
}

// Values assertion
async verifyNotesText(): Promise<void> {
  const actual = await this.getNormalizedLines(this.notesText);
  expect(actual).toEqual(ItemMetadataPage.EXPECTED_NOTES);
}

```

Basically I'm trying to cover both of these failing scenarios in the spec:

1. Note count is correct, but one of the notes is "wrong"
Result: Text-test fails, but Count-test passes.

2. The existing notes values are correct, but there is an "extra" notes line (either a duplicate, or different, note).
Result: Text-test and Count-test both fail.

sarangj: ah okay, so it's a matter of visibility to the developer rather than behavior, makes sense, and thank you for the detailed explanation! Yea, the helper method might make sense, but could be in a small follow up PR!

sarangj (approved):

PR #514, opened 2026-01-26
Implement grid view analytics

sarangj:
## Ticket:
- JIRA ticket [DR-4009](https://newyorkpubliclibrary.atlassian.net/browse/DR-4009)
- JIRA ticket [DR-4010](https://newyorkpubliclibrary.atlassian.net/browse/DR-4010)

Note, this is branched off https://github.com/NYPL/digital-collections/pull/497, I'll wait for that to go in before publishing this PR.

Adds two events for grid view to ga4:

  1. A simple event when a user explicitly selects a search view option (ie, clicking list or grid) 
  2. General search analytics (including the view type) on searches

## Open questions

For the search analytics, I implemented this by standardizing the search analytics data in the search manager object, and then using that data as dependencies for an effect (so that if the search params change, a new event is sent).

  1. Is the search manager a good place for that method? Seemed like a simple spot to put it, and I kept the search manager decoupled from the ga4 module
  2. Should the different `useEffect` calls be refactored into their own hook?
  3. Is there a better way to add the dependency than JSON encoding the analytics data? I found that if I just passed the data in directly, because how js object equality works, we'd end up firing duplicate events. If I passed each field separately as a dep, the array `filterNames` would still suffer the same issue. 
  
Tested on vercel by inspecting `window.dataLayer` in the console.

I'll add a couple unit tests as well.

### Checklist:

[DR-4009]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DR-4010]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/hooks/useSearchAnalytics.ts
+++ b/app/src/hooks/useSearchAnalytics.ts
@@ -1,0 +1,29 @@
+import { useEffect } from "react";
+import {
+  GeneralSearchManager,
+  SearchManager,
+} from "../utils/searchManager/searchManager";
+import { trackSearchResults } from "../utils/ga4Utils";
+
+const useSearchAnalytics = (searchManager: SearchManager) => {
+  const filters = searchManager.filters;
+  const filterNames = filters.map((f) => f.filter);
+  const data = {
+    searchTerm: searchManager.keywords ?? undefined,
+    filterNames: searchManager.filters.map((f) => f.filter),
+    searchResultsLayout: searchManager.viewMode,
+    searchType:
+      searchManager.filters.find((f) => f.filter === "rights")?.value ??
+      undefined,
+  };
+  useEffect(() => {
+    trackSearchResults(
+      data.searchResultsLayout,
+      data.filterNames,
+      data.searchType,
+      data.searchTerm
+    );
+  }, [JSON.stringify(data)]);
+};
+
+export default useSearchAnalytics;
--- a/app/src/hooks/useSearchAnalytics.ts
+++ b/app/src/hooks/useSearchAnalytics.ts
@@ -7,23 +7,26 @@ import { trackSearchResults } from "../utils/ga4Utils";
 
 const useSearchAnalytics = (searchManager: SearchManager) => {
   const filters = searchManager.filters;
-  const filterNames = filters.map((f) => f.filter);
-  const data = {
-    searchTerm: searchManager.keywords ?? undefined,
-    filterNames: searchManager.filters.map((f) => f.filter),
-    searchResultsLayout: searchManager.viewMode,
-    searchType:
-      searchManager.filters.find((f) => f.filter === "rights")?.value ??
-      undefined,
-  };
+  const filterNames = filters.map((f) => f.filter) ?? undefined;
+  const keywords = searchManager.keywords ?? undefined;
+  const searchType =
+    searchManager.filters.find((f) => f.filter === "rights")?.value ??
+    undefined;
+  // We don't send an event if the viewMode is the only thing that changed
+  const changeHash = JSON.stringify({
+    filterNames: filterNames,
+    searchType: searchType,
+    keywords: keywords,
+  });
+
   useEffect(() => {
     trackSearchResults(
-      data.searchResultsLayout,
-      data.filterNames,
-      data.searchType,
-      data.searchTerm
+      searchManager.viewMode,
+      filterNames,
+      searchType,
+      keywords
     );
-  }, [JSON.stringify(data)]);
+  }, [changeHash]);
 };
 
 export default useSearchAnalytics;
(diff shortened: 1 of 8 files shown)


app/src/utils/ga4Utils.ts
```diff
@@ -36,6 +36,14 @@ export const sendDownloadEvent = (
   });
 };
 
+export const sendLayoutSelectedEvent = (viewMode: "Grid" | "List") => {
```
sarangj: i should make these params lowercase like in the other ga4 event below

app/src/utils/searchManager/searchManager.tsx
```diff
@@ -102,6 +110,20 @@ abstract class BaseSearchManager implements SearchManager {
     return this.currentAvailableFilters;
   }
 
+  get analyticsData(): AnalyticsData {
+    const filters = this.filters;
+    const filterNames = filters.map((f) => f.filter);
+    const searchType =
+      filters.find((f) => f.filter === "rights")?.value ?? "default";
```
sarangj: this default behavior should probably go elsewhere

app/src/utils/searchManager/searchManager.tsx
```diff
@@ -102,6 +110,21 @@ abstract class BaseSearchManager implements SearchManager {
     return this.currentAvailableFilters;
   }
 
+  get analyticsData(): AnalyticsData {
```
avertrees: My only thought here is that this extends the scope of the SearchManager to handling logic beyond what supports Search or Search based pages. We might want to rename this? I don't know what the best practices are for handling these kinds of conundrums.

avertrees:
mmm I'm not seeing the data in `window.dataLayer` change when I change click the `List` button

app/src/utils/searchManager/searchManager.tsx
sarangj: yea, that's a very good point! I think maybe the right move here is to move this all into a hook that takes in a searchManager and just extract the data out of the searchManager there. Or even just a small helper that takes a searchmanager and returns analytics data, to keep the analytics stuff decoupled

avertrees:
> mmm I'm not seeing the data in `window.dataLayer` change when I change click the `List` button

nvm I was looking at the wrong location

app/src/utils/searchManager/searchManager.tsx
avertrees: sounds right to me. @EdwinGuzman @7emansell any thoughts? I know you guys have more experience/thoughts here.

app/src/hooks/useSearchAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+  const filters = searchManager.filters;
+  const filterNames = filters.map((f) => f.filter);
+  const data = {
+    searchTerm: searchManager.keywords ?? undefined,
+    filterNames: searchManager.filters.map((f) => f.filter),
+    searchResultsLayout: searchManager.viewMode,
+    searchType:
+      searchManager.filters.find((f) => f.filter === "rights")?.value ??
+      undefined,
+  };
```
sarangj: this is a tad screwy now, ill fix

app/src/utils/searchManager/searchManager.tsx
EdwinGuzman: After reviewing the PR, I agree that it makes sense to break this up so the analytics functionality isn't coupled into the `SearchManager`.

EdwinGuzman:
Reviewed this with @isastettler because she's also working on GA4 event tracking in the Scout app.

> Is the search manager a good place for that method? Seemed like a simple spot to put it, and I kept the search manager decoupled from the ga4 module
> Should the different useEffect calls be refactored into their own hook?

Looks like this was already made and it makes sense.

> Is there a better way to add the dependency than JSON encoding the analytics data? I found that if I just passed the data in directly, because how js object equality works, we'd end up firing duplicate events. If I passed each field separately as a dep, the array filterNames would still suffer the same issue.

This may not be the most ideal way to track object updates, but it works well and is a common pattern. It's doing string comparison over the shallow object diff. I'm surprised that `filterNames` had the same issue (as individual values in the dep array) since that's an array of strings based on a _new_ instance.

We verified the data is set in `window.dataLayer`. Just don't want to approve on a repo we don't work on, but code looks good!

7emansell (commented):
Agree that the `useSearchAnalytics` makes more sense as a hook and shouldn't be on the `searchManager`, and seeing expected event data in the console 👍

keithbauer (approved):

PR #516, opened 2026-01-29
DR-4008 Cache user's viewing preference in session storage

keithbauer:
## Ticket:
[DR-4008](https://newyorkpubliclibrary.atlassian.net/browse/DR-4008)

Saves the current viewMode to localStorage and tries to retrieve the value from localStorage.

## Open questions

Is this the correct place for this logic? It seems to work fine when I click around.

Tested this locally and can see the stored value in localStorage.
I also manually set the value to something other than `grid` or `list` and the grid view displays by default.

### Checklist:

[DR-4008]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/context/SearchProvider.tsx
+++ b/app/src/context/SearchProvider.tsx
@@ -34,13 +34,16 @@ export const SearchProvider = ({
   children: React.ReactNode;
 }) => {
   const urlSearchParams = useSearchParams();
-  const viewModeFromUrl = urlSearchParams.get("viewMode") as
-    | "grid"
-    | "list"
-    | null;
+  const viewModeFromUrl = urlSearchParams.get("viewMode");
+  const validViewMode = (
+    mode: string | null | undefined
+  ): mode is "grid" | "list" => mode === "grid" || mode === "list";
 
-  const initialViewMode =
-    viewModeFromUrl || searchParams?.viewMode || DEFAULT_VIEW_MODE;
+  const initialViewMode = validViewMode(viewModeFromUrl)
+    ? viewModeFromUrl
+    : validViewMode(searchParams?.viewMode)
+    ? searchParams.viewMode
+    : undefined;
 
   const lastFilterRef = useRef<string | null>(null);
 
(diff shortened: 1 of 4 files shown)


sarangj:
Can't seem to actually review this until the upstream is merged to qa, but lgtm!

sarangj (approved):
okay, figured out how to approve this lol, but let's wait til @avertrees 's work is in `qa` and then merge this into qa after!

avertrees:
If I select List view then open https://digital-collections-git-finalgridalytics-nypl.vercel.app/search/index in a new tab, it successfully loads in List view. However, if I then click the Grid view button, nothing changes. I think we need to update the default view to also use session storage.

app/src/context/SearchProvider.tsx
```diff
@@ -27,10 +34,16 @@ export const SearchProvider = ({
...
 }) => {
   const urlSearchParams = useSearchParams();
-  const viewModeFromUrl = urlSearchParams.get("viewMode") as
-    | "grid"
-    | "list"
-    | null;
+  const viewModeFromUrl = urlSearchParams.get("viewMode");
+  const validViewMode = (
+    mode: string | null | undefined
+  ): mode is "grid" | "list" => mode === "grid" || mode === "list";
```
sarangj: cool use of a typeguard!  Having the type guard on the same line as the function body makes it a bit difficult to read imo, could we do
```typescript
  const validViewMode = ( mode: string | null | undefined ): mode is "grid" | "list" => {
    return mode === "grid" || mode === "list";
  }
```

app/src/context/SearchProvider.tsx
```diff
@@ -27,10 +34,16 @@ export const SearchProvider = ({
...
+
+  const initialViewMode = validViewMode(viewModeFromUrl)
+    ? viewModeFromUrl
+    : validViewMode(searchParams?.viewMode)
+    ? searchParams.viewMode
+    : undefined;
```
sarangj: also, I think this might be a case where just if / else would be simpler (I find nested ternaries a bit hard to sort through).

sarangj (commented):
couple style suggestions, but it looks like it's working now!  May be some opportunity to clean up / consolidate, but let's get this out!

sarangj (approved):

PR #519, opened 2026-02-06
DR - 3900 - Image downloads and purchase

jbdalton:
## Ticket:

Adds playwright tests to verify that an approved (in this test case it's from Public Domain) image can be purchased, that the links go to our external Prints vendor, and that image-downloads are accessible from the "downloads" icon/button.

locally:
`npm run build`
`npx playwright test --workers=2`

ran all tests on a local build of prod

### Checklist:


playwright/pages/item-media.page.ts
```diff
@@ -37,6 +44,24 @@ export default class ItemMediaPage {
...
+    });
+    this.closeOverlayButton = this.downloadOverlay.getByRole("button", {
+      name: /Close/i,
+    });
+
+    // Share options
+    this.shareButton = page.getByRole("button", { name: "Share" });
+
+    // Purchase option
+    this.orderPrintButton = page.locator("#print-btn");
```
sarangj: Would it make sense to also make this a role locator like you have with the other buttons? Looking at the html on the page, I think something like 
```js
this.orderPrintButton = page.getByRole("button", { name: /Order Print/i });
```
should work

playwright/pages/item-media.page.ts
```diff
@@ -37,6 +44,24 @@ export default class ItemMediaPage {
...
     this.playButton = page.locator("button").filter({ hasText: /^Play$/ });
+
+    // Download options
+
+    this.downloadButton = page.getByRole("button", { name: /Download/i });
+    this.downloadOverlay = page.locator("div.overlay.download");
+    this.downloadOptionSmall = page.locator("button", { hasText: /Small/i });
+    this.downloadOptionStandard = page.locator("button", {
+      hasText: /Standard/,
+    });
```
sarangj: would it make sense to get these from the `downloadOverlay` explicitly, like we do with the closeOverlayButton below? So that we are more explicitly testing that these buttons are contained in the overlay?

jbdalton: adding this, and i'll push it shortly.  The tests ran fine with the scope.  I think last week I had initially tried it scoped and seemed to run into some kind of DOM-contention issues, so went with the global locators at that time.  But since this change is running fine today, and is more future-proof in terms of the scope, I'm all for it.

playwright/pages/item-media.page.ts
jbdalton: updated.  Needed to use `getByRole("link", { name: /Order Print/i });`  Because what 'looks like' a button, and surely feels like a button for the user, is actually a link.

jbdalton:
I think this is ready for re-review?  The tests had to be re-run in the CI for a few mysterious 502s on the qa-API (?).  But all seems to be stable now.

sarangj (approved):

PR #526, opened 2026-02-24
Make selecting already selected layout no-op 

sarangj:
## Ticket:

NOREF

If clicking the already selected view mode, just make the button no-op,
as that's probably what a user would expect anyway, and we don't want to
send dummy ga4 events or reload the search results.

### Checklist:


keithbauer (approved):
nice! looks great in the preview

PR #528, opened 2026-02-26
DRAFT: NO_REF / Add slow-flag to media-asset check that consistently has to retry in CI

jbdalton:
## Ticket:

- JIRA ticket [DR-xxxx](url)

Noticed one of the test-blocks in the check-media spec consistently has to re-run on CI, or even fails outright the first time, then works on a hard, manual kick.

Adding playwright's `test(slow);` flag to the offending block, to see if this increases the chance of the retry being successful when running the whole suite.

Testing will mainly be checked in the CI itself, since various flags can be added locally to the cmd-line etc to try out various retry strategies for slow or flakey test scenarios.  

This approach, `test(slow);`, is at least file-specific, and doesn't mess with existing global CI settings.  It's worked as expected on my localhost when running the whole suite at the `--workers=2` setting.  Which is to say it pauses the offending block long enough, it seems, that the inevitable retry has been successful on localhost.

### Checklist:

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -12,6 +12,8 @@ import { defineConfig, devices } from "@playwright/test";
  * See https://playwright.dev/docs/test-configuration.
  */
 
+process.env.IIIF_URL = process.env.IIIF_URL || "https://iiif.nypl.org";
+
 export default defineConfig({
   testDir: "./playwright",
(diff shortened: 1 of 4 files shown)


jbdalton:
Polling and a vastly sped up overall IIIF infrastructure has made this suggested improvement moot for now. Closing.

PR #533, opened 2026-03-04
Set server side IIIF_URL - NOREF

sarangj:
It looks like for next/image to use the correct value (instead of the fallback) on the server, we need to set this.

## Ticket:

NOREF 

### Checklist:


keithbauer (approved):
Definitely hitting qa-iif now!

PR #534, opened 2026-03-10
NOREF - Remove unused API calls / repoAPI checks

sarangj:
## Ticket:

NOREF

Was digging into another thing and realized we have some dead code. This gets rid of the `isRepoApi` flag and also removes the unused `getItemManifest` function, since we only get manifests from UV now.
### Checklist:


keithbauer (approved):
nicee

emu47 (approved):

PR #544, opened 2026-03-26
Use CanvasContext / url as source of truth - NOREF

sarangj:
## Ticket:

- NOREF 

Some followup cleanup as I've learned a bit more about nextJS / context

The state was a bit hard to follow between the data provided by the
canvas context and the props passed down explicitly. Additionally, it
seems that we can use nextJS native hooks to get / set the canvas index,
allowing us to remove some of the use of window APIs.

Disclosure - the updates to useCanvasContext were generated using
gemini. Everything else is artisanal. Also, note the use of `scroll:
false` in `router.push`. That was a manual update to prevent the focus
from being disrupted when paging to the next canvas.

### Checklist:

### DIFF
--- a/app/src/components/items/uv/universalViewer.tsx
+++ b/app/src/components/items/uv/universalViewer.tsx
@@ -11,7 +11,6 @@ import { useCanvasContext } from "../../../context/CanvasProvider";
 export type UniversalViewerProps = {
   config?: any;
   manifestId: string;
-  canvasIndex?: number;
   captureUuidToIdx: { [uuid: string]: number };
   onChangeCanvas?: (manifest: string, canvas: string) => void;
   onChangeManifest?: (manifest: string) => void;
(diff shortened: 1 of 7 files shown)


app/src/components/items/viewer/viewer.tsx
sarangj: i think these callbacks can probably be removed as well.

app/src/components/items/uv/universalViewer.tsx
```diff
@@ -73,25 +61,17 @@ const UniversalViewer: React.FC<UniversalViewerProps> = React.memo(
...
-      //     .filter(k => k.toLowerCase().includes("uv") || k.toLowerCase().includes("universal"))
-      //     .forEach(k => localStorage.removeItem(k));
-      // } catch {}
+      if (uv) {
+        uv._assignedContentHandler?.publish(
+          BaseEvents.CANVAS_INDEX_CHANGE,
+          canvasIndex
+        );
+      }
+    }, [canvasIndex, uv]);
```
sarangj, line 70: I believe just using `canvasIndex` as a dep here removes the need to track the last canvas index ref.

keithbauer (approved):
nice, looks good in the vercel preview

PR #545, opened 2026-03-31
Import uv css in uv component NOREF

sarangj:
## Ticket:

NOREF

Noticed this error popping up all the time when loading an item:

Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

The import trace pointed to this css import. Moving it to the uv component seems to fix it.

### Checklist:


emretetiknypl:
@avertrees are you able to reproduce this? I clicked around on DC and I wasn't seeing the error Sarang mentioned in the PR description come up at all. Should we close this?

PR #552, opened 2026-04-10
Make uv.html available for embed - NOREF

sarangj:
## Ticket:

NOREF

The UV package comes with a uv.html file that is handy to serve for
embed links. Use a webpack extension to copy the file into /_next/static
so it can be served publicly on DC.

Note - the embed link in the UV component refers to the /uv.html route.
While I *should* be able to override this using the ShareDialogue's
`embedPath` config option, that seems to not be working.

The alternative is to copy the uv files to the public/ dir at build
time, but the webpack extension doesn't seem to do this properly - I
could bake something into the build that directly copies to the repo's
public/ dir, but I'm going to try to avoid that for now - this is mostly
a proof of concept to try to use UV in the archives portal embed link

Second commit also consolidates the uv config into one place for clarity

## Open questions

I put the uvConfig into the public/ folder to make it available for archives to reference - however, it might make sense to just have archives supply its own uv config (do we need / want to show download options and such, idk).

update my local archives portal to pull point at a DC uv.html on this vercel preview and see that the embed link loads

https://github.com/user-attachments/assets/aa864440-ea45-45fc-bdac-47506caeac75

### Checklist:


package.json
```diff
@@ -72,6 +72,7 @@
     "@types/newrelic": "9.14.3",
     "@types/node": "^24.2.1",
     "@types/ua-parser-js": "0.7.39",
+    "copy-webpack-plugin": "^14.0.0",
```
sarangj, line 75: i should remove the '^'

app/embedded-item/[uuid]/page.tsx
```diff
@@ -0,0 +1,24 @@
...
+  params,
+  searchParams,
+}: ItemEmbedProps) {
+  revalidatePath("/");
+  const queryString = new URLSearchParams({
+    manifest: `${process.env.COLLECTIONS_API_URL}/manifests/${params.uuid}`,
+    config: "/uvConfig.json",
+    ...searchParams,
+  }).toString();
+  console.log(queryString);
```
Comment (sarangj) [inline app/embedded-item/[uuid]/page.tsx:22]: remove

sarangj:
superseded by https://github.com/NYPL/digital-collections/pull/601

PR #555, opened 2026-04-16
DR-4029  facet filter names tests

jbdalton:
## Ticket:

- JIRA ticket [DR-4029](https://newyorkpubliclibrary.atlassian.net/browse/DR-4029)

These are tests for the expanded facet-filter on Names.  The modal now allows for all of the names to be displayed, so these tests verify that "page11," etc. are available for a big result-set.  The tests also check the auto-complete feature for a search within the names list.

The Spec uses the the same `serial.test()` framework for playwright that we have used previously on other open-modal tests.  This allows us to iterate through several actions with the modal remaining open, rather than parallelizing those checks (which is the playwright default).  This gives us more speed, and a ux-driven test-flow for these types of specs.

## Open questions

I'm using the same "print" query that appeared on the Jira ticket's original description.  It brings back a gazillion results, which lets us test the "more than 10 pages" feature.   But it isn't too deterministic, since "print" will reliably always bring back enough results.  (We've got alot of prints!)

The test basics are:
1. open all-names modal
2. click page-next enough times to get to the page 10-11 boundary.
3. test auto-complete names-search with "P, O, S," verifying that each letter brings the total down (ie refines/filters on the specific search stem)
4. clearing the search term should display all names again.

run test-suite locally on production build (`npm run build`) with same workers as prod:

`npx playwright test --workers=2 --retries=1`

### Checklist:

[DR-4029]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


playwright/pages/filter-value-expand.page.ts
```diff
@@ -0,0 +1,156 @@
...
+  async verifyAutocompleteFlow() {
+    await this.clearSearch();
+
+    const initialLastPage = await this.getLastPageNumber();
+    let previousLastPage = initialLastPage;
+
+    // Loop through each character in the stem (e.g. "P", "O", "S")
+    for (const char of FilterValueExpandedPage.SEARCH_STEM) {
+      await this.modalSearchInput.press(char);
+      await this.page.waitForTimeout(400); // Give the pagination a moment to rebuild
```
samanthaandrews: Is it possible to remove this and replace it with a more robust solution? Ideally, we could wait for something on the page to appear that wasn't there before, i.e., expect [expect(locator).toBeVisible({ timeout: 400 })](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-visible)

playwright/pages/filter-value-expand.page.ts
```diff
@@ -0,0 +1,156 @@
...
+        name: `Page ${i}`,
+        exact: true,
+      });
+      await pageBtn.click();
+
+      // 1. Verify pagination navigation's active state
+      await expect(pageBtn).toHaveAttribute("aria-current", "page");
+
+      // 2. Verify name refresh (concurrently with navigation)
+      await expect(radios.last()).not.toHaveAttribute(lastRadioOldValue || "");
```
samanthaandrews: Looking at the [Playwright docs]([expect(locator).toBeVisible()](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-visible)), I think `toHaveAttribute` is expecting two arguments. The first being the attribute you're looking for, and the second being the value of the attribute. So we should make sure this is testing what we mean it to be testing.

playwright/pages/filter-value-expand.page.ts
```diff
@@ -0,0 +1,156 @@
...
+export class FilterValueExpandedPage extends SearchPage {
+  readonly viewAllNamesButton: Locator;
+  readonly valueExpandedModal: Locator;
+  readonly modalSearchInput: Locator;
+
+  static nameResultsUrl: string = "/search/index?q=print";
+  static SEARCH_STEM: string = "POS";
+  static TARGET_NAME: string = "Posada, José Guadalupe, 1852-1913";
+
+  constructor(page: any) {
```
samanthaandrews: this should be `constructor(page: Page)`. You should be able to import `Page` from @playwright/test

playwright/tests/filter-value-expand-name.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    ).toBeVisible();
+
+    await expect(
+      filterPage.valueExpandedModal.getByPlaceholder(/search names/i)
+    ).toBeVisible();
+
+    // Name List should Display 10 names
+    await expect(filterPage.valueExpandedModal.getByRole("radio")).toHaveCount(
+      10
+    );
```
samanthaandrews: Nit: We could make this slightly easier to update in the future and easier to understand for future devs by extracting out the "magic" numbers here and putting them at the top of the file using a variable name, e.g., `const numItemsListShouldDisplay = 10` or moving it to the page file.

playwright/tests/filter-value-expand-name.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    await expect(
+      filterPage.valueExpandedModal.getByRole("button", { name: /close/i })
```
samanthaandrews: Nit: Could extract some of these into the `page` file so they all elements live in one place

samanthaandrews (changes requested):
A few small things to update. It's looking good!

playwright/pages/filter-value-expand.page.ts
jbdalton: Good catch!  I've added the value to that attribute check now, so it won't keep "accidentally" passing, where the value to actually be wrong.

playwright/pages/filter-value-expand.page.ts
jbdalton: Looks for existence via (expect/await) of a different number for the page-total  here now, and waits until it "sees"/confirms it.  So it won't need the 400ms wait.  And the "is the value really what we expect" filters then check after this confirmation to make sure the math is right

playwright/pages/filter-value-expand.page.ts
jbdalton: added Page here and to import.

playwright/tests/filter-value-expand-name.spec.ts
jbdalton: added to a static constant in the POM file `NAMESLIST_COUNT`

playwright/tests/filter-value-expand-name.spec.ts
jbdalton: Good catch too!  I knew I had left these strewn in the spec on the first push.  This section's now refactored into `verifyExpandedNamesModal()` and `verifyDefaultNamesList()` methods in the POM.   Each has a test calling it from the spec in a single line now, similar to how the rest of the spec's workflow is set up.

jbdalton:
I think the changes are ready to re-review.  I've had to add `fixme()` in two spots in unrelated files to get the suite to pass, due to some temporary qa-environment instability.  It doesn't affect these tests, and as soon as the issues have been cleared up, those two files will get un-fixedme'd.

playwright/pages/filter-value-expand.page.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    this.viewAllNamesButton = this.page.getByRole("button", {
+      name: /view all names/i,
+    });
+    this.valueExpandedModal = this.page.getByRole("dialog", { name: "Names" });
+    this.modalSearchInput =
+      this.valueExpandedModal.getByPlaceholder(/search names/i);
+  }
+
+  async openAllNames() {
```
samanthaandrews, line 25: I don't see this function being called anywhere. Should we remove?

playwright/tests/filter-value-expand-name.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    // Assign to global variable and load the page object
+    filterPage = new FilterValueExpandedPage(page);
+    await filterPage.loadPage(FilterValueExpandedPage.nameResultsUrl);
+  });
+
+  test.afterAll(async () => {
+    await filterPage.page.context().close();
+  });
+
+  test("Open full names list and click view-all-names", async () => {
```
samanthaandrews: I think a more accurate test name here might be something like, "Names dropdown opens and shows view-all-names button" since we aren't actually clicking the view all names button in this test

playwright/tests/item-media.spec.ts
```diff
@@ -40,7 +40,9 @@ test.describe("Verify Video Viewer Controls", () => {
   });
 });
 
-test.describe("Verify Download Actions", () => {
+// FIXME:  Temporarily bypassing this due to UV change on download icon/description
```
samanthaandrews: These two FIXME notes should either be tracked as separate tickets (and referenced in the FIXME comment) or mentioned in the PR description so reviewers know they're intentional.

playwright/tests/filter-value-expand-name.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  JUSTIFICATION FOR .serial USAGE:
+  
+  These tests intentionally use .serial to maintain browser state across multiple test steps. This deviates from Playwright's recommendation for isolated tests for the following reasons:
+  
+  1. Performance: Avoids 6 separate page loads (60% faster: 16.57s vs 26.23s)
+  2. Resource efficiency: Uses single worker instead of 6 parallel workers
+  3. User journey testing: Simulates realistic workflow of opening modal once and applying multiple filters in sequence
+  4. State accumulation: Each filter builds on previous selections
+
+  Alternative considered: Single comprehensive test, but rejected because individual filter
```
samanthaandrews: Nit: I think this comment got cut off

samanthaandrews (changes requested):
The one thing I'm noticing about these tests is that they may break if the browser window changes in size, since it's possible that the page numbers wouldn't display in the same way on larger or smaller screens. It's not something that needs to be addressed in this PR, but something to follow up on, especially thinking about running any of these tests on a mobile device.

Otherwise, just a few additional small comments.

playwright/pages/filter-value-expand.page.ts
jbdalton: Turns out I'd left this as an unfinished call.  It's being called properly from the spec now, and replaces the interactions that were in that spec block.

playwright/tests/filter-value-expand-name.spec.ts
jbdalton: Updated the test name to reflect the action being tested.

playwright/tests/item-media.spec.ts
jbdalton: Down to only one fixme now!   I have created a ticket for this:  [DR-4130](https://newyorkpubliclibrary.atlassian.net/browse/DR-4130)

playwright/tests/filter-value-expand-name.spec.ts
jbdalton: updated.

jbdalton:
Ready for another look I think?   In terms of the pagination or numbers possibly changing on different viewport sizes, this is something I'm actively thinking about now.  One step I'm looking into (for iphone at least) is to get the suite running on webkit.   Also looking into adding configs for iphone/mobile testing.

samanthaandrews (approved):

PR #561, opened 2026-05-11
DR-4087 Add cta_click GA events for various buttons on the Item and Collection pages

emretetiknypl:
## Ticket:

- JIRA ticket [DR-4087](https://newyorkpubliclibrary.atlassian.net/browse/DR-4087)

This PR adds a `cta_click` GA event to the following buttons:

* Collection Page
    * Finding Aid
    * Catalog Record
* Item Page
    * Finding Aid
    * Research Catalog
    * Order Print

## Open questions

One thing worth noting is that, since the "Order Print" button actually takes you to a different domain, there already is a GA event that gets fired when it is pressed (`outbound_links`). Since we're now adding a second GA event for this button, GA will group these two events and place the data in the request payload, not query params (which is what all the others do). I highly doubt this will cause any issues, but I just wanted to mention it.

I ran my local digital-collections instance with these change and confirmed in the network tab that each of these CTAs now makes a GA call with the correct parameters.

### Checklist:

[DR-4087]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/collectionMetadata/collectionMetadata.tsx
+++ b/app/src/components/collectionMetadata/collectionMetadata.tsx
@@ -1,4 +1,5 @@
 import { capitalize } from "@/src/utils/utils";
+import { trackCTA } from "@/src/utils/ga4Utils";
 import {
   Flex,
   Heading,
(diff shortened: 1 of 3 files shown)


sarangj:
nice, looks great in the vercel preview

sarangj (approved):

PR #563, opened 2026-05-13
Dr 4091 type ahead

avertrees:
## Ticket:

- JIRA ticket [DR-4092](https://newyorkpubliclibrary.atlassian.net/browse/DR-4092)

Implements type-ahead autocomplete suggestions in the global search bar.

- **`app/api/suggest/route.tsx`** — New Next.js server-side proxy to `collections-api /search/suggest`. Keeps the auth token server-side; returns `{ suggestions: [] }` gracefully on error or queries under 3 characters.
- **`app/src/components/search/searchSuggestions.tsx`** — New accessible suggestion dropdown. Implements ARIA 1.2 Combobox/Listbox pattern (`role="listbox"`, `role="option"`, `aria-selected`). Title highlighting uses the same `formatHighlightText` → `highlightTitleWords` → `parse` pipeline as general search result cards, so matched words are bolded consistently.
- **`app/src/components/search/search.tsx`** — Adds suggestion state, 300ms debounced fetch, keyboard navigation (↑↓ Enter Escape), and ARIA combobox attributes on the input (`role="combobox"`, `aria-expanded`, `aria-autocomplete="list"`, `aria-controls`, `aria-activedescendant`).
- **`app/src/components/search/dcSearchBar.tsx`** — Threads `autoComplete`, `onKeyDown`, and `additionalInputProps` through to the underlying `TextInput` to support the combobox ARIA attributes.

[DR-4092]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/api/suggest/route.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const res = await fetch(apiUrl, {
+      headers: {
+        "x-nypl-collections-api-key":
+          process.env.COLLECTIONS_API_AUTH_TOKEN ?? "",
+      },
+    });
+
+    if (!res.ok) {
+      return NextResponse.json({ suggestions: [] });
+    }
```
sarangj: It looks like we're handling errors client side as well - I feel like this could be confusing to debug. Since we're handling the errors on the client, what do you think of having this return a 500 if the collections-api returns a bad response?  That way, if we see the suggestions aren't working, we'll know in the browser console it's because there was an error here?

app/api/suggest/route.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    const data = await res.json();
+    return NextResponse.json(data);
+  } catch {
+    return NextResponse.json({ suggestions: [] });
+  }
```
sarangj: Same idea / feedback here!

app/api/suggest/route.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const GET = async (req: NextRequest) => {
+  const q = req.nextUrl.searchParams.get("q") ?? "";
+
+  if (q.length < MIN_QUERY_LENGTH) {
+    return NextResponse.json({ suggestions: [] });
+  }
+
+  const apiUrl = `${
+    process.env.COLLECTIONS_API_URL
+  }/search/suggestions?q=${encodeURIComponent(q)}`;
```
sarangj: can / should we put this `fetch` into the api client?

app/src/components/search/search.tsx
```diff
@@ -24,6 +64,13 @@ const Search = () => {
     router.push(searchUrl);
   };
 
+  const handleSuggestionSelect = (title: string) => {
+    setKeywords(title);
+    closeSuggestions();
+    searchManager.setLastFilter(null);
+    router.push(`/search/index?q=${encodeURIComponent(title)}`);
```
sarangj, line 76: one question on this! If we are already on a search page with filters, *should* we include those filters when we use the autosuggestion? Maybe a question for Lana

app/src/components/search/search.tsx
```diff
@@ -45,19 +124,39 @@ const Search = () => {
...
-          id: "search-text",
-          labelText: "Search keyword(s)",
-          name: "keywords",
-          onChange: handleTextChange,
-          value: keywords,
-          placeholder: "Search keyword(s)",
-        }}
-        onSubmit={(e) => handleSubmit(e)}
-      />
+      {/* position="relative" creates the stacking context for the absolute dropdown */}
```
sarangj, line 132: thank you for this comment

sarangj (commented):
looks sick, just some thoughts on error handling and such! also, could we rename the api endpoint to be a noun like in the collections api, ie, '/api/suggestions/'?

app/src/components/search/search.tsx
avertrees: ah great catch

app/api/suggest/route.tsx
```diff
@@ -0,0 +1,39 @@
...
```
sarangj: still think we should put this in https://github.com/NYPL/digital-collections/blob/b9f1ba47bb6ac267980c7ab4fe04680220c3cabe/app/src/utils/apiClients/apiClients.tsx but could be in a later PR

sarangj (approved):

PR #565, opened 2026-05-26
improve experience for screenreaders

avertrees:
a11y update

### DIFF
--- a/app/src/components/search/search.tsx
+++ b/app/src/components/search/search.tsx
@@ -116,6 +116,12 @@ const Search = () => {
   const activeDescendant =
     activeIndex >= 0 ? `${LISTBOX_ID}-option-${activeIndex}` : undefined;
 
+  const statusMessage = isOpen
+    ? `${suggestions.length} suggestion${
+        suggestions.length === 1 ? "" : "s"
+      } available. Use up and down arrow keys to navigate.`
+    : "";
+
   return (
     <Box
       id="search-wrapper"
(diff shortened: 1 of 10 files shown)


sarangj (approved):

app/src/components/search/search.tsx
```diff
@@ -1,62 +1,34 @@
...
+    suggestions,
+    activeIndex,
+    isOpen,
+    isTouch,
+    wrapperRef,
+    closeSuggestions,
+    handleKeyDown,
+    handleWrapperBlur,
+    statusMessage,
+  } = useSearchCombobox({ keywords, listboxId: LISTBOX_ID });
```
sarangj, line 32: new hook just dropped 🚀

app/src/components/search/searchSuggestions.tsx
```diff
@@ -86,6 +95,25 @@ const SearchSuggestions = ({
           </Box>
         );
       })}
+      {isTouch && (
+        <Button
+          id={`${listboxId}-close-btn`}
+          variant="text"
+          onClick={() => {
+            onClose();
+          }}
```
sarangj: could just do `onClick={onClose}`

app/src/hooks/useSearchCombobox.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Manages all combobox state and side effects for the search typeahead.
+ */
+
+export function useSearchCombobox({
+  keywords,
+  listboxId,
+}: UseSearchComboboxOptions): UseSearchComboboxReturn {
+  const [suggestions, setSuggestions] = useState<SuggestResult[]>([]);
+  const [activeIndex, setActiveIndex] = useState(-1);
+  const [isOpen, setIsOpen] = useState(false);
```
sarangj: one small question - it seems like `isOpen` always corresponds to the `suggestions` array being non-empty. Could we replace the state here with just `const isOpen = suggestions.length > 0`, and remove the need for `setIsOpen`?

sarangj (commented):
nice - one question. When I hit the vercel preview, and i use my keyboard to navigate down to one of the suggestions and hit enter, it looks like it still submits the search, as opposed to just populating the search box. I assumed one of the fixes was to have hitting 'Enter' *only* populate the search box, but maybe I'm missing something where that *only* applies to screenreaders

Also, I'm looking at those tests

avertrees:
@sarangj you were right about submitting enter - guess that wasn't fixed. should be now - also the other updates were made to clean up the code

sarangj (approved):
wooo

PR #566, opened 2026-05-26
DR-4107 All Maps Frontend Implementation

emretetiknypl:
## Ticket:

- JIRA ticket [DR-4107](https://newyorkpubliclibrary.atlassian.net/browse/DR-4107)

- Reads the newly-added `hasAllMapsData` field from the backend.
- If this field is true, shows a button allowing the user to toggle between regular view and map view.
- Map view shows the capture superimposed on a map viewer. A brief animation centers the capture in the map view pane. 

Ran it locally to check that it renders correctly.

### Checklist:

[DR-4107]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/components/items/maps/allMaps.tsx
```diff
@@ -0,0 +1,88 @@
...
+      );
+      const imageId = captureWithMapData?.imageId;
+      console.log("using this imageId for the allmaps viewer: ", imageId);
+
+      const iiif_url = imageId
+        ? `https://iiif.nypl.org/iiif/2/${imageId}`
+        : null;
+
+      // generateId is async and returns a Promise
+      const hashed_iiif_image_id = await generateId(iiif_url);
```
sarangj: let's stick with camelCase var names here!

app/src/components/items/maps/allMaps.tsx
```diff
@@ -0,0 +1,88 @@
...
+      });
+
+      const captureWithMapData = item.captures.find(
+        (capture) => capture.hasAllMapsData
+      );
+      const imageId = captureWithMapData?.imageId;
+      console.log("using this imageId for the allmaps viewer: ", imageId);
+
+      const iiif_url = imageId
+        ? `https://iiif.nypl.org/iiif/2/${imageId}`
```
sarangj, line 46: I think we generally use iiif v3 urls here, (so we could / should swap the '2' to a '3').

As I look at this a bit though - I feel like we should generate this url on the server. All the other IIIF urls come from there, and it might be a bit weird if you're hitting the qa collections-api, and this url is for prod.

We can leave this as is and do a followup ticket to just add the iiif image url to the captures object (or not, just a thought, not a hard requirement).

app/src/context/CanvasProvider.tsx
```diff
@@ -15,6 +17,8 @@ export const CanvasProvider = ({ children }: { children: React.ReactNode }) => {
   const pathname = usePathname();
   const searchParams = useSearchParams();
 
+  const [isMapView, setIsMapView] = useState(false);
```
sarangj: As I look at this a bit, I feel like this state could just go in the map view component (or in the viewer.tsx?). Don't really see why this needs to be in the canvas provider, we really just need some state tied to the url, and the ability to update the maps param in the url. 

We could also define a new context / provider for this, but it *feels* like plain state would suffice here.

emretetiknypl: I agree that it doesn't really belong in CanvasProvider. I think we probably do need to define another provider though, since this state needs to be used by both `app/src/components/items/viewer/viewer.tsx` and `app/src/components/items/overview/external/overview.tsx` (unless we want to drill down props). So I could make a new provider (`MapViewProvider`?) orrrr we could keep this logic in `CanvasProvider` and rename `CanvasProvider` to a more generic `ItemProvider`. What do you think?

sarangj: if it's not too much trouble, I think I'd prefer to split this into a separate provider!

app/src/components/items/maps/allMaps.tsx
emretetiknypl: I don't know if there's flexibility on AllMaps's end to use any image url. We may have to use the same url that was used to enter the annotation to begin with. I'm actually not entirely sure how AllMaps makes the connection between a particular image and an annotation, given that an image can be accessed in multiple different ways. I've asked the AllMaps people about this. Stand by

app/src/context/CanvasProvider.tsx
emretetiknypl: done!

app/src/components/items/maps/allMaps.tsx
emretetiknypl: Did some discovery on this towards EOD yesterday. I have confirmed that AllMaps doesn't have the flexibility we'd currently need to use the v3 image URLs.

AllMaps uses the IIIF id to distinguish between images. So to AllMaps, `https://iiif.nypl.org/iiif/2/1524616/info.json` and `https://iiif.nypl.org/iiif/3/1524616/info.json` are distinct images (even if they have the same NYPL image id). I'm guessing that all our MapWarper data was entered when our IIIF server was on v2, because I can always find the image when I construct the URLs with v2 and cannot find them when I construct the URLs with v3. So this means that the integration will fail if we use v3 image URLs.

Bert did say that they're working on an API that would enable us to migrate the v2 URLs to v3 on AllMaps's end. He's also willing to manually run that migration for us.

My recommendation would that for now, we should hardcode the v2 url so that we can release in time for the demo (if that's a serious concern). Then, we can get the AllMaps people to migrate our URLs. Then we can start pulling our URLs from collections api, which will return the v3 ones. We're still not totally in the clear, because when we update our image API to v4, we will run into the same issue. But hopefully by then AllMaps will have developed a more durable solution for doing these mappings. (Bert said they intend to do this, but it doesn't seem like it's going to get done soon. [Here's](https://iiif.slack.com/archives/C04C22NDTCL/p1780001235257539?thread_ts=1779982150.746579&cid=C04C22NDTCL) a link to the slack thread where Bert explained all this to me.)

sarangj (approved):

sarangj (approved):

PR #568, opened 2026-06-02
Dr 4130 - Fix download overlay tests

jbdalton:
## Ticket:

- JIRA ticket [DR-4130](https://newyorkpubliclibrary.atlassian.net/browse/DR-4130)

The download overlay button tests have been problematic for awhile.  This test-block has at times required 2 tries to succeed, which, though tolerable for CI purposes, wasn't ideal.   Recently they started failing even more often.  They shouldn't need a IIIF success, in terms of the image actually _displaying_ to pass.  The test itself is just a check that the overlay opens, has the correct download choices for a sample item, and can be closed.

These code changes remove the IIIF dependency, and "force" the test to click the dl-button that it "sees" in the DOM, via `.click({ force: true })`, before the UV gets a chance to throw a blanket on the whole image viewport due to qa-iiif slowness, blocking, etc.  Since we don't need the IIIF server to be able to click the download button, we use a try-catch to have playwright try the click a few times quickly (with a 750ms wait in between).   

These two changes seem to be working well for any race conditions where playwright may be over-eager to get on with things too quickly, before the DOM "settles."

I tried it without the catch/try, and got a 25% failure rate in 4 runs of the test-suite.  When tried with the catch/try, I got a 100% success rate for 6 runs of the test suite.

`npm run build --workers=2`

### Checklist:

[DR-4130]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/pages/item-media.page.ts
+++ b/playwright/pages/item-media.page.ts
@@ -90,11 +90,10 @@ export default class ItemMediaPage {
   // Downloads
 
   // Opens and verifies download overlay modal
-
   async openDownloadMenu(): Promise<void> {
     await this.downloadButton.waitFor({ state: "visible" });
 
-    // Retry click up to 3 times if the browser temporarily swallows the DOM click event
+    // Retry click up to 5 times if the browser temporarily swallows the DOM click event
     for (let attempt = 1; attempt <= 5; attempt++) {
       await this.downloadButton.click({ force: true });
 
@@ -107,9 +106,7 @@ export default class ItemMediaPage {
         });
         await expect(this.downloadOptionSmall).toBeEnabled();
         return;
-      } catch (error) {
-        console.warn(`Retrying download menu trigger: attempt ${attempt}/5`);
-      }
+      } catch (error) {}
     }
 
     throw new Error("Failed to open the download menu after 5 attempts.");
--- a/playwright/pages/item-media.page.ts
+++ b/playwright/pages/item-media.page.ts
@@ -91,7 +91,6 @@ export default class ItemMediaPage {
 
   // Opens and verifies download overlay modal
 
-  // NEW Option 1.0
   async openDownloadMenu(): Promise<void> {
     await this.downloadButton.waitFor({ state: "visible" });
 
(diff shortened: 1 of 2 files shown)


playwright/pages/item-media.page.ts
```diff
@@ -91,12 +91,27 @@ export default class ItemMediaPage {
...
   async openDownloadMenu(): Promise<void> {
-    await this.downloadButton.click();
-
-    // Wait for the popup to be visible
-    await this.downloadOptionSmall.waitFor({ state: "visible" });
+    await this.downloadButton.waitFor({ state: "visible" });
+
+    // Retry click up to 5 times if the browser swallows the DOM click event
+    for (let attempt = 1; attempt <= 5; attempt++) {
+      await this.downloadButton.click({ force: true });
```
sarangj, line 98: so this `force: true` is the main fix - do we still need the 5 retries in that case?

sarangj (commented):
nice! Just one question, but otherwise, glad we got it working!

playwright/pages/item-media.page.ts
jbdalton: I think it probably needs at least the 1 retry.  The retries don't reload the page, they are just a "hit-loop" where playwright asks "can I see it?" ... pause ...."can I see it now?" ... pause ... "can I really see it now?"  So it's that 700ms pause between the tries which is doing the "work" of pumping the brakes on playwright's natural over-eagerness to move on from the DOM at that moment, encounter an unloaded iiif/UV overlay, and fail.  

I'll set it to 3 tries and keep an eye on it.   If it ever hits a 3rd retry and fails, we'll see this announced in the CI log.  So far in testing on localhost it has "passed" on every first retry.

jbdalton: Reran this on localhost with 3 catch/try loops, and noticing this run did have to re-check for the dl button twice.   It's probably safer to leave it at 5 for now, in case it sometimes takes 5 x 700ms to get to a viable click button view?  Ignore the "n/5" text below: I hadn't changed the static text back to 3, but it was running on only a 3 try catch loop here.


```
Running 4 tests using 2 workers

  ✓  1 [chromium] › playwright/tests/item-media.spec.ts:32:7 › Verify Video Viewer Controls › primary video-specific controls are visible (1.3s)
  ✓  2 [chromium] › playwright/tests/item-media.spec.ts:13:7 › Verify Image Viewer Controls › primary image-specific controls are visible (855ms)
  ✓  3 [chromium] › playwright/tests/item-media.spec.ts:49:7 › Verify Download Actions › should display all available download options (2.2s)
  ✓  4 [chromium] › playwright/tests/item-media.spec.ts:73:7 › Verify Purchase Option › should display order text and correct link (516ms)
Retrying download menu trigger: attempt 1/5
Retrying download menu trigger: attempt 2/5

  4 passed (8.0s)

```

playwright/tests/item-media.spec.ts
```diff
@@ -43,7 +43,7 @@ test.describe("Verify Video Viewer Controls", () => {
 // FIXME: dl modal is not showing correct links within playwright even though they can be observed
```
samanthaandrews: Nit: Can we remove this comment FIXME comment now?

playwright/pages/item-media.page.ts
samanthaandrews: This isn't an incorrect implementation, but Playwright does have some utilities for retries that we should probably at least try and see if they're effective. Knowing this would be useful to share with the rest of the QA team.

The remaining two are:
- `expect.poll()` <- probably not the right fit here, but good to know about https://playwright.dev/docs/test-assertions#expectpoll
- `expect.toPass()` https://playwright.dev/docs/test-assertions#expecttopass

Here's what I `expect.toPass` might look like in this test:

```
async openDownloadMenu(): Promise<void> {
  await this.downloadButton.waitFor({ state: "visible" });

  await expect(async () => {
    await this.downloadButton.click({ force: true });
    await expect(this.downloadOptionSmall).toBeVisible();
  }).toPass({ timeout: 5000, intervals: [750, 750, 750, 750, 750] });

  await expect(this.downloadOptionSmall).toBeEnabled();
}
```

samanthaandrews (changes requested):
Love this small PR. One suggestion below that I think is worth trying out.

playwright/pages/item-media.page.ts
jbdalton: I tried this - with some additional logging - and it sporadically failed on the same "empty download" box that the original version struggles with.  

The issue is that `toPass()` re-executes the entire block upon a failed assertion. If the toggle click is inside that block, it ends up re-clicking the button and shutting the menu before the lazy-loaded Next.js chunks can finish rendering in the DOM.  Even with the wait pushed up to 20 seconds, the click toggle gets caught in that same open/close loop.

Going to revert to the explicit try/catch loop.  Because it lives outside of playwright's internal test-runner logic, it forces playwright to stop being so over-eager, pause for that one 750ms break, try again, take a breath (for the full 750ms!) and then try again.   Calm down, playwright, it's ok! 😄

playwright/tests/item-media.spec.ts
jbdalton: removed.

samanthaandrews (approved):

PR #570, opened 2026-06-05
Add opacity control to AllMaps viewer

emretetiknypl:
## Ticket:

- JIRA ticket [DR-4144](https://newyorkpubliclibrary.atlassian.net/browse/DR-4144)

Add an OpacityControl to the All Maps viewer.

Deployed it locally.

### Checklist:

[DR-4144]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/items/maps/OpacityControl.tsx
+++ b/app/src/components/items/maps/OpacityControl.tsx
@@ -1,8 +1,7 @@
 import { IControl, Map } from "maplibre-gl";
 import { WarpedMapLayer } from "@allmaps/maplibre";
 import React from "react";
-import { createRoot, Root } from "react-dom/client";
-import { Icon, DSProvider } from "@nypl/design-system-react-components";
+import { Icon } from "@nypl/design-system-react-components";
 
 const STYLES = {
   container: {
@@ -32,9 +31,7 @@ const STYLES = {
   },
 };
 
-type WarpedMapLayer = typeof WarpedMapLayer;
-
-const OpacityControlComponent = ({
+export const OpacityControlComponent = ({
   warpedMapLayer,
 }: {
   warpedMapLayer: WarpedMapLayer;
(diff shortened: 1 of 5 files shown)


app/src/components/items/maps/OpacityControl.tsx
```diff
@@ -0,0 +1,88 @@
...
+  },
+};
+
+/**
+ * A custom MapLibre control to change the opacity of an AllMaps WarpedMapLayer.
+ */
+export class OpacityControl implements IControl {
+  private _map: Map | undefined;
+  private _container: HTMLDivElement | undefined;
+  private _warpedMapLayer: any;
```
sarangj: could we add a type here instead of `any`? Seems like we could just import the `WarpedMapLayer` or define an interface that matches it in this module

app/src/components/items/maps/OpacityControl.tsx
```diff
@@ -0,0 +1,88 @@
...
+    Object.assign(slider.style, STYLES.slider);
+
+    slider.addEventListener("input", (e) => {
+      const opacity = parseFloat((e.target as HTMLInputElement).value);
+      this._warpedMapLayer.setOpacity(opacity);
+    });
+
+    this._container.appendChild(slider);
+
+    return this._container;
```
sarangj: hmmm, would it be possible to extract this into a react component that gets returned by `onAdd`? If it's too unwieldy to do so, we could add a TODO or something.

sarangj (commented):
overall looks good, just one small and one bigger question!

app/src/components/items/maps/OpacityControl.tsx
emretetiknypl: good catch! in order to use it, I'll have to define the type like this:

`type WarpedMapLayer = typeof WarpedMapLayer;`

Otherwise I get "cannot use namespace as type"

app/src/components/items/maps/allMaps.tsx
```diff
@@ -93,7 +105,16 @@ const AllMapsViewer = ({ item }: ItemProps) => {
         width: "100%",
         height: "100%",
       }}
-    />
+    >
+      {opacityControlContainer &&
+        warpedMapLayerRef.current &&
+        createPortal(
```
emretetiknypl, line 115: @sarangj I was reading about `createPortal`... it's a way to render a component within MapLibre's DOM tree (which is outside of React) while still retaining the context of the React component tree. So if we use this we already have access to `DSProvider` and don't need to add it again.

sarangj (approved):

PR #572, opened 2026-06-12
DR-4149 Add GA event for Maps button

emretetiknypl:
## Ticket:

- JIRA ticket [DR-DR-4149](https://newyorkpubliclibrary.atlassian.net/browse/DR-4149)

- Adds a `cta_click` GA event when the `View on Map` button is pressed.
- 

Tested on local instance of DC.

### Checklist:

### DIFF
--- a/app/src/components/items/overview/external/overview.tsx
+++ b/app/src/components/items/overview/external/overview.tsx
@@ -61,7 +61,17 @@ const ExternalLinksOverview = ({ catalogLink, archivesLink, isInAllMaps }) => {
 
         {isInAllMaps && isLargerThanSmallTablet ? (
           <Button
-            onClick={handleMapViewToggle}
+            onClick={() => {
+              handleMapViewToggle();
+              // Only track the CTA if we're switching to the map view, not if we're closing it
+              if (!isMapView) {
+                trackCTA(
+                  "View on Map",
+                  window.location.href,
+                  "Item Page View on Map"
+                );
+              }
+            }}
             id={"all-maps-btn"}
             aria-label={`view in All Maps Viewer`}
             variant="secondary"
--- a/app/src/components/items/overview/external/overview.tsx
+++ b/app/src/components/items/overview/external/overview.tsx
@@ -62,7 +62,6 @@ const ExternalLinksOverview = ({ catalogLink, archivesLink, isInAllMaps }) => {
         {isInAllMaps && isLargerThanSmallTablet ? (
           <Button
             onClick={() => {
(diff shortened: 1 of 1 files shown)


app/src/components/items/overview/external/overview.tsx
```diff
@@ -61,7 +61,17 @@ const ExternalLinksOverview = ({ catalogLink, archivesLink, isInAllMaps }) => {
...
+              handleMapViewToggle();
+              // Only track the CTA if we're switching to the map view, not if we're closing it
+              if (!isMapView) {
+                trackCTA(
+                  "View on Map",
+                  window.location.href,
+                  "Item Page View on Map"
+                );
+              }
+            }}
```
sarangj, line 74: maybe worth considering if we do this before the toggle - if the toggle throws an error, do we still want to track the click? Maybe a small and philosophical thing to think about lol

sarangj (approved):

app/src/components/items/overview/external/overview.tsx
emretetiknypl: I can add some try/catch/exception handling!

emretetiknypl: Actually I don't know if we want to do that... we probably only want to track it if it's working? lol

sarangj: ya, maybe a question for Lana - do we want to track that someone was interested enough in clicking the button, even if something broke when they did?

emretetiknypl: I asked Aussie about this and he gave a thoughtful reply here: https://newyorkpubliclibrary.atlassian.net/browse/DR-4149?focusedCommentId=168532

TLDR we _do_ want the event to fire even if the functionality is broken. I'll add try/finally

sarangj: do we need a try / catch or could we just move the tracking code above the toggle call?

emretetiknypl: we can do that too

emretetiknypl: actually, isn't the whole point moot because the actual logic is just toggling a piece of state? So even if toggling that state triggers an error downstream that would presumably be in a different call stack?

emretetiknypl: Updated to move it above the toggle call. Glad we could resolve this weighty technical problem 😎 😌

PR #573, opened 2026-06-12
DR 4107 - Maps overlay tests

jbdalton:
## Ticket:

- JIRA ticket [DR-4107](https://newyorkpubliclibrary.atlassian.net/browse/DR-4107)

These playwright tests check if a map-button exists (it will, since we are passing it a known sample record), and if you click it will it then say "close map."

Also tests whether the AllMaps overlay is produced, and clicks the zoom-in button on that map UI until it has fully zoomed it. We can tell if we are fully zoomed in because the "active" zoom-in toggle no longer accepts clicks and is greyed out.  We test the zoom-out button by simply clicking it once while fully zoomed in to verify that the "zoom-in" button again becomes active.

## Open questions

The biggest hurdle was getting this to run headless, since it's testing WEBGL-type actions for the AllMaps plugin.   Consulted Gemini re. chromedriver workarounds for Mac M-chips that would allow the visible interactions to load and be stable in a headless browser session.   The tests can now be run on either CI or an Apple Silicon environment via a platform-aware set of flags that were added to the playwright.config file called `HEADLESS_OPTIMIZATION_FLAGS`.  Currently, only this one SPEC is calling that import.  

The main purpose of most of the flags was to disable background resource throttling and additionally force the hardware layer to behave sanely, especially on Mac M-chips.   Google and Apple M-chips don't seem to get along when it comes to chromedriver:  what a surprise!

locally run `npx playwright test --workers=2`

Ran test suite locally.

### Checklist:

[DR-4107]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -4,13 +4,42 @@ import { defineConfig, devices } from "@playwright/test";
  * Read environment variables from file.
  * https://github.com/motdotla/dotenv
  */
-// import dotenv from 'dotenv';
-// import path from 'path';
+import dotenv from "dotenv";
+import path from "path";
 // dotenv.config({ path: path.resolve(__dirname, '.env') });
+
+const parsedEnv =
+  dotenv.config({ path: path.resolve(__dirname, ".env.local") }).parsed || {};
 
 /**
  * See https://playwright.dev/docs/test-configuration.
  */
+
+const headlessOptimization = process.env.OPT === "true";
+
+export const HEADLESS_OPTIMIZATION_FLAGS =
+  process.platform === "darwin"
+    ? [
+        // --- LOCAL MAC SILICON ACCELERATION ---
+        "--disable-background-timer-throttling",
+        "--disable-backgrounding-occluded-windows",
+        "--disable-renderer-backgrounding",
+        "--disable-dev-shm-usage",
+        "--no-sandbox",
+        "--enable-gpu",
+        "--use-gl=angle",
+        "--use-angle=metal",
+      ]
+    : [
+        // --- LINUX, ETC CI ACCELERATION ---
+        "--disable-background-timer-throttling",
+        "--disable-backgrounding-occluded-windows",
+        "--disable-renderer-backgrounding",
+        "--disable-dev-shm-usage",
+        "--no-sandbox",
+        // Linux CI handles headless canvas rendering via software wrappers (gl=egl or swiftshader)
+        //'--use-gl=egl'
+      ];
 
 export default defineConfig({
   testDir: "./playwright",
@@ -35,6 +64,10 @@ export default defineConfig({
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: "on-first-retry",
+    /* set optional flags to optimize headless chromedriver visual interactions*/
+    launchOptions: {
+      args: headlessOptimization ? HEADLESS_OPTIMIZATION_FLAGS : [],
+    },
   },
 
   /* Configure projects for major browsers */
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -6,11 +6,8 @@ import { defineConfig, devices } from "@playwright/test";
  */
 import dotenv from "dotenv";
 import path from "path";
+
 // dotenv.config({ path: path.resolve(__dirname, '.env') });
-
-const parsedEnv =
-  dotenv.config({ path: path.resolve(__dirname, ".env.local") }).parsed || {};
-
 /**
  * See https://playwright.dev/docs/test-configuration.
  */
--- a/playwright/pages/item-map-overlay.page.ts
+++ b/playwright/pages/item-map-overlay.page.ts
@@ -63,6 +63,17 @@ export default class ItemMapOverlayPage {
     await expect(this.mapToggleButton).toHaveText(/close map/i);
   }
 
+  async waitForMapReady(): Promise<void> {
+    const canvasContainer = this.page.locator(".maplibregl-canvas-container");
+
+    await expect(
+      canvasContainer,
+      "Map failed to reach 'interactive' state"
+    ).toHaveClass(/maplibregl-interactive/, {
+      timeout: 10_000,
+    });
+  }
+
   // Zoom in functions
   async verifyZoomIn(): Promise<void> {
     await this.mapViewerContainer.scrollIntoViewIfNeeded();
--- a/playwright/pages/item-map-overlay.page.ts
+++ b/playwright/pages/item-map-overlay.page.ts
@@ -68,24 +68,26 @@ export default class ItemMapOverlayPage {
     await this.mapViewerContainer.scrollIntoViewIfNeeded();
     await expect(this.zoomInButton).toBeVisible();
 
-    // click the zoom-in button until we reach a non-active state
-    const MAX_ZOOM_ATTEMPTS = 12;
-
-    for (let i = 0; i < MAX_ZOOM_ATTEMPTS; i++) {
-      const isDisabled = await this.zoomInButton.getAttribute("aria-disabled");
-
-      if (isDisabled === "true") {
-        break;
-      }
-
-      // wait for apx a sec and try again
-      await this.zoomInButton.click({ force: true });
-      await this.page.waitForTimeout(1500);
-    }
-    // verify we broke out of loop due to a match and not to max attemps reached
-    await expect(this.zoomInButton).toHaveAttribute("aria-disabled", "true", {
-      timeout: 1000,
-    });
+    // Keep clicking zoom-in until MapLibre reports max zoom (aria-disabled="true").
+    // expect.poll re-runs the callback until it passes or the timeout elapses,
+    // so there's no fixed sleep or manual attempt counter needed
+    await expect
+      .poll(
+        async () => {
+          const disabled =
+            (await this.zoomInButton.getAttribute("aria-disabled")) === "true";
+          if (!disabled) {
+            await this.zoomInButton.click({ force: true });
+          }
+          return disabled;
+        },
+        {
+          message: "zoom-in button never reached max zoom (aria-disabled=true)",
+          timeout: 20_000,
+          intervals: [500, 1000, 1500], // backoff between polls
+        }
+      )
+      .toBe(true);
   }
 
   // Zoom out functions
--- a/playwright/pages/item-map-overlay.page.ts
+++ b/playwright/pages/item-map-overlay.page.ts
@@ -42,7 +42,6 @@ export default class ItemMapOverlayPage {
 
   // New view-on-map button can be clicked
   async verifyMapIsClosedState(): Promise<void> {
-    await expect(this.page).not.toHaveURL(/\?viewAs=map/);
     await expect(this.mapToggleButton).toBeVisible();
     await expect(this.mapToggleButton).toHaveText(/view on map/i);
     // await expect(this.mapViewerContainer).not.toBeVisible();
@@ -50,7 +49,6 @@ export default class ItemMapOverlayPage {
 
   // Close map button appears when map is opened
   async verifyMapIsOpenState(): Promise<void> {
-    await expect(this.page).toHaveURL(/\?viewAs=map/);
     await expect(this.mapToggleButton).toBeVisible();
     await expect(this.mapToggleButton).toHaveText(/close map/i);
     // await expect(this.mapViewerContainer).toBeVisible();
--- a/playwright/pages/item-map-overlay.page.ts
+++ b/playwright/pages/item-map-overlay.page.ts
@@ -1,0 +1,64 @@
+import { expect, Locator, Page } from "@playwright/test";
+
+export class ItemMapOverlayPage {
+  readonly page: Page;
+
+  static readonly MAP1_UUID = "8b2b3160-c5d5-012f-d95c-58d385a7bc34";
+  static readonly MAP2_UUID = "25a47180-c55f-012f-3759-58d385a7bc34";
+
+  readonly mapToggleButton: Locator;
+  readonly mapViewerContainer: Locator;
+  readonly zoomInButton: Locator;
+  readonly zoomOutButton: Locator;
+
+  static getItemURL(uuid: string): string {
+    return `/items/${uuid}`;
+  }
+
+  constructor(page: Page) {
+    this.page = page;
+
+    const dataSourceHeading = this.page.getByRole("heading", {
+      name: /data source/i,
+    });
+    // find button in parent (data source) that contains "map"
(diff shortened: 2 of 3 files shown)


playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
+import { test, expect } from "../base";
+import ItemMapOverlayPage from "../pages/item-map-overlay.page";
+import { applyRouteFilters } from "../utils/routeFilters";
+import { HEADLESS_OPTIMIZATION_FLAGS } from "../../playwright.config";
+
+test.use({
+  launchOptions: {
+    args: HEADLESS_OPTIMIZATION_FLAGS,
+  },
+});
```
samanthaandrews, line 10: I _think_ this will override anything in `playwright.config.ts`, so I'm not sure if the gating in that file is being applied.

playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        await itemMapOverlay.toggleMapView();
+        await itemMapOverlay.verifyMapIsOpenState();
+
+        // give map time to render/refresh
+        await new Promise((resolve) => setTimeout(resolve, 5000));
+      });
+
+      test("AllMaps overlay should now be displayed", async ({}, testInfo) => {
+        await itemMapOverlay.verifyMapIsOpenState();
+      });
```
samanthaandrews: This test seems to be doing the same thing as the one above. We can probably remove it.

playwright/pages/item-map-overlay.page.ts
```diff
@@ -0,0 +1,103 @@
...
+export default class ItemMapOverlayPage {
+  readonly page: Page;
+
+  readonly mapToggleButton: Locator;
+  readonly mapViewerContainer: Locator;
+  readonly mapCanvas: Locator;
+  readonly zoomInButton: Locator;
+  readonly zoomOutButton: Locator;
+  readonly compassButton: Locator;
+  readonly compassIcon: Locator;
```
samanthaandrews: It looks like these two are defined but not used. We should remove them.

playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+  JUSTIFICATION FOR .serial USAGE:
+  
+  These tests intentionally use .serial to maintain browser state across multiple test steps. This deviates from Playwright's recommendation for isolated tests for the following reasons:
+  
+  1. Performance: Avoids separate page loads for re-use of modal
+  2. Resource efficiency: Uses single worker instead of parallel workers
+  3. User journey testing: Simulates realistic workflow of opening modal once and applying multiple filters in sequence
+  4. State accumulation: Each filter builds on previous selections
+  */
```
samanthaandrews, line 26: Is this comment still correct based on the map component? Or are we using serial differently here since the map component is slightly different than the modal?

playwright.config.ts
```diff
@@ -4,14 +4,43 @@ import { defineConfig, devices } from "@playwright/test";
...
  */
-// import dotenv from 'dotenv';
-// import path from 'path';
+import dotenv from "dotenv";
+import path from "path";
 // dotenv.config({ path: path.resolve(__dirname, '.env') });
 
+const parsedEnv =
+  dotenv.config({ path: path.resolve(__dirname, ".env.local") }).parsed || {};
+
```
samanthaandrews: I think this can be removed

playwright/pages/item-map-overlay.page.ts
```diff
@@ -0,0 +1,103 @@
...
+  constructor(page: Page) {
+    this.page = page;
+
+    const dataSourceHeading = this.page.getByRole("heading", {
+      name: /data source/i,
+    });
+
+    // find button in parent (data source block) that contains "map"
+    this.mapToggleButton = dataSourceHeading
+      .locator("..")
```
samanthaandrews: Ha this locator is interesting. What do these two dots represent? A CSS class?

playwright/pages/item-map-overlay.page.ts
```diff
@@ -0,0 +1,103 @@
...
+      name: "Zoom out",
+    });
+    this.compassButton = this.mapViewerContainer.getByRole("button", {
+      name: /drag to rotate map/i,
+    });
+    this.compassIcon = this.compassButton.locator(".maplibregl-ctrl-icon");
+  }
+
+  async loadPage(gotoPage: string): Promise<void> {
+    // await this.page.setViewportSize({ width: 1280, height: 4000 });
```
samanthaandrews: Let's remove this if it's not being used

playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    await itemMapOverlay.loadPage(ItemMapOverlayPage.itemMapUrl);
+  });
+
+  test.describe("Sample Map 1", () => {
+    test("Item should display a clickable map button", async () => {
+      await itemMapOverlay.verifyMapIsClosedState();
+    });
+
+    test.describe("After button is clicked", () => {
+      test("button text should say 'close map'", async ({}, testInfo) => {
```
samanthaandrews: testInfo seems like a param that is unused. should probably remove.

playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        await itemMapOverlay.toggleMapView();
+        await itemMapOverlay.verifyMapIsOpenState();
+
+        // give map time to render/refresh
+        await new Promise((resolve) => setTimeout(resolve, 5000));
+      });
+
+      test("AllMaps overlay should now be displayed", async ({}, testInfo) => {
```
samanthaandrews: testInfo here as well

playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        });
+
+        test("Click Zoom Out", async () => {
+          await itemMapOverlay.verifyZoomOut();
+        });
+      });
+    });
+
+    test.describe("After map is closed", () => {
+      test("button text should say 'view map'", async ({}, testInfo) => {
```
samanthaandrews: testInfo here ast well

playwright/pages/item-map-overlay.page.ts
```diff
@@ -0,0 +1,103 @@
...
+    for (let i = 0; i < MAX_ZOOM_ATTEMPTS; i++) {
+      const isDisabled = await this.zoomInButton.getAttribute("aria-disabled");
+
+      if (isDisabled === "true") {
+        break;
+      }
+
+      // wait for apx a sec and try again
+      await this.zoomInButton.click({ force: true });
+      await this.page.waitForTimeout(1500);
```
samanthaandrews: I realize this is probably a tricky component to test, but I am wondering if instead of adding arbitrary wait command we could try to do something that conforms to Playwright's best practices here. `expect.poll` seems like the right call. Something like:

```
async verifyZoomIn(): Promise<void> {
  await this.mapViewerContainer.scrollIntoViewIfNeeded();
  await expect(this.zoomInButton).toBeVisible();

  // Keep clicking zoom-in until MapLibre reports max zoom (aria-disabled="true").
  // expect.poll re-runs the callback until it passes or the timeout elapses,
  // so there's no fixed sleep and no manual attempt counter
  await expect
    .poll(
      async () => {
        const disabled =
          (await this.zoomInButton.getAttribute("aria-disabled")) === "true";
        if (!disabled) {
          await this.zoomInButton.click({ force: true });
        }
        return disabled;
      },
      {
        message: "zoom-in button never reached max zoom (aria-disabled=true)",
        timeout: 20_000,
        intervals: [500, 1000, 1500], // backoff between polls
      }
    )
    .toBe(true);
}
```

playwright/tests/item-map-overlay.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      await itemMapOverlay.verifyMapIsClosedState();
+    });
+
+    test.describe("After button is clicked", () => {
+      test("button text should say 'close map'", async ({}, testInfo) => {
+        await itemMapOverlay.toggleMapView();
+        await itemMapOverlay.verifyMapIsOpenState();
+
+        // give map time to render/refresh
+        await new Promise((resolve) => setTimeout(resolve, 5000));
```
samanthaandrews: Instead of this manual "wait", can we replace it with something that uses Playwright best-practices? Perhaps a `itemMapOverlay.waitForMapReady()` util?

Something like this. Not 100% sure if checking the zoom button is the right call here or not, but something similar to check that the map has settled.

```
// in ItemMapOverlayPage
async waitForMapReady(): Promise<void> {
  await expect(this.mapViewerContainer).toBeVisible();
  await expect(this.mapCanvas).toBeVisible();
  await expect(this.zoomInButton).toBeEnabled();
}
```

samanthaandrews (changes requested):
A few comments; looking good

playwright/pages/item-map-overlay.page.ts
jbdalton: Updated this to use `expect.poll()`. In an earlier attempt, I couldn't get it to work, but using the "backoff intervals" for the polling seems to address that!

playwright/tests/item-map-overlay.spec.ts
jbdalton: Yes, I'm considering "modal" here to basically be the AllMaps component.  The serialization works the same as previous uses otherwise.  The goal is to open a triggering event (Maps Viewer in this case) and then remain there while selecting various criteria (zoom-in, zoom-out, close-button) that might also take several separate - and sometimes chained - actions.  An example of a chained action here would be the multiple, consecutive clicks on zoom-in in order to get to the "max-zoom" state on the Maps Overlay.

The overall test should be speedier like in the earlier uses, since we aren't reloading the entire IIIF UV and then re-clicking the open-maps button, waiting for that, and then testing only 1 set of interactions asynchronously each time.  We are instead opening the maps component (and waiting for its required dependencies, like the UV/IIIF manifest) only once.

playwright/tests/item-map-overlay.spec.ts
jbdalton: added a util in the POM that replaces the need for the 5 second "let maps settle down" wait.   The method looks for the css indicator from AllMaps that the maps overlay has loaded: `maplibregl-canvas-container`.  It doesn't poll for that check, but does have a 10 sec timeout, in case that element is not found.   The scenario seems to pass using this util.

I'll probably check one more time on a "cold-cache" IIIF image/map, to make sure this will still succeed in that case.

playwright/tests/item-map-overlay.spec.ts
jbdalton: These `testInfo` params were used in some earlier scaffolding for diagnostics on the screen state.  Not relevant anymore, so will remove.

playwright/pages/item-map-overlay.page.ts
jbdalton: It's an xPath short-hand for "go up exactly 1 level here."    Basically it let's you establish a parent-child locator from the bottom up, instead of "//div/find-the-thing['button']" where "div" is our "Data Source" heading.

The non-xpath shorthand way to do it alternatively would be via the longer (but more readable perhaps?) 

`
    this.mapToggleButton = this.page
      .locator("*")
      .filter({ has: dataSourceHeading })
      .last()
      .getByRole("button", { name: /map/i });
   `
   
  Where the `.last()` basically directs playwright to "top-down, go through all nodes of the DOM and find the Data Source heading and then the button directly below it that has 'map' in its name"

playwright.config.ts
jbdalton: this was from an earlier ENV-based attempt at optimization for headless in order to see the maps components display.  That should now be handled by the `process.platform` flags.

playwright/tests/item-map-overlay.spec.ts
jbdalton: removed.

playwright/tests/item-map-overlay.spec.ts
jbdalton: removed.

playwright/tests/item-map-overlay.spec.ts
jbdalton: removed.

playwright/tests/item-map-overlay.spec.ts
jbdalton: It does override the global config block for the `launchOptions`'s filter on the optimization args. This was actually intentional here to bypass the headlessOptimization environment gate (`process.env.OPT`) for this spec only.

Because MapLibre/WebGL canvas elements choke in headless mode without these flags, forcing them via test.use() means local headless runs can pass without requiring devs to pass `OPT=true` in their terminal.

The OS-level gating will still happen, since `HEADLESS_OPTIMIZATION_FLAGS` evaluates based on process.platform at the top of the config file.  So the imported array will already contain the correct platform-specific flags before `test.use()` executes.

If we find we are needing this `test.use()` override copied into other specs down the line, we could make it a tag-based filter using playwright.config's project-based settings (one for WebGL type interactions, and another default for regular headless).

playwright/tests/item-map-overlay.spec.ts
jbdalton: removed and refactored

playwright/pages/item-map-overlay.page.ts
jbdalton: removed

playwright.config.ts
jbdalton: removed.  I'm not sure if I may again need something down the line for running this kind of test specifically in playwright's GUI mode?  But i can adjust that later if it is needed.  I'm pretty sure the `HEADLESS_OPTIMIZATION_FLAGS` in the spec now handle all of this.

playwright/tests/item-map-overlay.spec.ts
jbdalton: This has continued to pass, even on the next day.  So I think removing the 5 sec. delay was a solid choice!

jbdalton:
I think this PR is ready for another pass?

jbdalton:
Jumped the gun on this.  But NOW it should actually be ready to review I think? 

There was an orphan comment that still needed removed.  And I've learned about the "changes since last review" screen in github!  (where I would've seen this, had I used that screen.)

samanthaandrews (approved):

PR #575, opened 2026-06-23
Minor locustfile update

jdohan:
The load test script for the search page has a task intended to simulate a user paginating through search results. Prior to this update, the user would mistakenly be choosing a new search term for each page they navigated to, so this addresses that to achieve a more realistic simulation of engagement.


sarangj (approved):
🚀

PR #582, opened 2026-06-25
Dr 4169/fix sort tests for ci

jbdalton:
## Ticket:

- JIRA ticket [DR-4169](https://newyorkpubliclibrary.atlassian.net/browse/DR-4169)

This issue has cropped up in a recent feature branch for grid-image options.  The tests for the search page sorting options passed on local machines, but did not on the GitHub CI.  

The solution is to force playwright to wait until “all cards” have loaded.  Then do the test assertions.  

So we’ve added this requirement to the beforeEach hook, instead of only checking that the “refine results” heading text has loaded.  That check had worked fine on CI previously, but started to fail on the above mentioned recent branch.  This should resolve these CI failures.

## Open questions

Tried several approaches, as the commits show, including polling with `toPass()` interval arguments.  But the approach we've settled on seems to be the cleanest, and required the least code changes as well.

Run tests in Github CI (they passed on localhost, so this is not an assurance of success).

### Checklist:

[DR-4169]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,7 @@ jobs:
         run: npm install
 
       - name: Install Playwright browsers
-        run: npx playwright install --with-deps chromium
+        run: npx playwright install --with-deps
 
       - name: Run build
         run: npm run build
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,7 @@ jobs:
         run: npm install
 
       - name: Install Playwright browsers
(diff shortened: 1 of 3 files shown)


samanthaandrews (approved):
LGTM

PR #585, opened 2026-07-01
DR-4168 - remove extra browsers from CI

jbdalton:
## Ticket:

- JIRA ticket [DR-4168](https://newyorkpubliclibrary.atlassian.net/browse/DR-4168)

We download and install 3 different browsers every time the test suite runs on CI.  The current suite only uses Chromium.   Removing the other 2 (safari and firefox) could easily save a minute or more off of our test runs.

## Open questions

I also added a var to the playwright.config file that assigns `chromium` to the "project" if no browser arg appears in a `--project=<browser>` flag.  This way, on localhost one can run the suite against a default chromium without having to declare a specific browser.  But they can also choose firefox, safari, or even all of them by just adding the appropriate "project" flags (e.g. `--project=webkit`, etc.).

### Checklist:

[DR-4168]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,7 @@ jobs:
         run: npm install
 
       - name: Install Playwright browsers
-        run: npx playwright install --with-deps
+        run: npx playwright install chromium --with-deps
 
       - name: Run build
         run: npm run build
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -34,6 +34,7 @@ jobs:
         run: npm run build
 
       - name: Run Playwright tests
+        # Explicitly enforce Chromium
         run: npm run playwright -- --project=chromium
 
       - uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


samanthaandrews (approved):
LGTM. Since this change has some tradeoffs to consider, I think we should follow up on this ticket with a doc outlining our Playwright test strategy on DC, as discussed in 1:1. In that doc we might figure out a test suite (full regression for example) where all browsers run so that we make sure we do test on Safari and Firefox at least occasionally. Can you make a ticket for this and link it in this PR? Also link in our 1:1 doc so we remember to talk about it. Chris is currently writing a similar doc so we can use his as a starting point once it's ready to share.

jbdalton:
Added [DR-4178 ](https://newyorkpubliclibrary.atlassian.net/browse/DR-4178) to track follow-up work on configs that will allow CI to run all-browsers (or specific set) for occasional regression suites, or other scenarios where a browser other than chrome is needed.

PR #590, opened 2026-07-10
DR-4168 - Fix typo where Firefox is on Chromium's browser assignment

jbdalton:
## Ticket:

- JIRA ticket [DR-4168](https://newyorkpubliclibrary.atlassian.net/browse/DR-4168)

This PR fixes a silent configuration typo affecting Chromium emulation.

Changes:

Firefox Profile Typo

Corrected a configuration error where the chromium single project block was accidentally assigned the Desktop Firefox device emulation profile. It now will correctly load Desktop Chrome on localhost if selected (by default it was already correctly loading chromium on localhost but if separate, per-browser project tags were used it might have failed due to that typo).

ran `npx playwright test --workers=2` on localhost

### Checklist:

[DR-4168]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -84,7 +84,7 @@ export default defineConfig({
       : [
           {
             name: "chromium",
-            use: { ...devices["Desktop Firefox"] },
+            use: { ...devices["Desktop Chrome"] },
           },
           {
             name: "firefox",


samanthaandrews (changes requested):
This isn't wrong, but it feels unnecessarily complex. What if we just run Chromium browser in CI and then locally default to run all 3 browsers? If you want to only run Chromium locally, you can pass the `--project` arg in CLI.

jbdalton:
I appreciate the idea on simplifying the config!

After removing the extra process introspection and variables in this branch, playwright does work on a single test-file when running all browsers now:  e.g. `npx playwright test --grep="search-item-results" --project=webkit --project=firefox --project=chromium` .  But if I run it against a larger set of files or the whole suite, it breaks down and the non-chromium tests error-out.

It looks like I'll need to either:

1.  keep the extra vars and env analysis in place, which also retain the ability to run a default `npx playwright test` on localhost that only hits Chromium, or
2.  remove the earlier customization entirely, and just require that the command line will always contain a `--project=<browser>` flag.  This basically would return our playwright to a default config, and thus end my adventures in developer experience improvement in this one area  : -).  A playwright test run with no project flag will in that case engage all 3 browsers by default, unless they're commented out in the playwright.config file.

Since we have another ticket ([DR-4178](https://newyorkpubliclibrary.atlassian.net/browse/DR-4178)) in the backlog to normalize the process of running multiple browsers on CI, I'm leaning toward option 2 for right now.   Just thought I'd lay out the options above.



[DR-4178]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samanthaandrews:
Hmm ok maybe I'm not quite understanding these changes and how they relate to the other recent PRs. Why can't we run the tests on all 3 browsers? They need to be able to run on all 3. I understand that you wanted to optimize in CI to only run on Chromium, but in general, failing on the other 2 shouldn't happen.

Go ahead and update the title and description of this PR for now to reflect the typo fix and I'll approve it. The next fix should be ensuring that running all 3 browsers works both locally and in CI.

jbdalton:
tl;dr -  The next PR will return playwright's localhost config to its default "--project flag required even for one browser" mode.

To explain further, I think this is just a case of me trying to have my cake (run stripped-down command on localhost that defaults to only chromium, without requiring explicit `--project=chromium`") colliding with how playwright handles parallel workers' ENV (in this case for browser) assignments.   Playwright understands multiple `--project=<browser>` flags natively _if you use the flags as playwright intended_.  The moment you introduce a process or env var in front of that explicit assignment, though, things can go awry when running parallel processes.

We were able to get away with the stripped down command (where `npx playwright test` defaults to running against chromium) before, because the test suite had all the secondary browsers commented out in playwright.config.ts. 

Once we uncommented those browers in playwright.configuration.ts, though, if we keep earlier code that allowed the "stripped down command" to run, it now collides with playwright's normal operations when running on multi-browser (aka "--project") flags. 

That's what I meant by "two options" seem to be  available to run correctly on multi-browser runs:  

1. either KEEP the "stripped-down command option as it is now, but monkeypatch the process args further upstream in the config so that the parallel workers don't lose their browser assignments, or 

2. use the `--project=<browser>` flag as it's supposed to be used, which will require us to type (or alias!) "--project=chromium" even if we are hitting only one browser on localhost.  ie. there will be no "default" option if the browser isn't declared explicitly via the `--project` flag.

The next PR will fix the config settings so that they can run  on localhost in multi-browser mode according to Option 2: "convention over customization".    After that, the next bit of work will be to get some tagging implemented that will allow the CI to also run in multi-browser mode.

jbdalton:
UPDATE: The specific browsers (except chrome) still don't work with the process.argv patch taken out.  If aimed at only one test, they often will work, but when run on the suite, they fail.  

Will open a new PR to fix it, based on option 2 as mentioned.  Right now, there are dependencies that assume chrome is running in the `npx playwright test` scenario.  And since the patch did fix this, I believe that the other option of removing those dependencies and requiring default `--project=<browser>` flags for _any_ browser assignment will also fix it.

jbdalton:
Creating a new PR for the post-fireox typo fix.  This PR only addresses the typo so should be safe to push - ie wont break anything that's not already broken (localhost firefox/webkit suite).

samanthaandrews:
> UPDATE: The specific browsers (except chrome) still don't work with the process.argv patch taken out. If aimed at only one test, they often will work, but when run on the suite, they fail.

Is one file in the suite in particular problematic? Or is it random?

samanthaandrews (approved):

jbdalton:
> > UPDATE: The specific browsers (except chrome) still don't work with the process.argv patch taken out. If aimed at only one test, they often will work, but when run on the suite, they fail.
> 
> Is one file in the suite in particular problematic? Or is it random?

Pretty random.  Once I also removed the runSpecificProject var and its dependency in the `projects[]` array (see [PR-591](https://github.com/NYPL/digital-collections/pull/591)) , then there was only one failing firefox and webkit test.  Which doesn't surprise me, since its a modal-open view.  And the same test passes on chromium but fails on those two browsers. I'll need to investigate why this is, now that "all browsers" mode has been stabilized on localhost.

PR #591, opened 2026-07-14
DR-4168 - Get all browsers to run on local host again

jbdalton:
## Ticket:

- JIRA ticket [DR-4169](https://newyorkpubliclibrary.atlassian.net/browse/DR-4168)

Removes a customization called "runSpecificProject,"  whose implementation has outlived its usefulness.  The original intent was to allow devs to type a simple "npx playwright test" and have it magically just run on chrome.  

But to support multi-browsers  simultaneously running on localhost, we need plain-vanilla "--project" defaults back.

I've removed the variable assignment, and the projects array now just filters on whether we are on CI (run only chromium, please), or on localhost (do whatever you want, but you must type in at least one browser `--project=<browser>` flag).  

If you do run it without typing in a `--project` flag, playwright will use its own default, which is to run ALL the browsers uncommented in your `projects[]` array.  So remember to include at least one, if you don't want to spawn a 300 test run across 3 browsers on localhost!

## Open questions
If you really, really want to run your own playwright suite against a "default" browser with the shortest possible command?  Answer: write an alias for it, or you could also experiment with your local .env.local setup to achieve some version of this.

Ran `npx playwright test --project=chromium`

Also ran  `npx playwright test --project=webkit --project=firefox`

And various combinations of one, two or three browsers.

NOTE: `npx playwright test` on localhost will now default to the native playwright implementation, which is to run against all listed browsers in the playwright.config.ts file's `projects[]` array. 

### Checklist:

[DR-4169]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -12,9 +12,6 @@ import path from "path";
  * See https://playwright.dev/docs/test-configuration.
  */
 
-const runSpecificProject = process.argv.some(
-  (arg) => arg.includes("--project") || arg === "-p"
-);
 const headlessOptimization = process.env.OPT === "true";
 
 export const HEADLESS_OPTIMIZATION_FLAGS =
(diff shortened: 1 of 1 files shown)


samanthaandrews:
LGTM. Thank you for simplifying.

If you want to make an easy Chromium shortcut in the project, you could add something like this to the `package.json`:

`"playwright:chromium": "npx playwright test --project=chromium`"

then you could run `npm run playwright:chromium` in the command line and in the GitHub Action.

samanthaandrews (approved):

PR #596, opened 2026-07-21
DR-4199 Add overrides for vulnerable packages, bump NPM

emretetiknypl:
## Ticket:

- JIRA ticket [DR-4199](https://newyorkpubliclibrary.atlassian.net/browse/DR-4199)

- Adds global overrides for vulnerable packages (`xmldom`, `happy-dom`, `tar`) and upgrade to npm `11.9.0`. This later change was necessary because `npm@11.9.0` is bundled with an older version of `tar`, so adding a global override for `tar` was not sufficient to get rid of the vulnerability.

## Open questions

@avertrees I'm not an npm expert, but slapping on a bunch of version overrides kinda feels like an anti-pattern. I think it's OK if we're just trying to buy time until we can move forward with the next.js upgrade? Let me know what you think.

### Checklist:

[DR-4199]: https://newyorkpubliclibrary.atlassian.net/browse/DR-4199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -247,7 +247,6 @@
       "integrity": "sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.7",
         "@babel/generator": "^7.29.7",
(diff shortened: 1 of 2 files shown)


avertrees:
I'm also not an npm expert - I think we'd eventually want to know which packages these are dependencies for and update those instead? but I think this is OK for now

avertrees (approved):

emretetiknypl:
@avertrees `xmldom` and `happy-dom` are universal viewer dependencies. But I looked into the more recent version of UV and it still depends on the vulnerable versions of those two packages, so upgrading to the latest UV wouldn't solve this.

PR #597, opened 2026-07-24
DR-3758 - feedback sheets integration - DO NOT MERGE

jbdalton:
## Ticket:

- JIRA ticket [DR-3758](https://newyorkpubliclibrary.atlassian.net/browse/DR-3758)

Adds several things to the feedback form automation:

1. checks if the test is accessing the new DEV sheet
2. will use the new DEV sheet's ID once we have added that to task definitions for QA.
3. after test runs, the new rows added in the DEV sheet are checked for expected values
4. cleanup/teardown ONLY occurs on the rows added (this is an extra safety measure to make sure that the wrong sheet is not accidentally deleted).

SUMMARY:

* DEV Sheet Safety Check: Validates that tests only target the dedicated DEV sheet before performing any operations.

* Environment Configuration: Prepared to use `SPREADSHEET_ID` from environment variables once provisioned in task definitions.

* Data Verification: Verifies that form submissions correctly populate expected row data in the DEV sheet.

* Targeted Teardown: Cleanup logic only deletes test rows created during the specific test run (identified by unique timestamped signatures), ensuring pre-existing or manual testing data remains completely untouched.

## Open questions

I've put a `skip()` around the new integration tests, but I have been able to run them locally successfully on localhost against the new DEV sheet that Lana added to the DC google account.

These are only for running automation against the DEV sheet.  If we ever want to test sheets on PROD, I think it would be best to do that manually by adding data, and having folks check the PROD sheet manually.

If changes ever get added to the PROD sheet (new rows), we will want to add those headers to the DEV sheet too.  It is my understanding that template changes don't occur very often on PROD sheet, but we should also verify this.

SUMMARY:

Integration Test skip:
* The new integration tests are currently wrapped in .skip().
* They have been verified locally against the new DEV sheet created by Lana in the DC Google account and pass cleanly.

DEV vs. PROD Testing Strategy:

* These automated integration tests are designed exclusively for the DEV sheet.
* For PROD sheet verification, I recommend sticking to manual spot-checks rather than automated writes/deletions.

Sheet Schema Syncing:

* If row headers/columns are ever updated on the PROD sheet, those changes should be mirrored on the DEV sheet. 
* (Template changes are rare, but worth keeping in mind for schema maintenance.)

Ran suite locally.  

Also ran un-skipped tests locally and verified end-to-end execution against the DEV sheet ID:  row creation, data assertions, and targeted batch deletion all succeeded.  Also verified safety checks against hitting the wrong sheet by testing against different ID values.

### Checklist:

[DR-3758]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbdalton:
This test checks for correct values in the TYPE field of the DEV spreadsheet after the form-submittion automation runs for the end-to-end integration assertions.  I have slimmed down the logging a little more, so that there is enough to alert us were we to try to hit the wrong spreadsheet initially, and a basic teardown signal of which rows we have deleted (it should always only be the 3 that were added in the test).  In the setup part of that helper, we also make sure that in addition to the spreadsheet ID, there are some other signals that we aren't accidentally on the prod sheet (e.g. number of rows to expect) and that the file-title is the one expected to appear on the DEV sheet.

playwright/utils/feedbackSetupTeardown.ts
```diff
@@ -0,0 +1,196 @@
...
+
+const EXPECTED_DEV_TITLE = "AUTOMATED-DEV-FEEDBACK-SHEET-DONT-TOUCH";
+const KNOWN_PROD_TITLE = "NEW Feedback form - Digital Collections";
+const MAX_ALLOWED_DEV_ROWS = 50;
+
+export class SheetsTeardownUtil {
+  private spreadsheetId: string;
+
+  constructor() {
+    this.spreadsheetId = process.env.SPREADSHEET_ID || "";
```
sarangj, line 20: you can get rid of the `||` here since you're raising an error anyway if the value is empty.

playwright/utils/feedbackSetupTeardown.ts
```diff
@@ -0,0 +1,196 @@
...
+    const auth = new google.auth.GoogleAuth({
+      credentials: {
+        client_email: process.env.GOOGLE_SHEETS_CLIENT_EMAIL,
+        private_key: process.env.GOOGLE_SHEETS_PRIVATE_KEY?.replace(
+          /\\n/g,
+          "\n"
+        ),
+      },
+      scopes: ["https://www.googleapis.com/auth/spreadsheets"],
+    });
```
sarangj, line 38: you could consider doing this in the constructor and assigning it to the instance, ie, move all this code into `constructor` and then do
```
this.sheets = google.sheets({ version: "v4", auth });
```
then you don't need to reinstantiate the sheets client in the different methods

playwright/utils/feedbackSetupTeardown.ts
```diff
@@ -0,0 +1,196 @@
...
+      feedback,
+      timestamp,
+      page,
+      ip,
+      browser,
+      version,
+    };
+  }
+
+  public async deleteTestRows(commentSignatures: string[]): Promise<void> {
```
sarangj, line 127: This implementation is good but I do wonder if it would be simpler to make a tab in the sheet per test run (say with some unique id), have the rest write to the tab, and then just delete the whole tab after. Thoughts? Would that require changes to app code too?

jbdalton: We did consider this in the first pass, given it would be a classic setup/teardown scenario.  But settled on re-use of a sheet partly because the SHEET_ID is declared in env settings.  Right now, the app only takes the SHEET_ID and defaults to writing to the primary tab.

Also, since our Google API credentials are provisioned via IT, adding more file/tab creation/deletion API calls on multiple tests could add to any shared Drive API rate limits.  

Dynamic creation of sheets could also add some setup/teardown latency, where we could leave orphaned test files in the Drive if a CI job were to crash mid-flight.

Keeping it constrained to the single test spreadsheet avoids those risks, but does also mean we get guaranteed fresh "rows" but not a guaranteed "fresh" sheet every time OTOH.

jbdalton:
Made updates in new PR here https://github.com/NYPL/digital-collections/pull/599

PR #599, opened 2026-07-29
DR-3758/enable dev sheet

jbdalton:
## Ticket:

- JIRA ticket [DR-3758](https://newyorkpubliclibrary.atlassian.net/browse/DR-3758)

This updates the work from previous DRAFT in closed [PR #597](https://github.com/NYPL/digital-collections/pull/597) 

If tests are run locally and you "unskip" the specific integration tests, while adding the correct sheet ID value to your own `.env.local` file, you can run the full set of tests against the DEV sheet.   You'll need to also have the correct google-account and key values in your env file.

## Open questions

If we add the google-account and key info to gha secrets, we can start running the tests unskipped on CI.  Since the Sheet ID (and name) are  different, this isolates the DEV sheet from the one used for PROD.  And the PROD values will still get propagated when the application is running on prod.

`npm run build` 

followed by cmd below after the build is finished
`npx playwright test playwright/tests/feedback-form.spec.ts --project=chromium --workers=2`

### Checklist:

[DR-3758]: https://newyorkpubliclibrary.atlassian.net/browse/DR-3758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/utils/feedbackSetupTeardown.ts
+++ b/playwright/utils/feedbackSetupTeardown.ts
@@ -110,7 +110,7 @@ export class SheetsTeardownUtil {
       return null;
     }
 
-    const [type, feedback, timestamp, page, ip, platform, browser, version] =
+    const [type, feedback, timestamp, page, ip, _platform, browser, version] =
       matchedRow;
 
     return {
--- a/playwright/utils/feedbackSetupTeardown.ts
+++ b/playwright/utils/feedbackSetupTeardown.ts
@@ -15,17 +15,16 @@ const MAX_ALLOWED_DEV_ROWS = 50;
 
 export class SheetsTeardownUtil {
   private spreadsheetId: string;
+  private sheets: ReturnType<typeof google.sheets>;
 
   constructor() {
-    this.spreadsheetId = process.env.SPREADSHEET_ID || "";
+    this.spreadsheetId = process.env.SPREADSHEET_ID!;
     if (!this.spreadsheetId) {
       throw new Error(
         "SAFETY ERROR: SPREADSHEET_ID environment variable is missing."
       );
     }
-  }
 
-  private getSheetsClient() {
     const auth = new google.auth.GoogleAuth({
       credentials: {
         client_email: process.env.GOOGLE_SHEETS_CLIENT_EMAIL,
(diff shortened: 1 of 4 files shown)


jbdalton:
Removed `skip()` since this PR now fully implements the integration tests that run against an existing DEV sheet. 

Added a serial wrapper on only the integration tests.  This helps guarantee that state is preserved on the sheet, rather than having multiple workers hit it.  It also cuts down on the googleapi calls during the feedback data checks.

playwright/utils/feedbackSetupTeardown.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    console.log(`Scanned ${rows.length} total rows in "${tabName}"...`);
+
+    const matchedRowsInfo: { index: number; text: string }[] = [];
+
+    rows.forEach((row, idx) => {
+      const feedbackText = String(row[1] || "");
+      const isMatch = commentSignatures.some((sig) =>
+        feedbackText.includes(sig)
+      );
```
sarangj, line 144: So are the comment signatures basically a way to partition feedback rows across test runs? Would it be possible to have each test run have its own tab? Mostly just to reduce the amount of logic needed here.

sarangj: if that's not possible without updating app code, then I'm cool with this approach!

jbdalton: That's a good idea! But, yeah, this would require refactoring the APP code too, since playwright is just clicking that "submit" button. We can make the teardown function do whatever we want, but the test has no control (outside of the SHEET_ID that we _are_ now manipulating for the test environment) over where the APP is storing that feedback data.

jbdalton: Added UUIDs to the signatures so they won't risk a collision if two simultaneous tests are running (like from two different devs, or CI and a dev, etc.). Just using a timestamp makes such a collision more likely.

sarangj (approved):

